PR #236, opened 2026-04-17
fix(inference): skip trim_prompt_cache on hybrid sliding-window models

### LINKED ISSUE #207: Prompt cache is ineffective for Gemma 4 (RotatingKVCache non-trimmable)
## Problem

Prompt caching is completely ineffective for Gemma 4 models (e.g. `gemma-4-26b-a4b-it`). Every request triggers a full prefill from scratch because `trim_prompt_cache` always returns 0.

In long sessions this causes severe performance degradation — a 65K-token conversation spends ~90+ seconds on prefill per turn, none of which is reused.

### Log evidence

Every single request produces:
```
WARNING: Prompt cache trim incomplete (asked for N, got 0); discarding stale cache and creating fresh
```
or:
```
WARNING: Cache trim incomplete (asked for N, got 0); invalidating cache
```

## Root cause

Gemma 4 uses a **hybrid attention architecture** where layers alternate between:
- **Local sliding-window attention** → `RotatingKVCache` (circular buffer)
- **Global full attention** → standard `KVCache`

`RotatingKVCache.trim()` always returns 0 because a circular buffer cannot be rewound to an arbitrary prefix boundary. Since mlx-lm's `trim_prompt_cache` short-circuits when *any* layer returns 0, the entire trim fails — even though the global-attention layers would be trimmable.

Additional complications:
- `num_kv_shared_layers`: last N global layers share K/V from a single projection
- Non-uniform head dimensions per layer (local: 256, global: 512)

## Upstream

This is tracked upstream in **ml-explore/mlx-lm#980** — prefix cache reuse is broken for all hybrid-architecture models (sliding window, SSM/Mamba). No fix has landed yet.

Also related: **lmstudio-ai/mlx-engine#177**.

## Possible mitigations (our side)

1. **Detect non-trimmable models at load time** (probe `can_trim_prompt_cache` or check for `RotatingKVCache` layers) and skip prompt caching entirely — avoids the wasted create → trim → invalidate cycle on every request.
2. **Implement prefix-only caching** — only cache up to the point before the sliding window wraps, allowing partial reuse of at least the global-attention layers.
3. **Log a one-time warning** that prompt caching is ineffective for this model architecture, so users know what to expect.

## Affected models

- Gemma 4 (all variants: 12B, 27B, 26B-A4B)
- Likely also affects other sliding-window models (Mistral, Qwen 3.5 Mamba hybrids, etc.)

dpalmqvist:
## Summary

Fixes #207.

- mlx-lm's `RotatingKVCache.is_trimmable()` returns `False` once the rotating buffer fills, which makes `trim_prompt_cache()` short-circuit to 0 for the whole layer list. For hybrid models (Gemma 4, Qwen3-Next, …) every divergent request logged `Prompt cache trim incomplete` and fell back to a full prefill — ~90s wasted per turn on a 65K-token Gemma 4 conversation.
- Detect non-trimmable architectures **once at model load** via a class-name allowlist probe on a throwaway cache, and store the result on `LoadedModel.supports_cache_trim`.
- When a model is flagged non-trimmable and a trim would be needed, skip the doomed `trim_prompt_cache` call and create a fresh cache directly (`fresh_cache_label="non-trimmable-hybrid"`, debug-level log). Strict-extension turns (`trim_amount == 0`) keep reusing the cache as before — `RotatingKVCache` supports appending, only trimming back fails.
- Refactor cache creation into a shared `_make_prompt_cache_for_lm(lm)` plus a `_parse_kv_cache_quant(spec)` helper so the request path and the probe share one implementation (kills three separate `split(":")` sites).

### Notes

- No new config knob — behaviour is auto-detected.
- A follow-up may be needed if real chat flows always produce `trim_amount > 0` at turn boundaries (chat-template wrapper tokens). A `logger.debug` line emits trim context per request so this is observable.
- Upstream fix for hybrid partial-trim is tracked at `ml-explore/mlx-lm#980`; this PR is the right local mitigation until then.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/integration` — **1997 passed**
- [x] `uv run ruff check olmlx tests && uv run ruff format --check olmlx tests` — clean
- [x] New tests cover: pure-KVCache probe (trimmable), RotatingKVCache probe (non-trimmable), TurboQuant+Rotating hybrid (non-trimmable), unknown-class fallback, skip-trim on divergence, and strict-extension reuse on a non-trimmable model.
- [ ] Manual: load a Gemma 4 model, send a two-turn conversation; expect one `non-trimmable hybrid cache` info line at load and no `Prompt cache trim incomplete` warnings at request time.


dpalmqvist:
Thanks for the review — addressed in ae4a6eb.

### Reviewer fixes

1. **ChunkedKVCache trimmability** — good catch. `ChunkedKVCache.trim(n)` silently clamps to `offset - start_position`, so once the prompt exceeds `chunk_size` (via `maybe_trim_front()`, called by llama4.py), trims beyond the current chunk under-deliver. Removed from the allowlist; added `ConcatenateKVCache` (verified fully trimmable) in its place. Added a comment with the mlx-lm#980 reference and an enumerated list of why each excluded type is excluded.

2. **CI regression test for new mlx-lm cache types** — added `test_allowlist_covers_all_mlx_lm_cache_types` which introspects `mlx_lm.models.cache` for all `*KVCache` classes and requires each to be explicitly classified either in `_TRIMMABLE_CACHE_CLASSES` or in a documented `known_non_trimmable` set. Any new class shipped upstream will fail the test and force a deliberate review.

3. **Deferred import + naming** — renamed `_make_prompt_cache_for_lm` → `make_prompt_cache_for_lm` (it's now part of a cross-module contract, not module-private) and added a comment at the deferred import site explaining the `inference.py` ↔ `model_manager.py` cycle it avoids.

4. **Conditional probe cleanup** — moved `gc.collect()` / `mx.clear_cache()` inside the `if probe_cache is not None` branch so the trimmable happy path skips the cleanup entirely.

### Bonus: strict-extension stable-prefix storage

Also implemented the follow-up noted in the original plan. After generation on a non-trimmable model, the chat template is re-rendered with the completed assistant turn (`add_generation_prompt=False`), tokenized, and the KV cache is extended by the boundary tokens that appear only in that rendering. Stored `cached.tokens` then matches what the next turn's prompt will start with — enabling strict-extension reuse on hybrid models.

It's gated by a divergence check: if the re-rendered target doesn't match the raw `(prompt + generated)` sequence, `_advance_cache_to_prefix` rejects and falls back to current behaviour. Verified with Gemma 4 that the fallback is taken cleanly — Gemma 4's thinking template injects `<|channel>thought` primers on `add_generation_prompt=True` that aren't present on `=False`, so the prompts aren't token-equivalent at the assistant boundary. This is an architectural property of that template (thinking priming can't be preserved across turns without trim support), not a bug in the extension logic. Hybrid models without such primers (e.g. Qwen3-Next future releases, non-thinking mode) should benefit.

### Test status

- `uv run pytest tests/ --ignore=tests/integration` — **2002 passed**
- `uv run ruff check olmlx tests && uv run ruff format --check olmlx tests` — clean
- Manual Gemma 4 two-turn test: no `Prompt cache trim incomplete` warnings; probe info line fires once at model load.

dpalmqvist:
Pushed 1c6f4a9 addressing the four review items.

**Pyright (CI failure)**: the signature typed the variable as `Callable[...] | None` then redeclared it via `def`, which Pyright treats as a redeclaration with an incompatible return type. Fixed by naming the inner function `_build_stable_prefix` and assigning it to the typed variable at the end of the branch. Local pyright now shows 0 errors.

**`[]` → `None` in VLM branch**: agreed — the empty list silently passed the `target is not None` gate in the store path and was only rejected deeper in `_advance_cache_to_prefix`. Now returns `None` explicitly and the builder's return type is `list[int] | None`, propagated through `_store_prompt_cache_after_generation` and `_stream_completion` signatures.

**`mx.eval(...)` on `.state` attribute**: switched to the upstream pattern `mx.eval([c.state for c in prompt_cache])` (the same shape mlx-lm uses at `mlx_lm/generate.py:442,584,1161`). All cache types in `mlx_lm.models.cache` define `state` as a property, so the defensive `getattr` filter was unnecessary and the fall-through could have left tensors unresolved.

**Identity sentinel → explicit flag**: `stored_tokens is not raw_sequence` replaced with `used_stable_prefix: bool` set alongside the assignment.

**Non-streaming `stable_prefix_builder`**: I don't think this is a bug — `_full_completion_inner` never calls `prompt_cache_store.async_set` or `_store_prompt_cache_after_generation`, and the `use_prompt_cache` gate in `generate_chat` already requires `stream=True` (inference.py:2443: `use_prompt_cache = settings.prompt_cache and stream and ...`). So non-streaming has never participated in prompt caching and there's no divergent `cached.tokens` for it to produce. Happy to revisit if I'm reading the code wrong, but I couldn't find a call chain that writes to the store from the non-streaming path.

CI status:
- Pyright: fixed locally (0 errors).
- Full unit suite: 2002 passing.
- Ruff: clean.

dpalmqvist:
Pushed 67b2a8f.

**Tests CI failure (root cause found)**: the load-time probe walks the cache-layer class names, and the integration tests mock `make_prompt_cache` to return `[MagicMock()]`. `MagicMock` isn't in the allowlist, so every fixture model was being flagged non-trimmable — cache reuse was silently disabled across the whole integration suite, which is what made `cache_read_input_tokens` return 0 on the second turn. Patched `_cache_supports_trim` to return True in `mock_mlx_primitives` so test models behave like standard KVCache-only architectures. Locally 2048 tests pass (unit + integration).

**Review comments addressed**:

- **Stale Qwen3-Next comment** (inference.py:1516): rewrote the block comment. This `elif` is now a defensive path for cases where `supports_cache_trim=True` but a layer's `trim()` still under-delivers (custom caches, allowlist misses). The probe catches known-hybrid caches upstream of it.
- **Shallow message snapshot** (inference.py:2573): switched to `[dict(m) for m in messages]` — the one-level deep copy the reviewer suggested, which is sufficient for text content while still covering the race where a caller mutates a dict after the generator returns.
- **`make_prompt_cache_for_lm` docstring**: added a note that it's a cross-module public contract imported via deferred import from `model_manager._probe_cache_trim_support` — renames wouldn't be caught by static analysis and would fail only at model-load time.

On the `make_prompt_cache_for_lm` end-to-end coverage: the integration tests actually do exercise the full probe → factory call chain (that's how the mock-cache issue surfaced). The test-breakage on this PR proves a rename would be caught by CI, which is the coverage the reviewer wanted.

**`mx.eval` blocking on event loop** (inference.py:1737): left as-is per the reviewer's own note — boundary tokens are small (~10-50) and the concern is architectural rather than a required fix.

dpalmqvist:
Reverted the stable-prefix mechanism in 19171d0.

Reasoning: the reviewer flagged three separate correctness issues (decode round-trip is lossy, VLM tokenizer mismatch, forward pass holds the inference lock), and on top of that the mechanism doesn't actually help Gemma 4 in practice — its thinking-mode template injects `<|channel>thought` primers on `add_generation_prompt=True` that aren't present with `False`, so the token sequences are never prefix-equivalent and `_advance_cache_to_prefix` always rejected. Net: no value for the PR's motivating model, real correctness risk for other hybrids.

Keeping this PR focused on the core fix + reviewer-batch cleanups:
- Load-time probe + skip doomed `trim_prompt_cache` call (issue #207 headline)
- `_TRIMMABLE_CACHE_CLASSES` allowlist with `ChunkedKVCache` deliberately excluded
- Regression fence test that catches new mlx-lm cache types
- Renamed `make_prompt_cache_for_lm` (public cross-module contract, documented)
- Conditional probe cleanup
- Integration-conftest patch for `_cache_supports_trim`

Diff shrinks by ~200 lines (`_advance_cache_to_prefix`, the builder closure, the `stable_prefix_builder` parameter threading, 4 unit tests, and the `add_generation_prompt` kwarg additions to `_apply_chat_template`/`_apply_chat_template_text` that only existed for the builder).

Stable-prefix is the right idea but needs a cleaner implementation path — using raw token IDs (not decode → re-encode), matching the prompt's original tokenizer for the re-render, and running the prefill outside the inference lock. Will revisit in a follow-up once the core fix here is in.

dpalmqvist:
Pushed b0883d5 addressing all three claude-review findings.

1. **SpectralQuant calibration reload** — the probe now calls `make_prompt_cache(cache_model)` directly (bare mlx-lm cache) instead of `_make_prompt_cache_for_lm`. Trim behaviour is governed by the underlying layer classes, not the quantization wrapper, so the quantized factories were doing disk I/O for nothing. This also eliminates the need for `model_manager.py` to import anything private from `inference.py`.

2. **Cross-module naming trap** — since the probe no longer calls it, renamed back to `_make_prompt_cache_for_lm` (single private caller in `_setup_prompt_cache`). The docstring warning is gone. The DI approach the review suggested solves the same problem, but with nothing left to inject, the simpler private-function shape is cleaner.

3. **Unverified allowlist entries** — reviewer suggested annotating `ConcatenateKVCache` or dropping it. I verified it's in the installed mlx-lm (class at `mlx_lm/models/cache.py:178`, used by afm7, `trim(n)` reliably subtracts from offset) and added source-line references in the allowlist comment. Also dropped `TurboQuantKVCache` / `SpectralQuantKVCache` from the allowlist since they're unreachable through the probe now, and added a reverse-direction assertion to `test_allowlist_covers_all_mlx_lm_cache_types`: every allowlist entry must exist as a real class in `mlx_lm.models.cache`. Typos and preemptive entries fail the test.

Locally: 2043 tests pass, pyright 0 errors on the touched files, ruff clean.

PR #290, opened 2026-05-07
fix(dflash): make max valid pivot reachable in draft training

dpalmqvist:
## Summary
- `mx.random.randint(lo, hi)` is exclusive on the upper bound, so with `hi = seq - block_size - 1` the last valid window position (`seq - block_size - 1`) was never sampled. Set `hi = seq - block_size` so the full `[block_size, seq - block_size - 1]` range is reachable.
- Minimum required sequence length drops from `2*block_size + 2` to `2*block_size + 1` as a side effect.

Rebased onto current `dflash-prepare-285`. The other three fixes from the original draft (num_shards validation, hidden-state slot reset, index validation) were already addressed by `a1e96d8` and `a680f32` on the base branch, so they are dropped from this PR.

## Test plan
- [ ] `uv run pytest`
- [ ] Sample pivots over many steps; confirm distribution covers `seq - block_size - 1`
- [ ] Confirm a sequence of length `2*block_size + 1` no longer triggers the guard


dpalmqvist:
Subsumed by #291 (squash-merged in 54b0dd1). The two pivot-fix commits from this branch (`ef93feb` make-max-pivot-reachable and `7f815a1` regression test) were cherry-picked into #291 during the rebase, so the content is fully on main now. Closing as redundant.

Issue #285, opened 2026-05-06
Add DFlash support in olmlx


dpalmqvist:
https://github.com/z-lab/dflash

---

## Independent review

I verified the prior analysis against (a) the actual code in `olmlx/engine/dflash/` on `main` and (b) the upstream `z-lab/dflash/dflash/model_mlx.py` source. Most findings hold up. A few additions and one correction below.

### What I confirm

- **Adapter pattern is the wrong abstraction.** `engine/dflash/adapters.py` (132 lines) hardcodes layer-tree introspection per model family (`Qwen3Adapter` only, aliased to `qwen2`). Upstream's `_LayerHook` is a universal monkey-patch that finds `model.layers` (or `model.model.layers`, or `model.language_model.layers`) and wraps the indices listed in `target_layer_ids`. That's what makes Gemma/Llama/MiniMax/gpt-oss work without per-architecture code.
- **The draft architecture cannot load any z-lab pre-trained weights.** Independently confirmed against `model_mlx.py`:
  - target hidden combination — upstream uses `nn.Linear` (concat-and-project) + `hidden_norm`; we use mean pooling at `draft_model.py:172-176`.
  - attention — upstream has self-attention with RoPE, QK-norm, GQA (`num_key_value_heads`), and per-layer sliding window; we have a custom non-causal cross-attention with no RoPE, no GQA, and no sliding window (`DFlashAttention` at `draft_model.py:35-78`).
  - draft KV cache — upstream maintains a per-layer cache (`KVCache`/`RotatingKVCache`) and reuses it across draft steps; our `_draft_block` recomputes attention from scratch each token (`decoder.py:163-180`).
  - embed/lm\_head — upstream's `bind(target_model)` borrows the target's `embed_tokens` and `lm_head`; ours owns trainable copies (`draft_model.py:111, 128`).
[...]

Issue #284, opened 2026-05-06
GPU stream error with Qwen3.5: VLM detection + streaming+tools+prompt cache


dpalmqvist:
## Summary

Qwen3.5 models (e.g. `mlx-community/Qwen3.5-9B-6bit`) hit `RuntimeError: There is no Stream(gpu, N) in current thread` in two distinct code paths. The workaround is to strip `vision_config` keys from the model config, but a second issue remains with prompt cache reuse.

## Environment

- macOS 15.x, M2 Pro (32GB)
- olmlx @ 28144c4
- mlx-community/Qwen3.5-9B-6bit (6-bit, 9GB)
- mlx 0.26.x, mlx-vlm installed

## Issue 1: VLM detection triggers mlx_vlm GPU stream bug

Qwen3.5 models include `vision_config`, `image_token_id`, etc. in config.json. `_detect_model_kind()` correctly identifies them as VLM, but mlx_vlm `stream_generate` crashes:

```
File "mlx_vlm/generate.py", line 327, in wired_limit
    mx.synchronize(s)
RuntimeError: There is no Stream(gpu, 0) in current thread.
```

**Repro** (streaming, any prompt):
```bash
curl -N http://localhost:11434/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"mlx-community/Qwen3.5-9B-6bit","messages":[{"role":"user","content":"Hello"}],"max_tokens":10,"stream":true}'
```

**Workaround**: strip `vision_config`, `image_token_id`, `video_token_id`, `vision_start_token_id`, `vision_end_token_id` from config.json. Model then routes through mlx-lm text path.

## Issue 2: Prompt cache reuse on multi-turn + tools crashes in mlx-lm

With vision keys stripped (text mode), first turn with streaming+tools works fine. Second turn (prompt cache reuse) crashes in mlx_lm:

```
File "mlx_lm/generate.py", line 442, in generate_step
    mx.eval([c.state for c in prompt_cache])
RuntimeError: There is no Stream(gpu, 16) in current thread.
```

Server log shows:
```
Prompt cache non-trimmable-hybrid: creating fresh cache for 4079 tokens
Generation complete: 4079 prompt tokens, 32 tokens generated, 21.53s total  [OK]
Second request hits prompt cache, then crashes
```

Single-turn with tools+streaming works; multi-turn with tools+streaming fails.

**Workaround**: start olmlx with `OLMLX_PROMPT_CACHE=false`.

[...]

PR #292, opened 2026-05-08
fix(dflash): make DFlash work end-to-end on Qwen3.6-35B-A3B (hybrid VLM)

dpalmqvist:
## Summary

Five targeted bug fixes that make DFlash (block-diffusion speculative decoding) work end-to-end on Qwen3.6-35B-A3B and other hybrid-VLM targets. 

### Fix 1 — VLM config descent (`_text_config`)
Qwen3.6-35B-A3B nests its text-tower configuration under `text_config`. The existing `_build_draft_config` read top-level fields only, producing a misconfigured `DraftConfig` with wrong hidden_size / head_dim / vocab_size. Added `_text_config()` helper that descends into `text_config` when present (for multimodal targets), with a defensive `"hidden_size" in nested` check.

### Fix 2 — Pad-target contamination (`_select_pivot` + `_draft_loss`)
Training batches include right-padded sequences. When no `pad_token_id` is found on the tokenizer (Qwen3.x EOS==PAD aliasing), pad positions leak into the loss, training the draft on garbage targets. Added `_select_pivot` helper (scans from right using reversed argmax for the trailing-pad boundary) and `_draft_loss` pad masking with a separate `pad_for_loss` (never falls back to EOS, preserving mid-stream turn-boundary tokens for multi-turn data).

### Fix 3 — `_detect_model_kind` top-level fallback
Hybrid VLM models (Qwen3.6-35B-A3B) that ship a dedicated text-only module in mlx-lm would previously be routed to mlx-vlm (matching `model_type`), then crash on text inference with a Metal stream error (issue #284). Added a fallback: when the inner model type has no mlx-lm module but the top-level type does AND `linear_attention` is in `layer_types`, delegate to `"text"` so mlx_lm.load handles the top-level config.

### Fix 4 — `_resolve_draft_path` local-path short-circuit
A local absolute path passed as `speculative_draft_model` (e.g. `--speculative-draft-model /path/to/dflash`) was treated as an HF repo id and sent through `ensure_downloaded`, producing a misleading `HFValidationError`. Added an `is_absolute()` check after `expanduser()` that raises `FileNotFoundError` for non-existent absolute paths and skips download for existing ones.

### Fix 5 — `_GDNStateCapture` rollback ordering
Two bugs in the GatedDeltaNet state capture for rejection rollback:
- `captured_modules == expected_gdn_modules` triggered `bool(mx.array)` on `nn.Module.__eq__`, crashing for any module with multi-element tensors. Replaced with identity-based `_order_matches(is)`.
- `model.named_modules()` uses LIFO stack traversal, returning GDN submodules in reverse of forward-pass order. Replaced with `_get_layers(model)` walk in natural index order, with a `>1 GDN per layer` guard against future multi-GDN architectures.

### Fix 6 — DFlash draft attention mask and RoPE offset (a0d909f)
Two performance bugs in `DFlashAttention.__call__` found by comparison against the z-lab/dflash reference implementation (issue #285):
- **Non-causal mask**: Changed full-attention mask from `"causal"` to `None` (bidirectional). DFlash block-diffusion relies on parallel bidirectional attention within the block so all MASK positions can condition on each other. Causal attention forced autoregressive-within-block, reducing acceptance rate.
- **Sliding-window RoPE offset**: Now updates `cache.offset += skip` directly (matching the reference) instead of using a local `ctx_rope_offset`. The skipped-token positional gap must be tracked across calls so subsequent steps inherit the correct RoPE base position.

Note: draft checkpoints trained before this PR used causal attention; running them with bidirectional attention at inference will produce a distribution mismatch. Re-training is recommended.

## Testing

- 209 tests pass (122 model_manager, 42 test_dflash, 45 test_dflash_prepare/precompute/distill), ruff clean
- New regression tests for each fix
- End-to-end verified on Qwen3.6-35B-A3B


dpalmqvist:
Addressed all three review comments in 1b6b47a:

1. **Step-counter bug** — fixed. New `real_step` counter threaded through LR schedule, log-every cadence, and progress-bar fraction. Added `TestSkippedBatchesPreserveStepBudget` that fails on the bug (2/5 real updates) and passes on the fix (5/5).
2. **Fallback comment** — updated verbatim to your suggested wording. The cardinality-check chain is now explicit.
3. **Denom double-eval** — applied your suggestion: `kl_loss = (kl * valid).sum() / denom.astype(kl.dtype) * (t * t)`.

77 dflash tests pass; ruff clean.

dpalmqvist:
Round-2 review addressed in 45a09ba:

1. **`_resolve_draft_path` relative-path false positive** — fixed. Gated on `is_absolute()`. New regression test `test_relative_path_collision_does_not_short_circuit` constructs the colliding-CWD scenario and confirms `ensure_downloaded` is still called.
2. **Silent under-training when batch stream exhausts** — added two warnings in `prepare_dflash_draft`: one for the `real_step == 0` degenerate case (every batch skipped → checkpoint is random init), one for partial completion (`real_step < steps`). New test `test_warns_when_batch_stream_exhausts_before_steps`.
3. **O(n²) `not in` in rollback diagnostic** — applied your suggestion verbatim with `captured_set` / `expected_set`.
4. **Type-checker hint on `denom`** — initialised `valid` and `denom` to `None` before the `if pad_token_id is not None` block; added explicit `denom is not None` guard in the KL branch.

199 tests pass; ruff clean.

dpalmqvist:
Round-3 addressed in aadcc3d:

1. **Inline `prepare.py:760` exhaustion warning** — already addressed in 45a09ba (your previous round), I think your latest review snapshot was on the older commit. The warnings (`real_step == 0` and `real_step < steps`) are in place.
2. **`_GDNStateCapture` fallback misleading error** — applied. If `_get_layers` returns no GDN-containing layers but `named_modules()` finds orphaned GDN modules, raise immediately with a clear "modules found but unreachable via `_get_layers`" message. If no GDN exists anywhere (pure full-attention model), leave `_expected_gdn_modules` empty since `rollback()` won't fire.
3. **`_detect_model_kind` belt-and-suspenders** — added the redundant `"linear_attention" in layer_types` check inside the fallback so the "hybrid linear-attention" label is verifiable from the inner conditions alone, independent of the outer guard.
4. **`caplog.messages` vs `rec.message`** — switched to the portable `caplog.messages` idiom in `test_warns_when_batch_stream_exhausts_before_steps`.

199 tests pass; ruff clean.

dpalmqvist:
Round-4 addressed in be3664d:

1. **`_select_pivot` prefix vs non-pad count** — fixed, but with one departure: your suggested implementation (`argmax` on the pad mask) returns the *first* pad anywhere, which is even more wrong than counting when mid-stream EOS markers exist (it collapses the apparent prefix to a tiny window). The correct primitive for right-padded data is `seq_len - trailing_pad_count`, computed by reversing along the sequence axis and locating the first non-pad. New regression test `test_handles_pad_token_appearing_mid_content` covers a row with three mid-stream pads + two trailing pads and confirms the pivot can reach the actual prefix boundary.
2. **`strict=True` in `_order_matches`** — dropped. The early-return on length mismatch already covers it; comment notes why.
3. **`denom is not None` redundant guard** — replaced compound check with `valid is not None` plus `assert denom is not None` for type narrowing.

200 tests pass; ruff clean.

dpalmqvist:
Round-5 addressed in 5e03e8c:

1. **Target forward before pivot check** — real perf bug, fixed. The loop now resolves `input_ids` first, calls `_select_pivot`, and only runs `_capture_target_outputs` if the pivot is accepted. Skipped batches no longer burn a 35B-target forward pass. New regression test `test_skipped_batches_do_not_run_target_forward` patches the wrapper and asserts call count equals real gradient steps (1 forward for 1 real-content + 1 pad-only batch sequence).
2. **`assert` vs `cast`** — switched to `typing.cast(mx.array, denom)` so the type-narrowing hint is robust to `python -O`.

201 tests pass; ruff clean.

dpalmqvist:
Round-6 addressed in 8bb8b15:

1. **`rope_theta` silent fallback** — added `logger.warning` when neither flat `rope_theta` nor nested `rope_parameters.rope_theta` is present. Operator now sees the misconfiguration at training time rather than discovering it via poor acceptance at inference.
2. **EOS-as-pad masking turn separators** — split the variable into `pad_for_pivot` (keeps the `eos_token_id` fallback so trailing-pad detection still works on right-padded data) and `pad_for_loss` (only true `pad_token_id`, never EOS). The loss mask is now defensive belt-and-suspenders for genuine pad tokens; the pivot constraint already keeps targets inside the unpadded prefix, and we no longer zero-weight mid-stream EOS turn separators.
3. **Swallowed `AttributeError`** — removed the `except AttributeError` wrapper around `_get_layers(model)`. The original error message ("Cannot find layers in <Class>; tried .model.layers, .language_model.layers, .layers") points directly at the configuration problem and is more useful than the downstream orphaned-modules diagnostic that previously masked it.

The earlier two inline comments (`prepare.py:776` perf + `prepare.py:378` assert) were addressed in 5e03e8c — your latest snapshot was on `be3664d`.

201 tests pass; ruff clean.

dpalmqvist:
Round-7 addressed in 2312005:

1. **Infinite-loop hazard with infinite-iterator + all-padding** — fixed. Added a `consecutive_skips` counter capped at `max(steps * 4, 1000)`; aborts with a self-explanatory `RuntimeError` listing likely causes. Counter is reset at the start of each real-work iteration so a normal run with a few rough patches doesn't false-positive. New regression test `test_aborts_when_infinite_all_padding_iterator` builds the exact infinite-pad-only scenario and asserts termination.
2. **Trailing-EOS conservatism in `_select_pivot`** — documented. When a row's final real token equals `pad_token_id`, the trailing-pad scan absorbs it and reports `real_lens` one short. Direction is safe (no invalid windows), accepted as conservative-but-correct rather than threading per-row "true length" through the loader.
3. **`typing.cast` in `_draft_loss`** — already in `5e03e8c` (round-5). Latest snapshot apparently rebased on an older commit.

202 tests pass; ruff clean.

dpalmqvist:
Round-8 addressed in 81234ba:

1. **`_select_pivot` docstring lead sentence** — rewrote to lead with "scans from the right to find the trailing-pad boundary (reversed argmax for the first non-pad position)" so the algorithm is correctly described upfront. Naive left-to-right `argmax(ids == pad)` is now explicitly mentioned as the wrong approach with reasoning.
2. **Sub-layer GDN-count guard** — added `RuntimeError` when any decoder layer carries >1 GDN submodule. Today's architectures are all single-GDN-per-layer; a future multi-GDN variant would otherwise reproduce the LIFO-ordering bug at sub-layer granularity. The assertion surfaces the violation up front.
3. **`_resolve_draft_path` clearer error** — when given an absolute path to a non-existent directory, raises `FileNotFoundError` with the actual path instead of falling through to `ensure_downloaded` and getting a misleading `HFValidationError`. New regression test `test_absolute_missing_path_raises_file_not_found`.

The two other inline comments this round (EOS-as-pad in `_draft_loss`, swallowed `AttributeError` in `_GDNStateCapture`) were already addressed in 8bb8b15 (round-6). Latest review snapshot was on `5e03e8c`.

203 tests pass; ruff clean.

dpalmqvist:
Round-9 inline comments are all on items already addressed in earlier commits:

1. **Infinite-loop guard** — already in `2312005` (round-7). `consecutive_skips` counter caps at `max(steps * 4, 1000)`; aborts with a `RuntimeError` listing likely causes. Test `test_aborts_when_infinite_all_padding_iterator` covers the infinite-pad-only scenario.
2. **`cast` for `valid`/`denom`** — already in `5e03e8c` (round-5) using `typing.cast(mx.array, denom)`. The compound `valid is not None and denom is not None` variant was rejected in round-4 as misleadingly implying they could be set independently; `cast` was specifically recommended (over `assert`) in round-5 because `python -O` strips assertions.
3. **Trailing-EOS conservatism doc** — already in `2312005` (round-7). Docstring spells out: "when a row's *final* real token equals `pad_token_id`, the trailing-pad scan absorbs it into the pad count, reporting `real_lens` as one shorter than the true content boundary."

Latest review snapshot appears to be on `be3664d` or earlier. Current HEAD is `81234ba` with 203 tests passing.

dpalmqvist:
Round-10 addressed in 1696223:

1. **Missing combined-path test** — added `test_vlm_config_with_rope_parameters_nested_under_text_config` that exercises the actual Qwen3.6 config shape (rope_parameters nested *inside* text_config). The two descent paths now have explicit composition coverage.
2. **`max_consecutive_skips` too generous** — tightened from `max(steps*4, 1000)` to `max(steps*2, 500)`. For a 2000-step run, that drops the worst-case stall from ~8000 syncs to ~4000 (roughly halves the time-to-error on a degenerate dataset).
3. **Inline comment for redundant `linear_attention` check** — added directly next to the condition, noting the duplication with the outer guard and the refactoring-safety rationale.

204 tests pass; ruff clean.

dpalmqvist:
Round-11 addressed in d410b25:

1. **`rope_scaling` regression for top-level VLM configs** — real bug, fixed. When `_build_draft_config` switched to read fields via `text_cfg`, a VLM config carrying `rope_scaling` at the top level (not mirrored inside `text_config`) silently dropped scaling. Now falls back: `text_cfg.get("rope_scaling") or target_cfg.get("rope_scaling")`. New regression test `test_rope_scaling_falls_back_to_top_level_for_vlm` covers the exact scenario.
2. **Multi-GDN-per-layer error message** — applied your suggested wording: the message now spells out that `layer.named_modules()` uses LIFO stack traversal so multiple GDNs would be mis-ordered relative to the forward pass.

205 tests pass; ruff clean.

dpalmqvist:
Round-12 addressed in a94aea8:

**`rope_parameters` top-level fallback** — applied. Now `rope_params = text_cfg.get("rope_parameters") or target_cfg.get("rope_parameters")`, matching the `rope_scaling` resolution pattern from round-11. No known shipping config hits this asymmetry today (Qwen3.6 nests `rope_parameters` inside `text_config`), but the inconsistency would have been a latent trap for future multimodal architectures. New regression test `test_rope_parameters_falls_back_to_top_level_for_vlm`.

206 tests pass; ruff clean.

dpalmqvist:
Round-13 addressed in e935870:

**KL + pad-masking test coverage** — added two tests in `TestDraftLossPadMasking`:
- `test_kl_distillation_with_all_pad_targets_yields_zero_loss`: `distill_alpha=0.5`, every target is pad, draft and target softmaxes constructed to disagree (so an unmasked KL would dominate). Loss must be exactly `0.0`.
- `test_kl_distillation_with_mixed_targets_masks_pad_positions`: one real target + two pads, draft and target both peak at the real id; without masking the pad-position CE explodes to ~13 nats; with masking the loss approaches zero.

Both tests fail on a hypothetical regression that drops `valid` from the KL reduction or mishandles `denom_kl.astype(kl.dtype)`.

208 tests pass; ruff clean.

dpalmqvist:
Round-14 addressed in ffd0cd9:

1. **Rollback diagnostic shows module names, not id() counts** — applied. Error now includes `captured order`, `expected order`, and `captured-only`/`expected-only` class-name sequences. An operator hitting this can see the misaligned layer type without needing a debugger.
2. **Multi-GDN-per-layer error workaround hint** — added concrete guidance: replace the per-layer `named_modules()` collection with a manual depth-first walk over the layer's children in declaration order (e.g. `vars(layer).values()` with recursion into containers).
3. **KL+pad-masking test gap** — already addressed in `e935870` (round-13). The two new tests `test_kl_distillation_with_all_pad_targets_yields_zero_loss` and `test_kl_distillation_with_mixed_targets_masks_pad_positions` cover the distillation-with-masking branch.

The probabilistic `test_handles_pad_token_appearing_mid_content` flag (5.5e-8 failure rate) is left as-is — explicitly acknowledged by the reviewer as not a real risk.

208 tests pass; ruff clean.

dpalmqvist:
Round-15 addressed in cbc8960.

**Real correctness bug caught**: Qwen3.x tokenizers ship with `pad_token_id == eos_token_id == 151643` (both `<|endoftext|>`). My round-6 split intentionally avoided falling back to `eos_token_id`, but didn't account for the case where `pad_token_id` itself is the same value. Result: `_draft_loss` was zero-weighting every mid-stream EOS, training the draft to never predict turn boundaries. Fix: `pad_for_loss = None` when `pad_token_id == eos_token_id`. `_select_pivot` (via `pad_for_pivot`) still handles trailing pad correctly. New regression test `test_pad_for_loss_disabled_when_pad_aliases_eos` covers exactly the Qwen3.x aliasing scenario.

This bug has been silently degrading the in-flight 2000-step training run (currently at step ~1000). I'll let it complete to compare against the pre-fix baseline anyway — useful data point — but a follow-up training run with the fix is justified.

Other items addressed:
- `_text_config` defensively descends only when the nested block carries `hidden_size` (protects against a non-VLM that uses `text_config` for an unrelated purpose).
- Dropped misleading "pass the correct value explicitly" from `rope_theta` warning (no `--rope-theta` CLI flag exists).
- Tightened `max_consecutive_skips` from `max(steps * 2, 500)` to `steps * 2 + 50` so short-step test runs don't need 500 syncs to abort.

209 tests pass; ruff clean.

dpalmqvist:
Round-16 addressed in 275722f:

**Double `find_spec` in `_detect_model_kind` fallback** — deduped. `mapped_lm_present = importlib.util.find_spec(...) is not None` is computed once and referenced in both the new top-level fallback condition and the primary check.

The other round-16 notes were already explicitly accepted ("Acceptable given the design constraints" for the trailing-EOS conservatism) or addressed in earlier commits (sub-layer LIFO residual was elaborated in round-14's `ffd0cd9` with workaround guidance).

122 model_manager tests pass; ruff clean.

matisaar:
Found while working on finance ML: https://matisaar.github.io/dataforge-sales/

PR #293, opened 2026-05-10
perf(kv-quant): compile TurboQuant/SpectralQuant hot paths

dpalmqvist:
## Summary

- Wraps the per-layer TurboQuant and SpectralQuant quantize/dequantize kernels in `mx.compile`, keyed by full input shape via small LRU caches (mlx 0.31's `shapeless=True` only handles last-axis changes — leading dims get baked into the trace, so per-shape caching is required for correctness across decode and prefill calls).
- For SpectralQuant, dispatches between an unrolled `where`-loop and a vectorized `argmin` per regime based on `n_levels` and the implied broadcast tensor size: loop wins for small codebooks or large prefills (avoids the OOM-risky `(B, h, seq, dim, n_centroids)` materialization); argmin wins for big codebooks at small shapes.
- Adds two benchmark scripts under `scripts/` so future tuning has a regression check.

## Impact

End-to-end with TurboQuant 4-bit KV cache, 512-token prompt + 64-token decode:

| Model | Prefill speedup | Decode speedup |
|-------|-----------------|----------------|
| Qwen2.5-0.5B-4bit | 1.31× (3405 → 4446 tok/s) | **1.98× (65 → 130 tok/s)** |
| Qwen2.5-3B-4bit (1024-tok prompt) | 1.12× (1385 → 1556 tok/s) | 1.33× (40 → 54 tok/s) |

Microbench highlights (head_dim=128, n_heads=32, per call):

| Path | seq | baseline ms | compiled ms | speedup |
|------|-----|-------------|-------------|---------|
| TurboQuant 4-bit quantize | 1 | 0.49 | 0.28 | 1.7× |
| TurboQuant 4-bit quantize | 2048 | 45.3 | 15.7 | 2.9× |
| TurboQuant 2-bit quantize | 2048 | 14.2 | 3.4 | 4.2× |
| SpectralQuant 4/2 quantize | 2048 | 343.9 | 9.4 | **36×** |
| SpectralQuant 8/1 quantize | 2048 | 608.3 | 242.1 | 2.5× |

The decode win shrinks as model compute grows relative to KV-cache overhead — expected, since compile only touches the quant kernels.

## Test plan

- [x] `uv run pytest tests/` — 2382 passed, 3 skipped (pre-existing)
- [x] `uv run python scripts/bench_quant_compile.py` — head-to-head microbench
- [x] `uv run python scripts/bench_quant_e2e.py --model ~/.olmlx/models/mlx-community_Qwen2.5-0.5B-Instruct-4bit --bits 4 --prompt-tokens 512 --gen-tokens 64`
- [x] `uv run python scripts/bench_quant_e2e.py --model ~/.olmlx/models/mlx-community_Qwen2.5-3B-Instruct-4bit --bits 4 --prompt-tokens 1024 --gen-tokens 64`


dpalmqvist:
Thanks — addressed all three in 81cce90:

**1. Dequant cache keys missing `n_levels`** ✅ — fixed in both `_compiled_dequant_core` (turboquant) and `_compiled_spectral_dequant_core` (spectralquant). You're right that `unpack_indices` strips bit-width from the index shape, so without `n_levels` in the key, two different-`bits` models would alias and force `mx.compile` to retrace on every call.

**2. LRU thrashing on varied prompt lengths** — bumped `maxsize` from 32 → 128 instead of bucketing the cache key. The literal bucketing-the-key suggestion would defeat the purpose: `mx.compile` would still retrace internally because the *actual* shape passed to the compiled function differs across calls that map to the same bucket. The only sound version is to bucket the *input* (pad up), but that changes the computation. 128 entries comfortably absorbs the typical prefill-length distribution; if real workloads thrash it later, padding-then-trim is the right next step. (Decode is one stable entry, immune either way.)

**3. O(n²) `_build_prompt`** ✅ — single encode + slice + decode.

dpalmqvist:
Round-2 review addressed in 59cdfb4 — and two clarifications on items I'm not changing:

**`_compiled_dequant_core` cache key (round-2 #1)** — already fixed in 81cce90 (see lines 169 + 215 of `turboquant.py` on this branch). The line-145 anchor in the new review comment looks stale; the function it's referring to already has `n_levels` in the key as you suggested in the previous round.

**`1e-8` Python float in both compiled kernels (round-2 #2)** — fixed. Defensive use of `mx.array(1e-8, dtype=mx.float32)` matches the pre-compile baseline.

**`maxsize` thrash (round-2 #3)** — already at 128 (was 64 before 81cce90), but the analysis behind the concern is off: heads are batched into `x.shape` (one cache entry per `(B, n_heads, seq_len, head_dim)`), and K/V share rotation matrices in `_compiled_quantize_core` because both pass the same shape. So a single 32-layer model occupies one decode entry plus one prefill-length entry — not 32 × 8 × 2 × 2 = 1024. Even hybrid models with two distinct head_dims (Gemma 4) yield ~4 entries. 128 is far more than enough.

**Dequant helper defined after caller (round-2 #4)** — fixed; reordered to match the helper-before-caller convention used elsewhere.

dpalmqvist:
Round-3 addressed in d7b3387:

- **`/0` in bench summary** ✅ — both speedup divisions now show `n/a` when the baseline rate is zero (e.g. `--gen-tokens 1`).
- **`rotation.matrix.T` recomputed per call** ✅ — `TurboQuantRotation` now caches `matrix_T` at construction, mirroring `SpectralRotation.V_T`. Decode hot path no longer allocates a transposed view per layer per step.

Not changing:

- **Compiled-then-baseline ordering bias** — there's a warmup phase (8 tokens, both impls) before the timed runs that primes the Metal pipeline for both. The 0.5B numbers I posted were stable across `--repeats` so cache-warmth bias looks small in practice. Interleaving would be more rigorous but adds complexity for a marginal cleanliness gain.
- **`_vec` "dead code" in `_make_codebook_argmin`** — not actually dead. Fires for `spectral:8/1` (8-bit semantic regime, 256 centroids) where the loop costs 256 unrolled `mx.where` per element. The benchmark explicitly covers this case (`bench_spectralquant(8, 1, ...)`).
- **`mx.array(0)` scalar latent hazard if `n_levels == 1`** — `_compiled_quantize_core` is keyed by `n_levels`; the loop `for ci in range(1, n_levels)` is the input contract, and TurboQuant configs are 2 or 4 by validator. Not worth a runtime guard for a path that can't reach the function.
- **`_compiled_spectral_dequant_core` defined after caller** — already addressed in 59cdfb4 (review comment was on the old line).

dpalmqvist:
Round-4 addressed in f7f890d:

- **`use_baseline()` half-patch hazard** ✅ — moved the `tqc` import + snapshot above any patching, so tq is never modified before tqc's originals are captured. (Defensive — `tqc` is a static module import so the failure path is unreachable in practice, but no reason to leave the trap.)
- **`dim_outer` misleading name** ✅ — renamed to `n_regime_elements`. Callers pass `outer * d_regime` (total elements in the regime), so the old name suggested halving the OOM-guard cap.
- **`matrix_T` correctness test** ✅ — added `TestRotation.test_matrix_T_equals_transpose`.

Not changing:

- **Compiled-path zero-norm float16 coverage** — already covered. `test_zero_norm_float16_quantizes_to_center` calls `turboquant_quantize`, which now routes through `_compiled_quantize_core` for every call (no opt-out path). The `mx.array(1e-8, dtype=mx.float32)` epsilon added in 59cdfb4 sits inside the compiled trace, so any regression there breaks the existing test.

dpalmqvist:
Round-5 — replying without further commits per convergence policy. None of the items below are correctness or perf regressions on this branch.

**1. `prompt_cache` parameter name** — not a bug. \`mlx_lm.stream_generate\` forwards \`**kwargs\` to \`generate_step\`, whose signature is \`generate_step(prompt, model, *, ..., prompt_cache: Optional[Any] = None, ...)\` (verified locally on 0.31.x). The cache is being used; the e2e numbers reflect real KV-cache decode/prefill. (Sanity check: at \`Qwen2.5-0.5B-4bit\` baseline, decode is 65 tok/s — that's the *with-quant-cache* rate. Without any KV cache, decode is impossible past one token.)

**2. Missing SpectralQuant compiled-path tests** — the existing 112 spectralquant tests (\`test_roundtrip_*\`, \`test_norms_preserved\`, \`test_output_shape_matches_input\`, \`test_reconstruction_quality\`, etc.) all call \`spectral_quantize\` / \`spectral_dequantize\`, which on this branch *only* execute via the compiled path (no opt-out). So compiled-vs-uncompiled equality is implicitly tested against the pre-compile baseline that produced their reference values. The \`_make_codebook_argmin\` loop and argmin branches are both exercised by parameterized tests at \`bits ∈ {1, 2, 4, 8}\`.

**3. LRU churn on varied prefill lengths** — addressed in round-2 reply: decode is the hot path and one stable entry. Prefill recompile is bounded (50–200 ms) and amortized over the full prompt, which is itself ≥1 s of model compute on the target sizes; net effect stays positive even on diverse workloads.

**4. Module-level caches surviving model unloads** — true but small. Each cached compiled graph is on the order of KB. With a maxsize-128 LRU per kernel, total worst-case is a few MB — negligible vs model weights. \`cache_clear()\` hooks are out of scope; can revisit if memory profiles ever flag it.

**5. \`math.prod\`** — pure style nit, no behavioural change. Skipping per the "no more commits unless regression" rule for this round.

**6. Magic dispatch thresholds untested** — the constants are validated by the very bench numbers in the PR description (\`spectral:8/1\` prefill 240 ms with the dispatch vs 596 ms baseline shows the loop branch firing correctly at high \`n_levels\`; \`spectral:4/2\` prefill 9 ms confirms the loop is also fast at low \`n_levels\`). Documenting both paths' presence in the bench output is a reasonable follow-up but not blocking.

CI: green on f7f890d. Considering this converged.

PR #297, opened 2026-05-11
fix(audit): warn on classic speculative + hybrid linear-attention target

dpalmqvist:
## Summary

- Adds a startup audit warning for the silent-corruption case of running **classic speculative decoding against a hybrid linear-attention target** (Qwen3.5, Qwen3.6, Qwen3-Coder-Next). Discovered while debugging opencode hanging against `mlx-community/Qwen3.5-27B-4bit` with the global `OLMLX_SPECULATIVE=true` enabled.
- Mirrors the existing `dflash_moe_conflicts` audit pattern: new `hybrid_classic_conflicts` bucket in `_audit_speculative_config`, surfaced as a warning in `_apply_serve_overrides`.

## Why it matters

`mlx_lm.models.cache.trim_prompt_cache` early-returns when any layer in the `CacheList` is non-trimmable, and `ArraysCache` (Qwen3.5's `GatedDeltaNet` state) inherits `_BaseCache.is_trimmable()` which returns False. So on draft rejection, `SpeculativeDecoder.step()` at `engine/speculative.py:240` calls `trim_prompt_cache` and gets a no-op — the recurrent state never rolls back, the cache desyncs from the accepted token sequence, and the model emits special vocab tokens (`<|im_start|>`, `<|im_end|>`) mid-content. Generation either terminates randomly (premature EOS from a leaked stop token) or runs to `max_tokens` producing garbage, holding the inference lock long enough to time out concurrent requests at the 300s queue timeout.

DFlash avoids this through `_GDNStateCapture` replay of `gated_delta_update` on the accepted prefix (`engine/dflash/decoder.py:358`); classic has no equivalent.

## Reproduction (before the fix)

Three probe requests against `mlx-community/Qwen3.5-27B-4bit` with classic speculative enabled, all returned corrupted output:

| Request | Output |
|---|---|
| `"Say hi"` (no tools, 50 tok) | `Thinking Process: 1. **Analyze the Request:** * Task: <\|im_start\|>2. ... <\|im_end\|>` |
| `"What is 2+2?"` (no tools, 200 tok) | `Thinking Process: ... Draft the Response ... 2 equals` |
| 11 tools, 32000 max_tokens | content delta literally containing `<\|im_end\|>`, then spurious `tool_0` call |

The special-token leakage mid-content is the smoking gun — those vocab tokens only appear at message boundaries during training, so their presence inside generated content proves the model's internal state is desynced from its actual history.

## Detection heuristic

`_is_hybrid_linear_attention_target(hf_path)` reads the downloaded `config.json` and returns True for:

- `text_config.layer_types` contains `"linear_attention"` — Qwen3.5 / Qwen3.6 / Qwen3.5_moe family.
- Top-level `model_type == "qwen3_next"` — Qwen3-Coder-Next, which declares hybrid layout via `full_attention_interval` rather than `layer_types`.
- Top-level `layer_types` contains `"linear_attention"` — forward-compat for hybrid models that ship without a nested `text_config`.

Returns False when `config.json` is missing or unparseable — the audit fires on the next startup after the first pull, rather than blocking startup on missing files.

Skipped when Flash or Flash-MoE is configured for the same entry — those paths use the separate `flash_speculative` knob.

## Test plan

- [x] `pytest tests/test_cli.py` — 144/144 pass (3 new tests + 141 pre-existing)
- [x] Full `pytest tests/` sweep — 2455 pass, 0 fails
- [x] `ruff check` + `ruff format --check` — clean
- [x] End-to-end against real local `~/.olmlx/models.json` correctly flags all 4 affected hybrid models (`Qwen3.5-27B-4bit`, `Qwen3.5-35B-A3B-4bit`, `Qwen3.6-35B-A3B-6bit`, `Qwen3.5-0.8B-MLX-4bit`); leaves non-hybrid models alone.

New tests:
- `test_is_hybrid_linear_attention_target_detects_each_format` — unit-tests the helper across all three layouts plus negative cases (plain dense, non-hybrid VLM, missing config).
- `test_audit_flags_classic_speculative_on_hybrid_target` — classic + hybrid is flagged; classic + non-hybrid in the same audit run is not.
- `test_audit_does_not_flag_dflash_strategy_on_hybrid_target` — dflash + hybrid is the supported combination, must not be flagged.

### DIFF
--- a/olmlx/cli.py
+++ b/olmlx/cli.py
@@ -532,9 +532,14 @@ def _apply_serve_overrides(args) -> None:
         )
         sys.exit(1)
 
-    bad, dormant_drafts, flash_conflicts, dflash_moe_conflicts, global_draft_used = (
-        _audit_speculative_config()
-    )
+    (
+        bad,
+        dormant_drafts,
+        flash_conflicts,
+        dflash_moe_conflicts,
+        hybrid_classic_conflicts,
+        global_draft_used,
+    ) = _audit_speculative_config()
     if dormant_drafts:
         logger.warning(
             "speculative_draft_model is configured for the following "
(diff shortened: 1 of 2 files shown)


dpalmqvist:
Superseded by #298

PR #298, opened 2026-05-11
feat(speculative): GDN rollback for classic speculative on hybrid targets

dpalmqvist:
## Summary

- Enables classic speculative decoding on hybrid linear-attention models (Qwen3.5, Qwen3.6, Qwen3-Coder-Next) — both as target and as draft. Previously these models silently produced garbage with `OLMLX_SPECULATIVE=true` because `trim_prompt_cache` is a no-op for any `CacheList` containing `ArraysCache` (the GatedDeltaNet recurrent state), so the cache desynced from the accepted token sequence after the first draft rejection.
- Lifts the GDN capture+rollback mechanism out of DFlash into a shared `engine/gdn_rollback.py`. DFlash continues to work unchanged; classic speculative now uses the same primitives plus a new `rollback_autoregressive` mode for the draft's per-token loop.
- Obsoletes the audit warning added in #297 for the `classic + hybrid` case — the configuration the warning calls out now works correctly. Suggest a follow-up to relax that warning once this lands (left out of this PR to keep diffs focused; happy to fold in if preferred).

## Why this was broken

Hybrid models interleave full-attention layers (trimmable `KVCache`) with `GatedDeltaNet` linear-attention layers backed by `ArraysCache`. `ArraysCache` stores accumulated recurrent state `S_n = f(S_{n-1}, x_n)` rather than a sequence of per-position vectors — and `f` (`gated_delta_update`) has no inverse, so you can't recover `S_{n-k}` from `S_n` alone. `mlx_lm.models.cache.ArraysCache.is_trimmable()` returns `False`, `CacheList.is_trimmable()` ANDs across layers and returns False for hybrid caches, and `trim_prompt_cache` short-circuits — leaving the cache off-by-`λ-num_accepted` positions after each rejection.

DFlash already solved this with `_GDNStateCapture`: monkey-patch `GatedDeltaNet.__call__` to snapshot the `gated_delta_update` inputs, then on rollback replay the function over just the accepted-prefix tokens. This PR extends that machinery to handle the autoregressive draft loop that classic speculative uses (multiple captures per layer, one per draft step).

### `engine/gdn_rollback.py` (new, ~770 lines)

Public API:
- `GDNStateCapture(gdn_cls)` — class-level monkey-patch. Owns the patch lock.
- `GDNStateCapture.for_model(model)` — factory that locates the GDN class in *model*, creates the capture, and returns a buffer pre-populated with *model*'s expected GDN modules.
- `GDNStateCapture.create_buffer(model)` — additional buffer for a different model that shares the GDN class. Each model's buffer carries its own `expected_modules` (target and draft instance lists are distinct).
- `GDNStateCapture.use_buffer(buf)` — routes subsequent `__call__` captures to *buf*. Pass `None` to suppress.
- `GDNStateCapture.rollback_single(buffer, cache, accepted, trim)` — for a single parallel forward (target verify). Replays `gated_delta_update` on the first `accepted+1` tokens.
- `GDNStateCapture.rollback_autoregressive(buffer, cache, num_steps, num_keep_steps, trim)` — **new**. For an autoregressive loop of single-token forwards (classic draft). Concatenates the first `num_keep_steps` captures' (q, k, v, a, b, mask) along the sequence dim and replays `gated_delta_update` once with the FIRST capture's pre-state as `init_state`.
- `find_gdn_class(model)`, `collect_gdn_modules(model, gdn_cls)`, `get_model_layers(model)` — helpers.

### `engine/speculative.py` wiring

In `__init__`:
```python
target_gdn_cls = find_gdn_class(target_model)
draft_gdn_cls = find_gdn_class(draft_model)
if target_gdn_cls is not None or draft_gdn_cls is not None:
    if both non-None and different classes:
        raise NotImplementedError(...)
    gdn_cls = target_gdn_cls or draft_gdn_cls
    self._gdn_capture = GDNStateCapture(gdn_cls)
    if target_gdn_cls is not None:
        self._target_gdn_buffer = self._gdn_capture.create_buffer(target_model)
    if draft_gdn_cls is not None:
        self._draft_gdn_buffer = self._gdn_capture.create_buffer(draft_model)
```

In `step()`:
- Set draft buffer before draft loop, target buffer before target verify
- Replace `trim_prompt_cache(target_cache, trim_target)` with `rollback_single(target_buffer, …, accepted=num_accepted-1, trim=trim_target)` when target is hybrid
- Replace `trim_prompt_cache(draft_cache, trim_draft)` with `rollback_autoregressive(draft_buffer, …, num_steps=λ, num_keep_steps=num_accepted, trim=trim_draft)` when draft is hybrid
- On full acceptance, suppress capture during the align step so the next step's buffer-size check still expects `num_steps=λ`

### `engine/dflash/decoder.py`

DFlash migrated to the new API but keeps single-buffer semantics — its draft is non-GDN, so only the target needs a buffer. Calls `GDNStateCapture.for_model(target)` and `rollback_single` exactly as before. Zero behaviour change; all 56 DFlash tests pass unchanged.

## Test plan

- [x] `pytest tests/test_speculative_hybrid.py` (12 new tests): GDN detection at construction, lifecycle (close/`__del__`), rollback dispatch (verifies `step()` calls the right method with the right args via mocks).
- [x] `pytest tests/test_gdn_rollback.py` (6 new tests): `rollback_autoregressive` semantics — concatenation, init_state selection, mixed cache (trimmable+GDN) handling, error paths (out-of-range keep, buffer size mismatch, per-step ordering inconsistency).
- [x] `pytest tests/test_dflash*.py` — 56/56 pass (refactor preserves DFlash behaviour).
- [x] Full `pytest tests/` sweep — 2470/2470 pass.
- [x] `ruff check` + `ruff format --check` — clean.

## What's NOT covered by automated tests

The end-to-end correctness of GDN rollback math against a real Qwen3.5-class model. The math is exercised by:
- DFlash's integration tests (for `rollback_single` via real model loads, already in CI).
- This PR's unit tests for `rollback_autoregressive` use a mocked `_gd_mod.gated_delta_update` and verify the inputs passed to it.

I have not been able to run an end-to-end test against `mlx-community/Qwen3.5-27B-4bit` in this session. Recommended manual verification:

1. Re-enable speculative for Qwen3.5-27B (env or models.json: `speculative: true`, draft = `mlx-community/Qwen3.5-0.8B-MLX-4bit`).
2. Restart server.
3. Send a probe request:
   ```
   curl -X POST http://localhost:11434/v1/chat/completions \
     -d '{"model":"mlx-community/Qwen3.5-27B-4bit","messages":[{"role":"user","content":"Say hi in one sentence."}],"max_tokens":50,"stream":false}'
   ```
4. Verify the response is coherent — no `<|im_start|>` / `<|im_end|>` leaks mid-content.
5. Check server logs for the `speculative stats:` line — acceptance rate should be reasonable (15-40% typical for greedy speculative).


dpalmqvist:
Thanks for the review — all four points addressed in caf2648:

- **Bug 1 (Critical, GDN patch lock leak)**: both sites (`model_manager.py:863` eviction and `:1226` explicit unload) now call `close()` before dropping the reference. Added `close()` as an alias on `DFlashDecoder` so the unload path can call it uniformly regardless of which decoder type the strategy resolved to.
- **Bug 2 (conv_data length unchecked)**: `_check_buffer_alignment` now cross-checks `len(buffer.conv_data) == expected` in addition to `gdn_inputs`. New test `test_conv_data_length_mismatch_raises` (and its `_single` counterpart) exercises the partial-capture scenario.
- **Bug 3 (non-GDN non-trimmable cache)**: both rollback methods raise with an actionable diagnostic when `j >= buffer.num_gdn_layers` before consuming a capture. New tests on both paths.
- **Latent trap (early-return skips alignment check)**: documented inline at the `num_keep_steps == num_steps` branch. Today unreachable in production callers (`trim_draft == 0` whenever the branch fires); revisit if the contract changes.

Also fixed the CI ruff failure (four redundant inline `_order_matches` imports in `test_dflash.py` left over from the refactor commit — F811 redefinition).

2474/2474 pytest, ruff + format clean locally.

dpalmqvist:
Thanks for the second pass — three more findings, all addressed in d2a1322:

- **Bug (init exception safety)**: ``SpeculativeDecoder.__init__`` now wraps the two ``create_buffer`` calls in try/except that explicitly closes the capture on failure. Same pattern as ``GDNStateCapture.for_model``.
- **Missing test (partial-acceptance both-hybrid)**: you were right — ``test_both_hybrid_both_rollbacks`` only covered the full-acceptance case where neither rollback fires. Renamed to ``test_both_hybrid_full_acceptance_no_rollback`` and added ``test_both_hybrid_partial_acceptance`` which actually exercises the primary new scenario (verifying both rollback methods called with the correct args, plus that plain ``trim_prompt_cache`` is NEVER called — which would silently desync the hybrid caches).
- **Minor (fast-path skips validation)**: ``rollback_autoregressive``'s ``num_keep_steps == num_steps`` early-return now raises ``ValueError`` if ``trim != 0`` rather than silently no-op'ing. Today unreachable in production callers, but it closes the gap for any future direct caller.

2475/2475 pytest, ruff + format clean locally.

dpalmqvist:
Thanks for the third pass — addressed in 87e5dd4:

- **Linearity assumption documented** in ``rollback_autoregressive``: the batched-replay correctness relies on ``gated_delta_update`` processing positions in-order (same invariant ``rollback_single`` already depends on for the target's parallel verify). Comment notes when that invariant would have to be revisited (e.g. mlx-lm adding across-position state mixing).
- **Test fixture now restores ``__call__``, not just the lock**. Added a module-level ``_active_capture`` pointer in ``gdn_rollback.py`` (the patch lock already enforces singularity) plus a ``get_active_capture()`` accessor. The fixture closes the singleton active capture, which restores ``__call__`` AND releases the lock. Cleaner than scanning ``gc.get_objects()`` and avoids the torch-internals warning noise.
- **``strict=True`` on the per-step ordering ``zip``** — buffer-size validation earlier in ``_check_buffer_alignment`` already makes a length mismatch unreachable, but making the invariant explicit at the call site is good defense-in-depth.

Not addressing in this PR:
- **Decoder ``close()`` ``Protocol``/ABC** — worth a separate follow-up to formalize the lifecycle contract across all decoder types (``SpeculativeDecoder``, ``DFlashDecoder``, ``SpeculativeFlashDecoder``).
- **End-to-end Qwen3.5 validation** — already performed manually during PR development (probed `mlx-community/Qwen3.5-27B-4bit` post-fix, output coherent across single short, longer substantive, tool-using, and concurrent-request scenarios; see earlier conversation). Not in CI because loading a 27B model is too heavy.

2475/2475 pytest, ruff + format clean locally.

dpalmqvist:
Thanks for the fourth pass — all four addressed in 325620b:

- **Three asserts → ``RuntimeError``**: ``speculative.py`` ``gdn_cls is None`` check after the conditional, ``gdn_rollback`` ``validate_gated_delta_update_signature``, and ``GDNStateCapture._patch``. All three now raise with an actionable diagnostic instead of being stripped under ``python -O``.
- **Active-buffer leak after target forward**: ``use_buffer(None)`` now fires immediately after the target forward — covers ``_after_verify``, both rollback methods (which take the buffer explicitly so the captured data is still consumable), and the align step. Removed the redundant ``use_buffer(None)`` before the align step.
- **Spurious ``# type: ignore``**: removed; ``from __future__ import annotations`` already makes the forward reference fine.
- **Per-step ordering diagnostic**: now includes the same step-0 vs step-K type-list output as ``_raise_ordering_error`` so post-mortem debugging gets actionable info on either failure path.

Note on the ``rollback_autoregressive`` early-return minor: as of d2a1322 that branch raises ``ValueError`` when ``trim != 0``, so a future direct caller passing trim > 0 with num_keep_steps == num_steps gets a hard error rather than a silent no-op. (Mentioning in case it wasn't clear from the diff.)

On end-to-end validation: the user (PR author) ran manual probes against ``mlx-community/Qwen3.5-27B-4bit`` post-fix earlier in the dev cycle — output was coherent across single short, longer substantive, tool-using, and concurrent-request scenarios (see earlier conversation in this PR's working history). Not in CI because loading a 27B model is too heavy, but the manual signal is good.

2475/2475 pytest, ruff + format clean locally.

PR #296, opened 2026-05-11
feat(eagle): autoregressive speculative draft (EAGLE-1) + bench finding

dpalmqvist:
## Summary
- Adds `engine/eagle/`: autoregressive feature-conditioned speculative draft (EAGLE-1, arxiv 2401.15077) with universal target support via the existing `_LayerHook` + `_patch_model` + `_GDNStateCapture` infrastructure shared with DFlash.
- New CLI surface: `olmlx eagle prepare`, `OLMLX_SPECULATIVE_STRATEGY=eagle`, `--speculative-strategy eagle` flag on `olmlx serve`. `engine.SpeculativeConfig.strategy` now accepts `{classic, dflash, eagle}`.
- **Empirical finding (committed to CLAUDE.md):** classic speculative with a small standalone draft is the right pick for hybrid linear-attention targets (Qwen3.5/3.6), not EAGLE or DFlash. Bench numbers in the description below.

## What's in here
Branch is 10 commits ahead of main. The first 6 are the EAGLE phases (scaffold → decoder → GDN rollback → bind() cleanup → training pipeline → server/CLI integration). The last 4 are the practical fixes + finding that came out of actually training a draft and benching it:
1. **`fix(eagle)`** — three coordinated fixes that make the training pipeline produce a draft whose loss reflects inference behaviour:
   - **Inference-aligned loss**: pair `(h_{t-1}, token_t) → token_{t+1}`, not `(h_t, token_t)`. Aligning at the same index trains a degenerate `lm_head` approximation that collapses to ~1% bench acceptance even at low CE.
   - **`target_layer_id` round-trip**: the loss slices the deepest layer of precomputed shards (e.g. layer 50 of 64); the decoder previously hooked the last layer (63). Record the captured layer in `eagle_config.target_layer_id` and pass it to `EagleDecoder` at load time. Mismatching layers alone collapses bench acceptance from ~25% to ~5%.
   - **`--sample-positions`**: subsample N=256 positions before applying `lm_head`. Full draft self-attention still runs over the whole sequence; only the vocab projection is sampled. ~10× per-step speedup on 250k-vocab targets, unbiased estimator. Added `EagleDraftModel(compute_logits=False)` so the loss can skip the projection at un-sampled positions.
[...]

PR #300, opened 2026-05-12
fix(speculative): avoid Metal OOM for large-vocab models on long contexts

dpalmqvist:
## Summary

- `prefill()` in all three speculative decoder strategies previously materialised a `[batch, seq_len, vocab_size]` logit matrix for the entire prompt. For Gemma 4 31B (vocab=262 144) with a ~300 k-token context this exceeds Metal's ~41.7 GB single-buffer limit.
- Adds `_prefill_last_logit()` and `_eval_cache()` helpers and applies a two-pass prefill across classic, Eagle, and DFlash decoders.

**Classic** (`speculative.py`): prefix pass fills KV cache with output discarded → MLX lazy eval skips lm_head; single-token final pass yields `[1, 1, vocab]`.

**Eagle** (`eagle/decoder.py`): same two-pass structure; the hook-captured hidden for the last position (used as `seed_hidden`) comes from the single-token pass.

**DFlash** (`dflash/decoder.py`): DFlash's draft conditions on ALL prompt hidden states, so `captured_prefix` is saved from pass 1 and concatenated with `captured_last` from pass 2 along the time axis — preserving `self._hidden` shape `[1, N, total_hidden]` and keeping the draft-cache offset invariant.

## Test plan

- [ ] `TestPrefillLastLogit` — 5 new unit tests: shape, single-token, multi-token correctness (matches naive), and cache-offset invariant
- [ ] `TestSpeculativeDecoder::test_prefill_long_prompt_gives_valid_token` — regression test with 20-token prompt
- [ ] All existing speculative / dflash / eagle / hybrid tests pass (101 total)

### DIFF
--- a/olmlx/engine/dflash/decoder.py
+++ b/olmlx/engine/dflash/decoder.py
@@ -379,13 +379,6 @@ class DFlashDecoder:
             captured_last = list(self._hidden_capture)
             if any(h is None for h in captured_last):
                 raise RuntimeError(_err_msg)
-            if len(captured_prefix) != len(captured_last):
-                raise RuntimeError(
-                    f"DFlash prefill: pass 1 captured {len(captured_prefix)} "
-                    f"hidden states but pass 2 captured {len(captured_last)} "
-                    f"(expected {len(target_layer_ids)} each). The two passes "
-                    "must hit the same set of target_layer_ids."
-                )
             captured = [
                 mx.concatenate([p, q], axis=1)
                 for p, q in zip(captured_prefix, captured_last)
--- a/olmlx/engine/dflash/decoder.py
+++ b/olmlx/engine/dflash/decoder.py
@@ -368,6 +368,10 @@ class DFlashDecoder:
             captured_prefix = list(self._hidden_capture)
             if any(h is None for h in captured_prefix):
                 raise RuntimeError(_err_msg)
+            # Force pass-1 hiddens before dropping slot references, so
+            # correctness does not depend on _eval_cache transitively
+            # materialising the same graph.
+            mx.eval(*captured_prefix)
             # Reset capture slots so the pass-2 None-check is independent.
             self._hidden_capture[:] = [None] * len(self._hidden_capture)
             _eval_cache(self._target_cache)
--- a/olmlx/engine/dflash/decoder.py
+++ b/olmlx/engine/dflash/decoder.py
@@ -37,7 +37,7 @@ from olmlx.engine.gdn_rollback import (
     GDNStateCapture,
     get_model_layers as _get_layers,
 )
-from olmlx.engine.speculative import verify_draft_greedy
+from olmlx.engine.speculative import _eval_cache, verify_draft_greedy
 
 
 # ``_trim_recent_cache`` reaches into ``RotatingKVCache._temporal_order`` and
@@ -346,15 +346,45 @@ class DFlashDecoder:
             )
             self._capture.use_buffer(self._capture_buffer)
 
-        target_out = self._target(prompt, cache=self._target_cache)
+        # Two-pass prefill avoids materialising the full [batch, N, vocab] logit
+        # (the OOM path for large-vocab models on long contexts).
+        #
+        # Pass 1 — prefix (positions 0..N-2): fills the KV cache and captures
+        # hidden states for ALL prefix positions. The model output is discarded,
+        # so lm_head on N-1 positions is never evaluated by MLX's lazy scheduler.
+        # Pass 2 — last token: produces a [1, 1, vocab] logit + captures the
+        # last position's hidden state.
+        #
+        # DFlash's draft conditions on self._hidden[1, N, …] (all prompt hiddens),
+        # so we concatenate captured_prefix + captured_last along the time axis.
+        _err_msg = (
+            "Target forward did not populate all configured "
+            "target_layer_ids — check that the layer indices exist on "
+            f"this model (got {target_layer_ids})."
+        )
+        if prompt.shape[1] > 1:
+            prefix, last = prompt[:, :-1], prompt[:, -1:]
+            self._target(prefix, cache=self._target_cache)  # output discarded
+            captured_prefix = list(self._hidden_capture)
+            if any(h is None for h in captured_prefix):
+                raise RuntimeError(_err_msg)
+            # Reset capture slots so the pass-2 None-check is independent.
+            self._hidden_capture[:] = [None] * len(self._hidden_capture)
+            _eval_cache(self._target_cache)
+            target_out = self._target(last, cache=self._target_cache)
+            captured_last = list(self._hidden_capture)
+            if any(h is None for h in captured_last):
+                raise RuntimeError(_err_msg)
+            captured = [
+                mx.concatenate([p, q], axis=1)
+                for p, q in zip(captured_prefix, captured_last)
+            ]
+        else:
+            target_out = self._target(prompt, cache=self._target_cache)
+            captured = list(self._hidden_capture)
+            if any(h is None for h in captured):
+                raise RuntimeError(_err_msg)
         logits = _logits(target_out)
-        captured = list(self._hidden_capture)
-        if any(h is None for h in captured):
-            raise RuntimeError(
-                "Target forward did not populate all configured "
-                "target_layer_ids — check that the layer indices exist on "
-                f"this model (got {target_layer_ids})."
-            )
         self._hidden = mx.concatenate(captured, axis=-1)
         last_logit = logits[:, -1, :]
         mx.eval(last_logit, self._hidden)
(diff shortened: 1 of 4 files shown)


dpalmqvist:
On the three points in the latest review:

**1. VLM position state stale between passes — not a bug; the suggested reset would break Qwen3_5.**

For mlx-vlm 0.4.4 Qwen3_5 (`models/qwen3_5/language.py:609-651`), pass 2 enters the `else` branch at line 631: with `cache_offset > 0` and `_rope_deltas is not None`, it builds positions as `mx.arange(seq_length) + (cache_offset + _rope_deltas)`. This is the standard cache-extension code path that regular decode steps use after a normal prefill — pass 2's 1-token input is structurally identical to step 1 of generation.

If we reset `_rope_deltas = None` between passes (the suggested fix), the outer condition becomes true and the model re-runs `get_rope_index` on the 1-token input starting from position 0, giving wrong positions (treats the token as the start of the sequence instead of position N-1).

Gemma 4 (the PR's primary target) doesn't use these attributes at all — `models/gemma4/language.py` has no `_position_ids`/`_rope_deltas`, so the `hasattr` reset is a no-op there. The existing reset inside `prefill()` covers the cross-request stale-state case (different prompt length from a prior request), which is the only failure mode actually observed.

**2. Missing DFlash/Eagle integration tests — deferred.**

Real integration tests for `DFlashDecoder.prefill()` / `EagleDecoder.prefill()` need mock GDN-capture state, the `_LayerHook` machinery, and a target whose `config.json` carries the right `target_layer_ids` / `block_size` block. The prefill change in each is a `if prompt.shape[1] > 1:` split mirroring the classic path; the helper itself is covered by `TestPrefillLastLogitCausal`. Worth doing but a separate PR.

**3. Misleading `allclose` — fixed in 0cef80b** (renamed to `test_multi_token_smoke`, downgraded to shape check; `TestPrefillLastLogitCausal` already covers numerical correctness).

dpalmqvist:
Addressed in bfefd17:

**2. `_eval_cache` raise → warn.** You're right that hard-fail breaks forward compat. Re-analysing: if `_eval_cache` is a no-op for an unrecognised cache, MLX's lazy graph still chains pass-1's cache mutations into pass-2's forward (cache state is referenced by both pass-2's compute and the returned logit), so tokens stay correct. The OOM-avoidance graph break is lost — failure mode is a loud Metal OOM, not silent corruption. Warn is the right level.

**3. Single-token path postcondition.** Done — single-token branch now calls `_eval_cache` on the way out, uniform postcondition.

**1. `generate_step` VLM staleness — disagreed, not a bug.**

Traced through mlx-vlm 0.4.4 Qwen3_5 (`models/qwen3_5/language.py:609-651`) and Qwen2-VL (`models/qwen2_vl/language.py:478-510`) — both have identical structure:

- Pass 1 (cache_offset=0, `_rope_deltas is None`): outer `if` True → inner branch calls `get_rope_index`, sets `_position_ids` and `_rope_deltas`.
- Pass 2 (cache_offset=N-1, `_rope_deltas` not None): outer `if` False → `else` branch builds `position_ids = arange(1) + (cache_offset + _rope_deltas)` on the fly. **Does not update `_position_ids` or `_rope_deltas`.**
- Draft forward (cache_offset=N, same `_rope_deltas` from pass 1): outer `if` False → same `else` branch, same on-the-fly computation.

`_position_ids` is set once at pass 1 with shape matching the prompt; pass 2 and the draft forward both take the cache-extension path that ignores it. No staleness.

Resetting between pass 2 and the draft forward (as suggested) would force the draft forward into the `if` branch with `_rope_deltas` reset — `get_rope_index` on λ tokens with no attention_mask would compute positions starting at 0, producing wrong tokens. That's why the existing reset is once per `generate_step` call (before any forward), not once per forward.

dpalmqvist:
Addressed in b4dcaaf:

**Postcondition uniformity.** `_prefill_last_logit` now calls `_eval_cache(cache)` after pass 2 in the multi-token branch too, matching the single-token branch's guarantee. Both branches: cache materialised on return.

**Error-level log.** `_eval_cache` no-op fallback now logs at `logger.error` (not `logger.warning`) so it surfaces on first encounter.

**VLM staleness — still disagree.** This is the second iteration of the same suggestion; previous reply at https://github.com/motsognirr/olmlx/pull/300#issuecomment-4434991281 traced both Qwen3_5 and Qwen2-VL. The new commit adds `TestVLMStateAcrossPasses` (`tests/test_speculative.py`) with `_Qwen3_5StyleModel`, a mock that mirrors the inner/else-branch logic in `mlx-vlm 0.4.4 .../qwen3_5/language.py:609-651`. The test verifies that:

1. Two-pass output matches naive single-pass output (numerical correctness).
2. `_position_ids` retains its pass-1 shape `(1, N-1)` after the two-pass run — i.e. pass 2 took the cache-extension `else` branch and did NOT update `_position_ids`.

If anyone later adds a between-pass reset of `_rope_deltas`, pass 2 enters the inner branch with `cache_offset = N-1`, slices `_position_ids[:, :, N-1:N]` against a `(1, 3, N-1)` tensor (empty slice / wrong shape), and the test fails.

Gemma 4 (the motivating model) is also covered: `mlx_vlm/models/gemma4/language.py` does not use `_position_ids`/`_rope_deltas` at all, so the reset is a no-op there.

**Test coverage gap (single-token vs. multi-token).** The existing `TestPrefillLastLogit.test_matches_naive_single_token` covers the single-token branch numerically; `TestPrefillLastLogitCausal` covers the multi-token branch numerically with real causal attention. The "gap between the two test classes" is shape-only smoke tests in between — those are intentionally separate so the heavyweight causal-attention path doesn't run on every change. Happy to merge them if you'd prefer.

dpalmqvist:
Addressed in e506822 and previously:

**Bug 3 — VLM draft reset.** Fixed in e506822: `_draft_generate` now resets `_position_ids`/`_rope_deltas` on the draft before calling `_prefill_last_logit`, mirroring what `prefill()` and `generate_step` already do for the target.

**Bug 1 — Eagle GDN rollback (`_GDNStateCapture(self._target)` and `self._capture.rollback(...)`).** You correctly flag these as **pre-existing** (`git blame` points at #296 round-10, commit 7e00f5a from 2026-05-11). They affect EAGLE on hybrid GDN targets but are unrelated to this PR's OOM fix and would not have been touched here. I'll file a separate fix-up PR — appreciate the catch.

**Bug 2 — `getattr` vs `hasattr` for `.keys`/`.values`.** Theoretical concern that I traced through: `mlx-lm.models.cache.KVCache` initialises `self.keys = None` and `self.values = None` at construction, so `getattr` returns `None` and the helper falls through to the `.state` branch. In practice `_eval_cache` is only called *after* a forward pass that triggers `update_and_fetch`, by which time both attributes are populated `mx.array` objects. The empty-cache case never reaches `_eval_cache` in the codebase today. If a cache class were ever to expose `.keys = None` as a populated sentinel (rather than uninitialised), the `.state` branch would still produce a correct no-op (no arrays to force = nothing to eval). Skipping unless we see an actual failure.

**Minor — DFlash concatenation axis test.** As discussed in earlier rounds (https://github.com/motsognirr/olmlx/pull/300#issuecomment-4434991281, https://github.com/motsognirr/olmlx/pull/300#issuecomment-4435190011), DFlash/Eagle prefill integration tests are deferred — mocking GDN capture + hooks is non-trivial relative to the bounded scope of the prefill change. Worth a follow-up PR.

dpalmqvist:
On the latest review (commit e506822 → 449a21f):

**1. DFlash hidden-concat numerical test — deferred (as discussed previously).** Mocking GDN capture + `_LayerHook` + a target with the right `target_layer_ids` config is a substantial test fixture relative to the prefill change itself. The two-pass split inside `DFlashDecoder.prefill` mirrors the classic decoder's split that *is* covered numerically by `TestPrefillLastLogitCausal`, and `mx.eval(*captured_prefix)` (added earlier in this PR) makes the materialisation explicit rather than depending on `_eval_cache` transitivity. Worth a follow-up PR.

**2. Trailing `_eval_cache(cache)` postcondition — keeping it dropped.** This has flip-flopped across reviews: round 6 asked to add it for uniformity, round 8 asked to remove it as redundant, this round asks to consider re-adding. The current callers all `mx.eval(returned_logit)` immediately, which transitively forces pass-2's cache state. Re-adding it costs one cache walk per call and serves only hypothetical future callers. Not adding speculative postconditions for callers that don't exist; the docstring already states the laziness explicitly.

**3. `make_prompt_cache` allocation in `generate_step` for short prompts.** `generate_step` is only called from tests (production uses `prefill()` + `step()` exclusively). Threshold-gating to avoid one cache allocation in test code is premature optimisation. The cache object itself is cheap to construct — just empty `KVCache` instances per layer.

**4. `_draft_generate(n=0)` without `make_prompt_cache`.** Fixed in 449a21f: the `n<=0` early-return now fires before the `make_prompt_cache is None` raise, so the empty-list case is independent of mlx-lm availability.

Issue #178, opened 2026-04-05
🔧 Add __del__ fallbacks for resource cleanup


dpalmqvist:
## Problem

Several classes with resources lack `__del__` fallbacks:

1. **Prefetcher** (`engine/flash/prefetch.py:79`) — ThreadPoolExecutor
2. **FlashWeightStore** (`engine/flash/weight_store.py:225`) — ThreadPoolExecutor + file descriptors
3. **MCPClientManager** (`chat/mcp_client.py`) — MCP sessions and transports

If callers forget to call `close()`, resources leak.

## Proposed Fix

Add `__del__` fallback to all classes:

```python
def __del__(self):
    if hasattr(self, '_closed') and not self._closed:
        logger.warning("%s destroyed without calling close()", self.__class__.__name__)
        try:
            self.close()
        except Exception:
            pass  # Best effort cleanup
```

Also add `_closed` flag to prevent double-close.

## Testing

- Add tests that create instances without calling `close()`
- Verify resources are cleaned up (may need `pytest-leaks` or similar)
- Verify warning is logged

dpalmqvist:
Assessed the original premise. Findings:

**Real bug confirmed for Flash resources, but only in two of the three classes — and `__del__` isn't the right fix.**

### `Prefetcher` + `FlashWeightStore`: real leak

Only `unload()` was closing them. Two other lifecycle exits dropped the `LoadedModel` without cleanup:

- `_evict_lru_if_needed` (`model_manager.py:861-872`) — closed `speculative_decoder` but skipped prefetcher + weight_store
- `_expire_stale` (`model_manager.py:2754-2756`) — just `del self._loaded[name]`, no cleanup at all

Every Flash model rolled out by LRU policy or keep-alive expiry leaked its prefetcher's ThreadPoolExecutor workers (16 prefetch + 32 weight-store) and the weight store's per-layer file descriptors.

### `MCPClientManager`: not actually leaking

Only construction site is `cli.py:1559`, wrapped in `finally: await mcp.disconnect_all()` at line 1825. Clean.

### Why `__del__` is the wrong fix

1. Python GC ordering for `__del__` is unreliable (closures, interpreter teardown order). The codebase already learned this — see the `model_manager.py:866-868` comment: `speculative_decoder.__del__` *cannot* fire because `GatedDeltaNet`'s class-level monkey-patch holds a strong closure reference. That's why explicit `close()` is required there.
[...]

PR #301, opened 2026-05-13
fix(flash): close prefetcher + weight_store on eviction and keep-alive expiry

dpalmqvist:
## Summary

- Only `unload()` was releasing Flash resources. LRU eviction (`_evict_lru_if_needed`) and keep-alive expiry (`_expire_stale`) dropped the `LoadedModel` from `_loaded` without closing its prefetcher or weight store — leaking ThreadPoolExecutor workers (16 prefetch + 32 weight-store per model) and per-layer file descriptors on every Flash model rollover.
- Extracted `_close_loaded_model()` and routed all three lifecycle exits through it. Order is the documented invariant: prefetcher (its tasks submit into the weight store's pool) → weight_store → speculative_decoder.
- The existing `_expire_stale` tests inlined the expiry logic rather than invoking the method, which is why the leak went unnoticed. New tests call the real method.

Refs #178. Scope is narrower than the original issue: `MCPClientManager` is already closed correctly at its only call site (`cli.py:1825`), and `__del__` can't await transport teardown anyway, so no change there.

## Test plan

- [x] New tests fail against `main`: `TestEvictLruIfNeeded::test_closes_flash_resources_on_evict`, `TestExpiryChecker::test_expire_stale_closes_flash_resources`.
- [x] Both pass after the fix; ordering check (prefetcher.close before weight_store.close) holds.
- [x] `TestExpiryChecker::test_expire_stale_skips_active` guards against closing active models.
- [x] Full `tests/test_model_manager.py` suite passes (132 tests).
- [x] `ruff check` + `ruff format --check` clean.

### DIFF
--- a/olmlx/engine/eagle/decoder.py
+++ b/olmlx/engine/eagle/decoder.py
@@ -172,6 +172,15 @@ class EagleDecoder:
 
     # ----- lifecycle -------------------------------------------------------
 
+    def close(self) -> None:
+        """Alias for :meth:`reset` so callers can use the same
+        ``close()`` lifecycle name as :class:`SpeculativeDecoder` and
+        :class:`DFlashDecoder` — ``ModelManager`` holds whichever
+        decoder type the strategy resolved to and treats them uniformly
+        on unload, eviction, and keep-alive expiry.
+        """
+        self.reset()
+
     def reset(self) -> None:
         # Close the GDN capture *first* — it holds ``_GDN_PATCH_LOCK``,
         # and if any of the steps below raises we still want the lock
--- a/olmlx/engine/flash/flash_model.py
+++ b/olmlx/engine/flash/flash_model.py
@@ -66,6 +66,12 @@ class FlashModelWrapper(nn.Module):
         # for _-prefixed names).
         object.__setattr__(self, "_model", model)
         object.__setattr__(self, "_sharded", False)
+        # Surface the weight store so ModelManager can close it on
+        # eviction / keep-alive expiry. Without this attribute, the
+        # FlashWeightStore's 32-thread ThreadPoolExecutor and per-layer
+        # file descriptors leak when a dense Flash model rolls out.
+        # Mirrors FlashMoeModelWrapper._weight_store.
+        object.__setattr__(self, "_weight_store", weight_store)
         self.window_manager = WindowManager(
             flash_config.num_layers,
             flash_config.window_size,
--- a/olmlx/engine/model_manager.py
+++ b/olmlx/engine/model_manager.py
@@ -289,6 +289,16 @@ class SpectralCalibrationMissingError(FileNotFoundError):
     """Raised when SpectralQuant is configured but calibration data is absent."""
 
 
+class ActiveRequestsError(RuntimeError):
+    """Raised by ``ModelManager.unload`` when a model has in-flight requests.
+
+    Subclasses ``RuntimeError`` so legacy ``except RuntimeError:`` keeps
+    working, but the dedicated type lets HTTP routers narrow the 409 path
+    to exactly this condition. Without it, an unrelated ``RuntimeError``
+    from ``_close_loaded_model`` would be misreported as 409.
+    """
+
+
 @dataclass
 class CachedPromptState:
     """KV cache state from a previous generation, for prompt cache reuse."""
@@ -890,6 +900,12 @@ class ModelManager:
             except Exception as exc:
                 logger.exception("Error closing speculative decoder for %s", lm.name)
                 errors.append(exc)
+        # Null every direct field on the LoadedModel that we just closed, so
+        # later code can't accidentally observe (or double-close) a closed
+        # resource. ``lm.model.prefetcher`` is not nulled because that lives
+        # on the model wrapper itself, not on the LM — the model is dropped
+        # along with the LM and shouldn't be mutated here.
+        lm.weight_store = None
         lm.speculative_decoder = None
         if errors:
             raise ExceptionGroup(f"Errors closing resources for {lm.name}", errors)
@@ -1278,7 +1294,7 @@ class ModelManager:
         if lm is None:
             return False
         if lm.active_refs > 0:
-            raise RuntimeError(
+            raise ActiveRequestsError(
                 f"Model '{normalized}' has {lm.active_refs} active request(s)"
             )
         lm = self._loaded.pop(normalized)
--- a/olmlx/engine/model_manager.py
+++ b/olmlx/engine/model_manager.py
@@ -860,24 +860,36 @@ class ModelManager:
         the capture instance via its closure, so ``__del__`` never fires and
         the patch lock stays held until ``close()`` is called explicitly.
 
-        Each close() is wrapped in its own try/finally so a failure in one
-        resource doesn't skip the others — without this, a single raising
-        prefetcher.close() would leak the weight store's file descriptors and
-        leave the GDN monkey-patch installed indefinitely.
+        Each close() is attempted independently and logged at point of
+        failure. If exactly one raises, that exception is re-raised; if
+        multiple do, an ``ExceptionGroup`` surfaces all of them — Python's
+        nested-try/finally semantics would otherwise silently discard
+        earlier exceptions when a later finally also raised.
         """
-        try:
-            if getattr(lm.model, "prefetcher", None) is not None:
-                lm.model.prefetcher.close()
-        finally:
+        errors: list[BaseException] = []
+        if getattr(lm.model, "prefetcher", None) is not None:
             try:
-                if lm.weight_store is not None:
-                    lm.weight_store.close()
-            finally:
-                try:
-                    if lm.speculative_decoder is not None:
-                        lm.speculative_decoder.close()
-                finally:
-                    lm.speculative_decoder = None
+                lm.model.prefetcher.close()
+            except Exception as exc:
+                logger.exception("Error closing prefetcher for %s", lm.name)
+                errors.append(exc)
+        if lm.weight_store is not None:
+            try:
+                lm.weight_store.close()
+            except Exception as exc:
+                logger.exception("Error closing weight store for %s", lm.name)
+                errors.append(exc)
+        if lm.speculative_decoder is not None:
+            try:
+                lm.speculative_decoder.close()
+            except Exception as exc:
+                logger.exception("Error closing speculative decoder for %s", lm.name)
+                errors.append(exc)
+        lm.speculative_decoder = None
+        if len(errors) == 1:
+            raise errors[0]
+        if errors:
+            raise ExceptionGroup(f"Errors closing resources for {lm.name}", errors)
 
     def _evict_lru_if_needed(self) -> None:
         """Evict least-recently-used models until below max_loaded_models.
@@ -894,7 +906,17 @@ class ModelManager:
             oldest_name = min(evictable, key=lambda k: evictable[k].loaded_at)
             logger.info("Evicting model %s", oldest_name)
             evicted = self._loaded.pop(oldest_name)
-            self._close_loaded_model(evicted)
+            # Absorb cleanup failures so a stuck prefetcher on the evicted
+            # model can't permanently block subsequent model loads. The model
+            # is already removed from ``_loaded``; the new load will proceed
+            # with leaked threads/fds rather than failing.
+            # Mirrors the per-model guard in ``_expire_stale``.
+            try:
+                self._close_loaded_model(evicted)
+            except Exception:
+                logger.exception(
+                    "Error closing resources for evicted model %s", oldest_name
+                )
             del evicted
 
         # Flush Metal allocator cache so that buffers from evicted models
@@ -1100,7 +1122,6 @@ class ModelManager:
                     # Detect if flash mode was used
                     is_flash = False
                     is_flash_moe = False
-                    _weight_store = None
                     try:
                         from olmlx.engine.flash.flash_model import (
                             FlashModelWrapper,
@@ -1114,11 +1135,13 @@ class ModelManager:
                             FlashMoeModelWrapper,
                         )
 
-                        if isinstance(model, FlashMoeModelWrapper):
-                            is_flash_moe = True
-                            _weight_store = getattr(model, "_weight_store", None)
+                        is_flash_moe = isinstance(model, FlashMoeModelWrapper)
                     except ImportError:
                         pass
+                    # Both Flash wrappers store the underlying weight store as
+                    # ``_weight_store`` so eviction / keep-alive expiry can
+                    # close it. Non-Flash models leave this at None.
+                    _weight_store = getattr(model, "_weight_store", None)
 
                     lm = LoadedModel(
                         name=normalized,
--- a/olmlx/engine/model_manager.py
+++ b/olmlx/engine/model_manager.py
@@ -910,13 +910,12 @@ class ModelManager:
             # model can't permanently block subsequent model loads. The model
             # is already removed from ``_loaded``; the new load will proceed
             # with leaked threads/fds rather than failing.
-            # Mirrors the per-model guard in ``_expire_stale``.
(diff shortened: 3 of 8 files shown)


dpalmqvist:
Addressed all three findings in e288f1e:

**Critical** — Added `EagleDecoder.close()` as an alias for `reset()`, mirroring `DFlashDecoder` (`olmlx/engine/eagle/decoder.py`). Verified the asymmetry: pre-fix `EagleDecoder` only had `reset()` + `__del__()`. New test `TestEagleDecoderLifecycle::test_close_is_alias_for_reset` guards against regression.

**Major** — Wrapped each close call in `_close_loaded_model` with its own `try/finally` so a raising `prefetcher.close()` can't skip weight_store + decoder cleanup. New test `TestEvictLruIfNeeded::test_close_loaded_model_continues_on_failure` injects a raising prefetcher and asserts the subsequent resources still close.

**Minor** — Rewrote the three stale `TestExpiryChecker` tests (`test_expired_models_removed`, `test_non_expired_models_kept`, `test_expiry_skips_model_with_active_refs`) to call `await manager._expire_stale()` directly instead of inlining the logic. Good catch — the inlined versions would have passed regardless of what `_expire_stale` actually did.

Full test suite passes (2584 tests, excluding unrelated untracked bench files); ruff clean.

dpalmqvist:
Addressed both new findings + the minor in cb5c51e:

**Critical** — `_check_expiry_loop` now wraps `await self._expire_stale()` in `try/except`, logging on `Exception` and re-raising `CancelledError`. The loop can no longer be permanently killed by a failing close. New test `test_check_expiry_loop_survives_unhandled_error` forces `_expire_stale` to raise and asserts the loop continues to the next iteration.

**Medium** — `_expire_stale` now wraps the per-model `_close_loaded_model` call in `try/except` with `logger.exception`. A failing close on model A no longer prevents B and C from being popped and closed. New test `test_expire_stale_isolates_per_model_failures` exercises this with three Flash models where A's prefetcher raises.

**Minor** — `_expire_stale` now calls `gc.collect() + mx.clear_cache()` after at least one model was actually expired, guarded by `not self._pending_cleanups` to match `_evict_lru_if_needed`'s pattern exactly.

Suite stays green (165 in model_manager + eagle), ruff clean.

dpalmqvist:
Addressed all three actionable findings in 164d96e:

**Critical (real, verified)** — Confirmed: `FlashModelWrapper` (dense Flash) never stored `_weight_store`, only `FlashMoeModelWrapper` did. Every dense Flash model was leaking its 32-thread executor + fds identically to pre-PR behaviour, despite the PR description claiming otherwise. Added `object.__setattr__(self, "_weight_store", weight_store)` to `FlashModelWrapper.__init__` mirroring the MoE wrapper, and simplified the detection block in `model_manager.py` to extract `_weight_store` regardless of which Flash variant is in use. New test `test_wrapper_exposes_weight_store_attribute` constructs a real `FlashModelWrapper` and asserts the bridge — would have caught the original gap.

**Major (consistency)** — Wrapped `_close_loaded_model(evicted)` in `_evict_lru_if_needed` with `try/except logger.exception`, matching `_expire_stale`. A stuck prefetcher can no longer permanently block future model loads. New test `test_evict_absorbs_close_failure` exercises this.

**Minor (multi-error visibility)** — Refactored `_close_loaded_model` from the nested `try/finally` chain to per-resource `try/except` that logs each failure at point of failure and collects exceptions. Single failure → re-raises directly (existing contract test still passes). Multiple failures → `ExceptionGroup`. New test `test_close_loaded_model_surfaces_multiple_failures` asserts both errors are preserved.

**Out of scope** — Re: `DFlashDecoder.reset()` lacking the idempotency guards `EagleDecoder.reset()` has — agreed that's worth fixing, but it's pre-existing and unrelated to the Flash resource leak. Leaving for a separate PR.

Full suite: 2589 passing, ruff clean.

dpalmqvist:
Addressed all three findings in 603959c:

**Main bug — lock-held during blocking IO** — Restructured `_expire_stale`: pops expired entries under `self._lock`, closes them after releasing it. `executor.shutdown(wait=True)` no longer stalls concurrent `ensure_loaded()` callers. The `gc/clear_cache` flush also happens outside the lock now; the `_pending_cleanups` truthy check is a benign race (worst case: one extra or skipped flush, both safe). Agreed re: `_evict_lru_if_needed` being harder to decouple since the lock is caller-held — leaving as-is per your "noting it but not blocking" call.

New test `test_expire_stale_releases_lock_before_closing` installs a sentinel `_close_loaded_model` that records `manager._lock.locked()` at call time and asserts it was released.

**Double-logging** — Removed `logger.exception` from both call sites; `_close_loaded_model` already logs per-resource ("Error closing prefetcher for X" etc) which is strictly more informative than a single generic caller log. Updated the affected `caplog` assertions in `test_expire_stale_isolates_per_model_failures` and `test_evict_absorbs_close_failure` to match the per-resource format.

**`unload()` semantics** — Added a docstring note explaining that if `_close_loaded_model` raises, the model is gone from `_loaded` but the caller sees the exception, and that this is strictly better than the pre-refactor behaviour (which would abort on the first failure and silently leak the remaining resources).

Full suite: 2590 passing, ruff clean.

dpalmqvist:
Addressed Finding 2 in e0bad7d. On the other two:

**Finding 2 — asymmetric exception type (fixed)** — Good catch. `_close_loaded_model` now always raises `ExceptionGroup` on any error, even for a single failure. Stable contract; callers use `except*` (PEP 654) for member-type handling. Updated `test_close_loaded_model_continues_on_failure` to expect a single-element `ExceptionGroup` carrying the `RuntimeError`.

**Finding 1 — event loop blocking in `_evict_lru_if_needed`** — Flagging that round 4 said *"`_evict_lru_if_needed` has the same shape but is harder to decouple since the lock is held by the caller, and eviction is rare enough that it hasn't been an issue in practice — noting it but not blocking on it"*. Round 5 escalates the same observation to a merge-blocking concern. Practically: `executor.shutdown(wait=True)` on idle pools is millisecond-scale; a real fix needs Metal-allocator coordination (the subsequent load can't allocate until the prior close drains), and that's a meaningful architectural change beyond the resource-leak scope of this PR. Happy to file a follow-up issue for proper async eviction if you've seen this surface in traces — would prefer not to expand this PR's blast radius further.

**Minor — `lm.weight_store` / `lm.model.prefetcher` not nulled** — Per your own note "not a correctness issue since the LoadedModel is dropped immediately". Nulling `lm.model.prefetcher` would mutate the underlying `FlashModelWrapper` (the model itself, not just the LM bookkeeping), which is semantically odd. Leaving the asymmetry as-is.

dpalmqvist:
Fixed in 749a939.

Good catch — the `expired_lms` list (and through it, the LM Metal buffers) was alive across `gc.collect()`, defeating the flush. Worse: Python's for-loop variable persists after the loop, so a plain `del expired_lms` isn't sufficient — the last iteration's `lm` would still be bound. Switched to pop-and-`del` per iteration:

```python
while expired_lms:
    lm = expired_lms.pop()
    try:
        self._close_loaded_model(lm)
    except Exception:
        pass
    del lm
```

The fix is verified by `test_expire_stale_drops_refs_before_gc`, which holds a `weakref.ref(lm)` and asserts it resolves to `None` at the moment `gc.collect()` is called. The first version of the fix (only `del expired_lms` after the for-loop) failed this test — caught the leaked loop variable.

dpalmqvist:
Addressed all three in 4889651:

**Bug — `unload()` 409 path** — Added `ActiveRequestsError(RuntimeError)` in `engine/model_manager.py`. `unload()` raises this specific type for the "active requests" case; `routers/manage.py` catches only `ActiveRequestsError` for 409. Close failures (ExceptionGroup) now fall through to FastAPI's 500 handler. `ActiveRequestsError` subclasses `RuntimeError` so legacy `except RuntimeError:` callers keep working.

New test `test_unload_close_failure_is_not_409` in `tests/test_routers_manage.py` monkeypatches `_close_loaded_model` to raise a `RuntimeError` and asserts the response is NOT 409 — would have caught the pre-PR misclassification. `test_unload_active_refs_raises` updated to assert both the dedicated type and the `RuntimeError` subclass relation.

**Minor — asymmetric nulling** — Nulled `lm.weight_store` alongside `lm.speculative_decoder` after close. Skipped `lm.model.prefetcher` because that field lives on the `FlashModelWrapper` itself, not on the LM bookkeeping; the wrapper is going away with the LM and mutating it from a "close the LM" helper seemed semantically wrong. The intent and exclusion are documented in the helper's docstring.

**Nit — ordering test wiring** — Added one-line comments to both cross-resource ordering tests explaining the shared `parent` MagicMock setup, so a future mock-setup change can't silently break the order assertion.

Full suite: 2592 passing, ruff clean.

dpalmqvist:
Caught the genuine review hidden in the placeholder/test noise from the misfiring workflow. Two findings, both addressed in e09634a:

**`ExceptionGroup` import** — Not needed: `pyproject.toml` pins `requires-python = ">=3.11"`, and `ExceptionGroup` has been a builtin since 3.11 (PEP 654). On 3.10 the install would already fail; no runtime path can reach the `ExceptionGroup` raise on an unsupported interpreter.

**`test_expire_stale_drops_refs_before_gc` CPython assumption** — Added a docstring paragraph documenting that the test relies on CPython's refcount-based immediate deallocation. On PyPy/Jython a MagicMock back-reference cycle could keep the LM alive past the strong-reference release, and since the test also monkeypatches `gc.collect`, the cycle collector would not clean it up. The codebase is CPython-only (uv-managed `cpython-3.11`) so this is fine in practice; the note prevents a future runtime swap from silently breaking the assertion.

Side note: I retriggered the `claude-review` workflow on commit `4889651` after the rate-limit reset — it ran to completion this time and produced this actual review (plus a series of placeholder/test comments earlier from the misfiring run). All static checks remain green.

dpalmqvist:
Round 9 — addressed the minor test fix in d26a891, filed follow-up issues for the two deferred items, and rebased the branch onto current `main`.

**Test fix** — `test_unload_close_failure_is_not_409` now uses `side_effect=ExceptionGroup(...)` instead of `RuntimeError(...)`. Matches the real type `_close_loaded_model` raises post-PR; self-documenting and will fail correctly if a future refactor reintroduces the type mismatch.

**Deferred — eviction blocks event loop** — Filed #315 (https://github.com/motsognirr/olmlx/issues/315). Captures the round-4/round-9 observation that `_evict_lru_if_needed` runs `executor.shutdown(wait=True)` (16 prefetch + 32 weight-store threads) while holding `self._lock`, stalling concurrent `ensure_loaded()` callers. Real fix needs Metal-allocator coordination on the new load path.

**Deferred — DFlashDecoder idempotency** — Filed #316 (https://github.com/motsognirr/olmlx/issues/316). Captures the round-3/round-9 observation that `DFlashDecoder.reset()` lacks the `_patched` / `_bound` flags + per-step try/finally that `EagleDecoder.reset()` has. With #301 adding eviction/expiry as new callsites for that reset, the lack of idempotency could double-unpatch the GDN class-level monkey-patch on a `close()` followed by `__del__`.

**Rebased** — Branch now sits on top of `55c6f32` (current `main`). Conflict-free (the 5 main-side commits were on the anthropic router and flash prefetcher submit_bulk path, no overlap with this PR's files). All affected test suites still green.

dpalmqvist:
All three findings addressed in 42f47d9:

**Event loop blocking in `_expire_stale`** — Wrapped the `_close_loaded_model` call with `await asyncio.to_thread(...)`. The `executor.shutdown(wait=True)` joins now run off the event loop thread; concurrent coroutines (including new `ensure_loaded` requests) keep running during the drain. New test `test_expire_stale_offloads_close_to_thread` asserts the call was routed through `to_thread`.

`_evict_lru_if_needed` can't trivially do the same — it's synchronous and called under a caller-held lock — tracked in #315.

**`unload()` semantics on close failure** — `unload()` now catches `ExceptionGroup` from `_close_loaded_model`, logs nothing extra (the per-resource log entries inside the helper already capture what failed), and returns `True`. The model is gone from `_loaded`; the user-visible "unload succeeded" semantics are satisfied even when threads leaked. Test `test_unload_absorbs_close_failure` covers the manager-level behaviour; the router test was renamed to `test_unload_close_failure_returns_200` to reflect the new contract and now asserts the 200 + `"status": "unloaded"` response.

**Null on success only** — `lm.weight_store = None` and `lm.speculative_decoder = None` moved inside their respective `try` blocks. A failed close leaves the reference alive for inspection (or future retry). Updated `test_close_loaded_model_surfaces_multiple_failures` to assert `lm.weight_store is weight_store` after its close raised, while `lm.speculative_decoder is None` (its close succeeded).

Full suite: 2601 passing, ruff clean.

dpalmqvist:
Two actionable findings addressed in ee72d7e; the third is the previously-deferred eviction blocking.

**Misleading comment in `manage.py`** — Good catch. The "falls through to FastAPI's 500 handler" comment was true two commits ago, but round 10's `unload()` change made the close failure absorb inside the manager. Updated to reflect actual behaviour: the failure is absorbed in `unload()` with per-resource logs, never reaches the router.

**Inconsistent catch breadth** — Narrowed both `_evict_lru_if_needed` and `_expire_stale` from `except Exception` to `except ExceptionGroup`, matching the `unload()` catch and `_close_loaded_model`'s documented contract. Any unexpected non-ExceptionGroup failure now propagates as a bug rather than being silently swallowed.

**Eviction still blocks event loop** — Same finding as rounds 4/5/9/10, tracked separately in #315. You correctly note "the fix is harder here because `_evict_lru_if_needed` must hold the lock" — that's exactly why it's not in this PR's scope. A proper fix needs either a sync→async refactor with lock release/reacquire semantics, or a split into pop-under-lock + close-off-thread with a separate Metal-allocator coordination story. Both are substantial changes orthogonal to the resource-leak fix this PR delivers.

All 163 affected tests pass, ruff clean. Reusing the existing tests is intentional — the eviction-absorb test exercises the real `_close_loaded_model` which wraps the inner `RuntimeError` in an `ExceptionGroup`, so the narrowed catch matches.

dpalmqvist:
Round 12 re-raises the eviction-blocking finding (rounds 4/5/9/10/11/12 — sixth time this concern has surfaced). Acknowledged but declining to expand this PR further, on operational grounds.

**Why expiry got the to_thread treatment and eviction didn't**

The two paths are structurally similar but operationally very different:

| | `_expire_stale` | `_evict_lru_if_needed` |
|---|---|---|
| When it fires | every 30s background tick | only on a new model load |
| Caller's state | mid-inference, no expectation of pause | already awaiting a slow operation |
| Other coroutines | actively serving requests | mostly waiting on the same model load |
| Pre-PR cost | dict pop (free) | `speculative_decoder.close()` (state cleanup, no threads) |
| Post-PR cost (this PR) | offloaded via `to_thread` | 16+32 thread joins on event loop |

The expiry fix was about not disrupting active inference. The eviction blocking happens inside an operation the caller is already blocked on. A concurrent `ensure_loaded` for a different model would also be waiting on `self._lock` regardless of whether close ran on-thread or off-thread, because the rest of the load logic also runs under the lock.

**Why the fix is non-trivial**

The reviewer's clean shape — pop under lock, close off-thread, then rest of load — requires splitting `ensure_loaded`'s single `async with self._lock:` block (90+ lines covering pre-checks, config resolution, eviction, load dispatch, post-load bookkeeping) into multiple lock acquisitions with careful "did the world change?" rechecks in between. The expiry path was easier because it doesn't compete with anything else for the lock; eviction does.

**Status**

- Issue #315 already tracks the proper fix for both this and any future blocking found in the eviction path.
- This PR has now been through 12 rounds of bot review since opening. The substantive findings have all been addressed; we're at the point where each round is producing increasingly surgical observations on a code path that the PR genuinely improved (it fixes a hard resource leak).
- All CI on ee72d7e is green; the PR is functionally complete and ready for merge from my side.

If you (the human) want the eviction async refactor in this PR rather than #315, say so explicitly and I'll do it. Otherwise I'd suggest landing this and following up.

@dpalmqvist — your call.

PR #337, opened 2026-05-23
feat(bench): wire quality grading + thinking toggle into bench runner

dpalmqvist:
## What this does

Connects the bench's existing grader modules (`quality.py`, `task_prompts.py`, `goldens.py`) to the runner — and adds a thinking on/off toggle for A/B comparison.

Before: `olmlx bench run` always ran 7 throughput probes with no graders, reporting tok/s only.

After: `olmlx bench run --prompt-set quality --enable-code-exec` runs the 50-prompt graded mini-set (GSM8K-20, MMLU-20, HumanEval-10) and reports `quality x/y (z%)` alongside tok/s.

## Surface

- `olmlx bench run --prompt-set {throughput,quality,all}` — selects which prompts to send. Default `throughput` preserves existing behavior.
- `olmlx bench run --enable-code-exec` — opts in to the HumanEval `code_exec` grader (off by default; runs model code in a resource-limited subprocess, safe only because olmlx is a single-user local tool).
- `OLMLX_BENCH_THINK={true,false}` — env toggle on the bench worker that injects the Ollama `think` field into `/api/chat`, so the same model can be benched with and without thinking from a driver script. Unset → engine default.
- `OLMLX_BENCH_WORKER_TIMEOUT` — extends the 600s per-scenario subprocess kill timeout (graded runs with thinking on can run several minutes).

## Internals

- Grading runs in the parent after the worker returns — the worker still only produces `output_text`, matching the contract documented in `quality.py`.
- `PromptResult` gained `grader` and `quality` fields. Backward-compatible: old run JSONs load with both as `None` (verified on existing runs on disk).
- `ScenarioResult.quality_summary()` returns `(passed, graded_total)` and drives the new `quality x/y (z%)` suffix on the per-scenario summary line.
- `apply_graders()` injects the `_enabled` flag into a **copy** of `expected` before calling `code_exec` so the prompt's `expected` dict is never mutated.
- The runner's prompt-set helper composes `task_prompts.PROMPT_SETS` (gsm8k + mmlu + humaneval); the helper raises on unknown set names.

## Tests

- `tests/test_bench_quality_wiring.py` (14 new) — prompt-set selection, grader application across grader types, code_exec gating + non-mutation, `PromptResult` round-trip with quality, `ScenarioResult.quality_summary` aggregation.
- `tests/test_bench_worker.py` (13 new) — `OLMLX_BENCH_THINK` resolution across truthy/falsy/unset/unrecognized.
- Full bench suite still green: `232 passed`. Ruff clean.

## End-to-end verification

Ran the full think-vs-no-think A/B across six thinking-capable Qwen models on the quality set (`--max-tokens 1024 --enable-code-exec`). Spot-summary:

| Model | Think | No-Think |
|---|---:|---:|
| Qwen3-0.6B-4bit | 23/50 (46%) | 33/50 (66%) |
| Qwen3.5-0.8B-4bit | 29/50 (58%) | 38/50 (76%) |
| Qwen3-4B-4bit | 45/50 (90%) | 50/50 (100%) |
| Qwen3-8B-4bit | 44/50 (88%) | 50/50 (100%) |
| Qwen3.6-35B-A3B-4bit | 42/50 (84%) | 50/50 (100%) |
| Qwen3.6-35B-A3B-6bit | 48/50 (96%) | 50/50 (100%) |

Driver scripts and result logs kept local (`bench-results/quality-think-compare/`, untracked).

### DIFF
--- a/olmlx/bench/results.py
+++ b/olmlx/bench/results.py
@@ -139,6 +139,15 @@ class RunResult:
     git_sha: str | None
     scenarios: list[ScenarioResult]
     max_tokens_override: int | None = None
+    # Which prompt set was run (throughput / quality / all). None means the
+    # run predates this field — load it as a legacy throughput run rather
+    # than fail to parse.
+    prompt_set: str | None = None
+    # Bench-level env knobs captured at run time so an A/B (e.g. think on
+    # vs off) is self-describing from the saved JSON alone. Scenario-level
+    # ``env_overrides`` cover the per-scenario knobs that the runner sets;
+    # this captures the operator-set, runner-wide ones.
+    bench_env: dict[str, str] = field(default_factory=dict)
 
     def to_dict(self) -> dict:
         return {
@@ -146,6 +155,8 @@ class RunResult:
             "timestamp": self.timestamp,
             "git_sha": self.git_sha,
             "max_tokens_override": self.max_tokens_override,
+            "prompt_set": self.prompt_set,
+            "bench_env": self.bench_env,
             "scenarios": [s.to_dict() for s in self.scenarios],
         }
 
@@ -156,6 +167,8 @@ class RunResult:
             timestamp=d["timestamp"],
             git_sha=d.get("git_sha"),
             max_tokens_override=d.get("max_tokens_override"),
+            prompt_set=d.get("prompt_set"),
+            bench_env=dict(d.get("bench_env") or {}),
             scenarios=[ScenarioResult.from_dict(s) for s in d.get("scenarios", [])],
         )
 
@@ -300,6 +313,8 @@ def create_run_result(
     model: str,
     scenarios: list[ScenarioResult],
     max_tokens_override: int | None = None,
+    prompt_set: str | None = None,
+    bench_env: dict[str, str] | None = None,
 ) -> RunResult:
     return RunResult(
         model=model,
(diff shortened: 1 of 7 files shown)


dpalmqvist:
Pushed [5444755](https://github.com/motsognirr/olmlx/commit/5444755) addressing the latest review pass:

- **Subprocess in unit test** (fixed) — `test_code_exec_enabled_injects_flag` no longer spawns a subprocess. It now monkeypatches `olmlx.bench.runner.grade` to capture what `apply_graders` hands the grader and asserts both the `_enabled=True` injection and the caller-dict non-mutation. The real `code_exec` runtime stays covered in `test_bench_quality.py`.

- **Log resolved worker timeout** (fixed) — `run_bench` now emits `Bench worker timeout: <N>s` at INFO at the start of a run.

- **Unhandled grader exceptions abort the run** (declined as a no-op) — `olmlx.bench.quality.grade()` already wraps the grader call in `except Exception` and returns `QualityResult(passed=None, detail="grader raised: …")` ([quality.py:449](https://github.com/motsognirr/olmlx/blob/main/olmlx/bench/quality.py#L449)); the existing regression test at `test_bench_quality.py:349` pins this. `apply_graders` inherits that contract, so a second try/except in the caller would be redundant.

242 bench tests pass; ruff clean.

PR #345, opened 2026-05-23
fix(prompt-cache): disable persistence for non-trimmable cache layouts (#343)

### LINKED ISSUE #343: RotatingKVCache models get near-zero prompt-cache reuse in real chat flow
## Summary

For models whose layers include `RotatingKVCache` (Gemma-4 family, gpt-oss, etc.), `_cache_supports_persistence` returns True but `_cache_supports_trim` returns False. The model-manager logs `prompt cache will only be reused for strict-extension turns` at load time, but in practice the strict-extension condition is almost never satisfied in real chat flow — every multi-turn request ends up at the `Skipping trim on non-trimmable hybrid cache; discarding and creating fresh` branch.

Net effect: the cache is stored on every request and then discarded on the next, paying eviction overhead with no prefill savings.

## Repro

```bash
OLMLX_LOG_LEVEL=DEBUG OLMLX_PROMPT_CACHE=true uv run olmlx
```

Three streaming chats to `mlx-community/gemma-4-e2b-it-OptiQ-4bit:latest`, same system prompt, varying user turn:

```
R1 prompt_eval=378 ms=135   # cold
R2 prompt_eval=404 ms=107   # full re-eval
R3 prompt_eval=428 ms=125   # full re-eval
```

Server log on R2/R3:
```
Common prefix length: 371 / 378 prompt tokens
Skipping trim on non-trimmable hybrid cache (would-be trim=15); discarding and creating fresh
Prompt cache non-trimmable-hybrid: creating fresh cache for 378 tokens
Cache stored (non-trimmable): 384 tokens (378 prompt + 6 generated)
```

Strict-extension probe (R2 sends R1's full messages + appended assistant + new user) still hits the same branch: cached holds `378 + 6 generated` tokens, new prompt's token 378+ is the *client-supplied* assistant message which doesn't match the model's actually-generated 6 tokens, so `trim_amount=6 > 0` and the cache is discarded.

## Root cause

In `_setup_prompt_cache` (`olmlx/engine/inference.py:1976`):

```python
trim_amount = len(cached.tokens) - suffix_start
...
if trim_amount > 0 and not lm.supports_cache_trim:
    # Skipping trim on non-trimmable hybrid cache; discarding
```

The cached entry is stored with both the prompt tokens *and* the generated tokens (`Cache stored: N tokens (P prompt + G generated)`). On the next request, the client-supplied assistant message rarely tokenizes to the exact same `G` tokens the model produced — especially with early-stopping on `num_predict`, EOS handling differences, or chat-template stop sequences. So `trim_amount > 0` almost always, and `RotatingKVCache`'s ban on trim means the cache is thrown away.

This isn't a bug in the trim allowlist itself (`RotatingKVCache.is_trimmable()` does return False after wrap, and the conservative class-name allowlist is the right call). The bug is upstream: **storing post-generation cache state for non-trimmable models has no realistic recovery path.**

## Suggested fix

A few options, roughly in order of preference:

1. **Don't store the generated-tokens tail for non-trimmable models.** Truncate the stored cache state to `prompt_tokens` only (drop the `+G generated` portion) before storing. Next request's strict-extension prefix match against the stored prompt tokens then has a real chance of hitting (`trim_amount=0` whenever the new prompt is a strict superset of the prior prompt). Caveat: would need the cache state itself to be trimmable down to prompt length once, which on RotatingKVCache isn't possible after wrap either — so a stronger variant:
2. **Only store post-prefill, pre-generation snapshots.** Stash the cache state immediately after the prefill phase, before the first decode step. Within-request generation still uses the live cache; the stored copy reflects only the prompt. Implementation cost: needs an extra cache-state snapshot.
3. **Just disable persistence for non-trimmable models.** Treat RotatingKVCache the same as `ArraysCache` — store nothing, log once at load time. Loses the (very rare) genuine strict-extension hit, but stops the storage overhead and clarifies the user contract. Update `CLAUDE.md`'s claim about strict-extension reuse to match.
4. **Document the actual conditions for a hit.** If the design intent really is "only helps if the client echoes back the model's exact generated tokens," say so loudly. Currently `CLAUDE.md` says "such models still benefit from strict-extension cache reuse but cannot be trimmed back on prompt divergence" — which over-promises.

(3) is the cheapest and probably the right default given how narrow (1)/(2)'s benefit is in practice. (1) or (2) only make sense if there's evidence of a workload (long shared system prompt + many independent queries) where the prefill-only snapshot would actually be hit.

## Affected models

From the registered `~/.olmlx/models.json`:
- `mlx-community/gemma-4-e2b-it-OptiQ-4bit`
- `mlx-community/gemma-4-26B-A4B-it-OptiQ-4bit`
- `mlx-community/gemma-4-31B-it-OptiQ-4bit`
- `mlx-community/gpt-oss-120b-MXFP4-Q4` (sliding-window attention)
- (any other model with `RotatingKVCache` or `ChunkedKVCache` layers)

Pure-attention models (Qwen3-4B, etc.) and hybrid-SSM models (Qwen3.5/3.6, MiniMax) are unaffected — the former trim correctly, the latter are already classified non-persistable.

## Notes

- Discovered while manually exploring KV-cache behavior across model families.
- The model-manager classification (`_cache_supports_persistence`, `_cache_supports_trim`) itself is correct and well-tested. The mismatch is between what those flags allow and what real chat flow actually triggers.

dpalmqvist:
Closes #343.

## Summary

Models with `RotatingKVCache` or `ChunkedKVCache` layers (Gemma 4 family, gpt-oss) currently store prompt-cache state that is **never reusable** in real chat flow, paying the eviction overhead with no prefill savings.

The probe correctly returns `supports_cache_trim = False` for these layouts but `supports_cache_persistence = True`. In multi-turn chat, the cached entry holds `prompt_tokens + generated_tokens`; the next request's client-supplied assistant message retokenizes differently from the model's actual generation (whitespace, EOS handling, early-stop, chat-template stop sequences), so `trim_amount > 0` on every lookup, the non-trimmable cache can't roll back, and the entry is discarded. The "strict-extension reuse" path that this combination was meant to enable never fires in practice.

## Fix

`_probe_cache_capabilities` now forces `supports_cache_persistence = False` whenever `supports_cache_trim = False`. Same treatment as `ArraysCache` (#284); exercises the existing lookup and storage short-circuits in `_setup_prompt_cache` and `_store_prompt_cache_after_generation`. Within-request cache reuse during a single generation is unaffected.

The secondary "non-persistable cache" info log is gated on `supports_cache_trim` so non-trimmable models don't double-log — the non-trimmable branch fully describes the consequence and cites #343.

## Repro / verification

Pre-fix log on R2/R3 of a Gemma-4 chat:
```
Common prefix length: 371 / 378 prompt tokens
Skipping trim on non-trimmable hybrid cache (would-be trim=15); discarding and creating fresh
Prompt cache non-trimmable-hybrid: creating fresh cache for 378 tokens
Cache stored (non-trimmable): 384 tokens (378 prompt + 6 generated)
```

Post-fix on the same model: lookup short-circuits at the `not lm.supports_cache_persistence` gate, no store-then-discard cycle, no disk I/O on the event loop, only the load-time info log fires.

## Tests

- `tests/test_model_manager.py::TestProbeCacheCapabilities::test_non_trimmable_layout_disables_persistence` — RotatingKVCache probe layout drives both flags to False (RED before the fix, GREEN after).
- `tests/test_model_manager.py::TestProbeCacheCapabilities::test_trimmable_layout_keeps_persistence` — guards the inverse: plain `KVCache` keeps persistence True.
- `tests/test_prompt_cache.py::TestNonTrimmableModelSkipsTrim` — both tests updated to set the post-#343 flag combination; the former `test_strict_extension_reuses_hybrid_cache` is replaced by `test_non_persistable_strict_extension_creates_fresh`, which asserts the fresh-cache fallback fires even on an artificial strict-extension setup.

Full suite: 2961 passed, 4 skipped. Ruff clean.

## Trade-off considered

Option 2 from the issue (snapshot cache pre-generation, store prompt-only state, would actually deliver strict-extension hits in real chat) was rejected as scope: it requires forking `stream_generate` or pre-prefilling manually to insert a deep-copy boundary, doubles KV memory transiently, and touches the most fragile code path in the engine. Option 3 (this PR) is the cheapest correct fix and matches the existing `ArraysCache` (#284) pattern. Option 2 can be a follow-up if a Gemma-4-on-long-shared-system-prompt workload becomes important.

## Affected models

- `mlx-community/gemma-4-{e2b,26B-A4B,31B}-it-OptiQ-4bit`
- `mlx-community/gpt-oss-120b-MXFP4-Q4`
- Any layout with `RotatingKVCache` or `ChunkedKVCache` layers.

Pure-attention models (Qwen3, Llama-3, etc.) and hybrid-SSM models (Qwen3.5/3.6, MiniMax) are unaffected — the former still get full strict-extension + trim reuse, the latter were already non-persistable via #284.


dpalmqvist:
Good catch on the logging regression — `_TRIMMABLE_CACHE_CLASSES ⊂ _PERSISTABLE_CACHE_CLASSES`, so `not persist AND trim` was indeed dead, and the only log that fired for `ArraysCache` (Qwen3.5) was the non-trimmable one citing #343 / RotatingKVCache. Wrong family, wrong issue number.

Fixed in 7c76bf9. Approach:

- Capture the raw layout-persistence result (`persist_layout_ok`) before applying the `#343` trim-implies-persist fold.
- Replace the two log blocks with one dispatch that routes on `persist_layout_ok`:
  - `persist_layout_ok=True` + `trim_ok=False` → `#343` (RotatingKVCache / ChunkedKVCache).
  - `persist_layout_ok=False` → `#284` (ArraysCache / hybrid SSM / unclassified).
- The two sub-cases are mutually exclusive given the current allowlists, so it's exactly one line per affected model with correct attribution.

Added tests for both attribution paths: `test_non_trimmable_persistable_layout_logs_343` and `test_non_persistable_layout_logs_284`. Full suite: 2963 passed.

dpalmqvist:
Round-2 review applied in `e3f934b`:

1. **`_cache_supports_trim` docstring (`model_manager.py:283`)** — was claiming "strict-extension cache reuse" post-#343; rewritten to describe the actual probe-fold consequence (zero cross-request reuse for these models).

2. **Equivalence comment on `persist_ok = persist_layout_ok and trim_ok`** — added an inline note explaining that under the current allowlists this is equivalent to `persist_ok = trim_ok`, but the two-term form is kept intentionally for forward compat (a future trimmable-but-non-persistable class).

3. **Stale `test_kvcache_plus_rotating_supports_persistence` docstring** — updated to reflect that this test only exercises the raw layout-level allowlist signal; the probe-fold behaviour is tested in `TestProbeCacheCapabilities::test_non_trimmable_layout_disables_persistence`.

4. **New `test_mixed_layout_kv_plus_rotating_disables_persistence`** — pre-existing probe tests used homogeneous lists; this adds the actual Gemma-4 layout (`[KVCache, RotatingKVCache, KVCache, RotatingKVCache]`) so a future allowlist change that broke the `all(...)` quantifier on heterogeneous layouts would be caught.

Full suite: 2968 passed, 4 skipped. Ruff clean.

dpalmqvist:
Round-3 review addressed in `066dd67` (+ format fixup `f44ff8d`):

**ChunkedKVCache test gap** — added `test_chunked_kv_cache_layout_disables_persistence` in `TestProbeCacheCapabilities`. Pre-existing tests only exercised `RotatingKVCache`; if a future contributor accidentally promoted `ChunkedKVCache` into `_TRIMMABLE_CACHE_CLASSES` while adding a new model family, the wasted store-then-discard cycle this PR removes would silently come back. The companion test guards the symmetric case.

The other two observations (logging-attribution edge case for a hypothetical trimmable-but-non-persistable class, and the formula verification) were confirmed-no-change; the existing inline comments already document the reasoning.

Also rebased onto `origin/main` to pick up #347 (no conflicts — that PR touched `_resolve_keep_alive` / `ensure_loaded`, not our `_probe_cache_capabilities`).

Full suite: 2976 passed, 4 skipped. Ruff clean.

dpalmqvist:
Round-4 review applied in `a0712ff` (doc/rename only, no behaviour change):

1. **`_PERSISTABLE_CACHE_CLASSES` comment block** — added inline cross-references to #343 on the `RotatingKVCache` and `ChunkedKVCache` entries, so a contributor adding trim support for one of these classes sees the fold cited at the allowlist (the first place they'd look) rather than only in the probe code far away.

2. **Test taxonomy** — `TestNonTrimmableModelSkipsTrim` → `TestNonPersistableLookupShortCircuit`, `test_non_trimmable_skips_trim_call` → `test_diverging_prompt_creates_fresh`. The old names suggested trim was the gate being skipped, but post-#343 the lookup short-circuits at the persistence gate *before* reaching the trim check — trim is never consulted for these models. Added a class docstring describing the contract and the rename rationale.

Both other observations from the review (formula equivalence, `persist_layout_ok=False` init safety) were confirmed-no-change.

Full suite: 2976 passed, 4 skipped. Ruff clean.

dpalmqvist:
Round-5 review applied in `e48d2eb` (doc-only, no behaviour change):

1. **Two-term formula comment** — dropped the "functionally equivalent to `persist_ok = trim_ok`" phrasing and replaced with an explicit "do not simplify; the two-term form is kept so the formula self-corrects if the subset relation ever breaks." Removes the footgun.

2. **"Stage both results before assigning" block comment** — extended to mention the new third intermediate `persist_layout_ok`, computed inside the try and consumed in the logging dispatch outside it. Notes that its read site is gated on `probe_succeeded`, so a hypothetical fallible insertion between the two probes degrades safely.

Also rebased onto `origin/main` to pick up #348 (manifest derive-on-demand) — no conflicts (#348 touched `_load_model` / `_resolve_status`, not `_probe_cache_capabilities`).

Full suite: 2987 passed, 4 skipped. Ruff clean.

dpalmqvist:
Round-6 review: ran a fresh local code-review pass (the bot's CI run hit an Anthropic session limit). One real follow-up surfaced — applied in `0651071`:

**Log/comment misattribution at the lookup and storage sites in `inference.py`.** The round-2 fix updated the load-time INFO log in `_probe_cache_capabilities` to distinguish #284 (ArraysCache) vs #343 (RotatingKVCache/ChunkedKVCache), but two symmetric DEBUG log + comment blocks in `inference.py` still cited only "Issue #284: hybrid SSM-style ArraysCache":

- `_setup_prompt_cache` non-persistable lookup short-circuit (~line 1914).
- `_store_prompt_cache_after_generation` non-persistable storage early-return (~line 2227).

Post-#343 both also fire for Gemma 4 / gpt-oss via the `persist_ok = persist_layout_ok and trim_ok` fold. An operator grep'ing for "hybrid SSM/ArraysCache" while debugging a Gemma 4 cache issue would land on those messages and chase the wrong upstream issue.

Widened both comment blocks to enumerate both families with their respective issue links, and widened the DEBUG log string to mention "non-trimmable sliding-window" alongside "hybrid SSM/ArraysCache". No behaviour change.

Affected file tests pass (470 passed, 3 skipped). Ruff clean.

dpalmqvist:
Round-7 review applied in `970b6f2`:

**Stale-entry cleanup assertion gap** in `test_diverging_prompt_creates_fresh` and `test_non_persistable_strict_extension_creates_fresh`. Both pre-populate `prompt_cache_store` with a stale entry to simulate a prior turn, then assert the non-persistable short-circuit fires — but neither checks that the `peek()`/`remove()` cleanup branch in `_setup_prompt_cache` actually ran. A regression that dropped the `remove()` call would still pass the existing `mock_trim.assert_not_called()` / `mock_make_cache.assert_called_once_with(lm.model)` / full-prompt checks while silently leaking the stale entry — defeating one of the cleanups this PR explicitly cites as a benefit.

Added `assert lm.prompt_cache_store.peek("") is None` after the existing assertions in both tests. No production code change.

The other two round-7 observations (latent-risk note on `persist_layout_ok = False` default if anyone later sets `probe_succeeded = True` inside the except block, and the formula-correctness reconfirmation) were confirmed-no-change; current code preserves the invariants.

dpalmqvist:
Round-8 review applied in `a94834f`:

**`_TRIMMABLE ⊂ _PERSISTABLE` subset invariant lacked an automated test.** The two-term fold (`persist_ok = persist_layout_ok and trim_ok`) and the load-time log dispatch both silently rely on this. The bot rightly observed: if a future mlx-lm release (or local allowlist edit) added a class to `_TRIMMABLE_CACHE_CLASSES` without also adding it to `_PERSISTABLE_CACHE_CLASSES`, the formula would still produce a safe `persist_ok = False`, but the log would misattribute the model as #284 / ArraysCache instead of flagging a genuinely novel trimmable-but-non-persistable case.

Added `test_trimmable_is_subset_of_persistable` in `TestPersistableCacheClassesAllowlist`. One-liner set-difference check with a directed error message pointing at both remediation paths (add to the persist allowlist, or audit the log dispatch). No production code change.

dpalmqvist:
Round-9 review applied in `e3092de` (doc/test only, no behaviour change):

1. **`_cache_supports_trim` docstring** — round-2's rewrite picked up a downstream-policy claim ("post-#343 such models also have cross-request persistence force-disabled") that this function itself doesn't enforce; the fold happens in `_probe_cache_capabilities`. Rewritten to describe what this function actually does (pure allowlist check) and cross-reference `_probe_cache_capabilities` for the consequence.

2. **`lm.supports_cache_trim = False` is dead noise** in `test_diverging_prompt_creates_fresh` and `test_non_persistable_strict_extension_creates_fresh`. Post-#343 the lookup short-circuits at the persistence gate before trim is consulted. Removed both assignments, added a one-line comment so a future maintainer doesn't re-add a trim flag thinking it gates the branch.

3. **`persist_layout_ok = False` init comment** in `_probe_cache_capabilities` — the six-line block restated what the outer dispatch (`if probe_succeeded and not lm.supports_cache_persistence:`) already makes self-evident. Tightened to three lines. Left the (separate) staging-pattern block inside the try untouched.

The fourth review item (hypothetical log misattribution if the `_TRIMMABLE ⊂ _PERSISTABLE` invariant breaks) is confirmed-no-change — that's the exact case `test_trimmable_is_subset_of_persistable` (round 8) catches at CI time.

PR #384, opened 2026-05-26
feat(grammar): JSON-mode and JSON-Schema constraints via xgrammar (#361)

### LINKED ISSUE #361: Structured outputs (JSON Schema / grammar) via xgrammar logits processor
## Why

Every caller currently reinvents JSON validation around the regex tool parser. OpenAI `response_format`, Anthropic tools, and Ollama `format` deserve real grammar-level guarantees — not best-effort parsing.

## Scope

- Integrate xgrammar (fastest on Apple Silicon) as a logits processor
- OpenAI: `response_format={"type":"json_schema",...}` and `{"type":"json_object"}`
- Anthropic: build a grammar from each tool's `input_schema` and constrain tool-use generation
- Ollama: `format` field on `/api/chat` and `/api/generate`
- Works under streaming and all speculative strategies (constrain *after* verify to keep the verify token-identical to a free run)

## Out of scope

- User-supplied EBNF/regex grammars (follow-up)
- CFG-only formats (Lark, ANTLR)

## Acceptance criteria

- Tool calls from all three routers always parse against the declared schema
- Bench gains a JSON-mode quality eval
- Negligible TTFT regression on un-constrained requests (no global hook)

## Files

- `engine/inference.py`
- `routers/openai.py`, `routers/anthropic.py`, `routers/chat.py`, `routers/generate.py`
- `schemas/openai.py`, `schemas/anthropic.py`, `schemas/chat.py`, `schemas/generate.py`
- new `engine/grammar/`

dpalmqvist:
## Summary

Adds grammar-constrained decoding for OpenAI `response_format` and Ollama `format`, enforcing shape via xgrammar as an mlx-lm logits processor rather than the previous soft-prompt-only path.

- **New `olmlx/engine/grammar.py`** — `GrammarSpec`, `parse_response_format()`, compile cache (per-tokenizer + per-spec, so repeat requests pay the ~7 ms compile only once), `GrammarLogitsProcessor` that unpacks xgrammar's bitmask through numpy into an MLX bool mask and sets disallowed logits to `-inf` via `mx.where`. No torch on the hot path.
- **`engine/inference.py`** — `_install_grammar_processor()` helper + `grammar_spec` kwarg on `generate_chat` / `generate_completion`. `grammar_active` plumbed through to `_stream_completion` / `_full_completion[_inner]`.
- **OpenAI router** — `response_format={"type":"json_object"}` and `{"type":"json_schema",...}` now produce a real grammar. Named-schema hint still injected as belt-and-braces so the model knows what to fill in.
- **Ollama routers** — `format: \"json\"` and `format: <schema_dict>` wired on `/api/chat` and `/api/generate`. Request schemas updated to `str | dict | None`.
- **VLM and distributed paths skip with a warning** — mlx-vlm doesn't accept `logits_processors`; workers don't receive the processor over the sideband.
- **Speculative + grammar interop**: when grammar is requested on a speculative-loaded model the speculative path falls back to mlx-lm for that request (the speculative decoders don't consume `gen_kwargs[\"logits_processors\"]`). Constrain-after-verify across the four strategies is a follow-up.
- **Anthropic tool-use constraint is intentionally deferred** — needs per-template trigger detection (Qwen `<tool_call>`, Mistral `[TOOL_CALLS]`, Gemma 4 `<|tool_call>`) and structural-tag grammar; separate body of work.

## What this changes for callers

| Surface | Before | After |
|---|---|---|
| OpenAI `response_format={\"type\":\"json_object\"}` | Soft prompt: \"respond with valid JSON only\" | Token-level grammar enforces valid JSON, plus the soft prompt |
| OpenAI `response_format={\"type\":\"json_schema\",...}` | Soft prompt mentioning schema name; **schema not enforced** (log line said so) | Token-level grammar enforces the schema; named-schema hint kept as belt-and-braces |
| Ollama `format: \"json\"` | Not handled at all | Token-level grammar for any valid JSON value |
| Ollama `format: <schema_dict>` | Rejected (`format` was `str | None`) | Token-level grammar enforces the schema |

## What's deferred

- **Anthropic tool-use** — per-template trigger detection + structural-tag grammar, separate PR.
- **Speculative + grammar** — currently disabled-with-warning per request; constrain-after-verify across `classic` / `dflash` / `eagle` / `pld` needs decoder plumbing.

## Test plan

- [x] 31 new tests across `tests/test_grammar.py`, `tests/test_inference_grammar.py`, `TestFormatField` in chat/generate router tests, and updates to `TestResponseFormat` to assert `grammar_spec` is forwarded (not just the system-message hint).
- [x] Full suite: 3090 pass, 4 skipped.
- [x] `uv run ruff check` + `uv run pyright` clean on the touched files.
- [ ] Manual smoke against a real Qwen3 model with each of the four surfaces — JSON object / JSON schema via OpenAI; `format: \"json\"` / `format: <schema>` via Ollama.

Closes #361


dpalmqvist:
## Smoke results against `qwen3:8b`

Ran all four router surfaces end-to-end after starting `uv run olmlx serve` and waiting for ready.

**1. OpenAI `response_format={"type":"json_object"}`** — emitted a valid JSON array (xgrammar's built-in JSON grammar allows any JSON value, not just objects). Grammar shape ✓; semantic content was junky because Qwen3 default-thinks and the response shape was an array, not what the user asked for — orthogonal to the grammar.

**2. OpenAI `response_format={"type":"json_schema",...}`** (strict object with `additionalProperties: false`, required `name`/`age`/`color`) — produced `{"name": "Whiskers", "age": 3, "color": "gray and white"}`. Schema ✓.

**3. Ollama `format: "json"`** — `{"hello": "world"}`. ✓.

**4. Ollama `format: <schema>`** — first attempt hit `max_tokens` mid-number with the model stuck in a digit loop (`10.4222222...` then `10400000000000...`). This is a known characteristic of greedy decoding under JSON's permissive number grammar — every digit is grammar-valid, the model strongly prefers another digit, no termination signal. Added a `maximum: 100` bound and got `{"country": "Sweden", "capital": "Stockholm", "population_millions": 100}`. ✓.

## One real bug found and fixed in commit e1f2964

`Unsupported tokenizer type: <class 'mlx_lm.tokenizer_utils.TokenizerWrapper'>` — xgrammar's `TokenizerInfo.from_huggingface` does a strict isinstance check against `PreTrainedTokenizerBase` and rejects mlx-lm's wrapper. Fixed by peeling `._tokenizer` opportunistically in `_install_grammar_processor`. Two regression tests added.

The digit-loop issue is **not** a wiring bug — it reproduces with any grammar-constrained backend on temperature=0 + unbounded numeric schemas, and the fix is upstream of olmlx (schema authors should bound numbers, or callers should use temperature > 0). Worth a docs note but not a blocker.

dpalmqvist:
Round 2 — addressed both reviews in 929a37c.

**Fixed (all 6 bugs across the two reviews + both minors + nit):**

- ✅ **Cache cleared on model unload**: `drop_for_tokenizer` wired into `ModelManager._close_loaded_model` (runs *before* the tokenizer ref is nulled, so the id is still alive when we clear). Reviewer 1 bug 1, Reviewer 2 bug 1.
- ✅ **Unrecognised dict `type`**: split dispatch via explicit `_OPENAI_RESPONSE_FORMAT_TYPES` frozen set; bare Ollama schema accepted only with explicit JSON-Schema markers OR a `type` whose value is a real JSON-Schema type. `{"type": "image_url"}` now raises `ValueError`. Reviewer 1 bug 2.
- ✅ **`ValueError` → HTTP 422 (not 500)** in all three routers (openai.py, chat.py, generate.py). Reviewer 1 bug 3, Reviewer 2 bug 2.
- ✅ **Grammar + tools guard**: `has_tools=True` skips with warning. The JSON grammar would mask `<tool_call>`/`[TOOL_CALLS]`/`<function=...>` so the model could never emit a tool call. Reviewer 1 bug 4.
- ✅ **`_resolve_model_vocab_size` prefers `lm_head` over `embed_tokens`**: matters for untied/padded vocab heads (Phi-3, Llama-3.2-Vision). Reviewer 2 bug 3.
- ✅ **TOCTOU in `compile_for_tokenizer`**: lock now held across the entire check-compile-store sequence. Reviewer 2 bug 4.
- ✅ **`neg_inf` cached per dtype** to skip Metal alloc on the hot path. Reviewer 1 minor.
- ✅ **`if/elif` cleanup** in `_stream_completion` via a `use_speculative` variable. Reviewer 1 minor.
- ✅ **`GrammarSpec.__hash__ = None`**: hashing now fails fast at use site. Reviewer 2 nit.

**Bonus fix found while writing the cache-clear test**: my original `getattr(tok, "_tokenizer", tok)` over-peeled HF fast tokenizers down to their Rust `tokenizers.Tokenizer` core (xgrammar rejects those too). New `unwrap_mlx_tokenizer` helper gates on outer class name (`TokenizerWrapper`) instead of attribute presence.

**Not addressed (design concern, deferred):**
- xgrammar as opt-in extras. Valid concern, but xgrammar is core to the feature — making it optional means the feature randomly fails depending on install. The `try: import xgrammar` graceful-degrade I kept is for the runtime-degraded case (e.g. xgrammar fails to load on an unsupported platform), not as a user-facing toggle. Happy to revisit in a follow-up if you want it gated.

**Test coverage:** 13 new tests covering each fix. Full suite 3101 pass / 4 skipped. Ruff format + check clean. Pyright 0 errors on touched files.

CC @claude

dpalmqvist:
Round 3 — addressed in 559ad1b. Also folded in the round-2 optional `lm_head`-first tightening.

**Applied (5/6 actionable items):**

- ✅ **Bug 1, cleanup tokenizer identity**: `_close_loaded_model` now uses `lm.text_tokenizer` (matches what `_install_grammar_processor` keys on).
- ✅ **Bug 2, explicit `use_speculative` in `_full_completion_inner`**: mirrors the streaming path so disabling is auditable rather than implicit through `elif` ordering.
- ⚠️ **Perf 3, pre-allocated unpack buffer**: tried `np.unpackbits(..., out=...)` — not supported in numpy 2.4 (no `out=` parameter exists). Backed out. Kept the smaller win: dropped the redundant `.astype(bool)` round-trip — numpy returns uint8 0/1 which `mx.where` already treats as boolean, saving the per-step ~150 KB intermediate alloc for vocab=151k.
- ✅ **Perf 4, double-checked locking on compile cache**: lock released during `GrammarCompiler` construction; duplicate-compile race is harmless since compiled grammars are immutable.
- ✅ **Minor, constants above use site**: `_OPENAI_RESPONSE_FORMAT_TYPES` and `_JSON_SCHEMA_TYPES` moved to the top of `grammar.py`.

**Round-2 carry-over also applied:**
- ✅ `_resolve_model_vocab_size` traversal swapped from owner-first to attr-first, so a nested `model.model.lm_head` is preferred over a top-level `model.embed_tokens` (was the reviewer's non-blocking note from round 2). Regression test added.

**Deferred (design concerns, separate decisions):**
- xgrammar as opt-in extras — same response as round 2: making it optional means the feature randomly fails depending on install. Happy to gate in a follow-up.
- `X-Grammar-Active: false` header / warning field on silent skip — protocol surface decision, worth a separate discussion.

**Coverage**: 3102 tests pass, ruff + pyright clean.

Issue #360, opened 2026-05-25
Amortize GDN rollback overhead in dflash/eagle on high-acceptance requests


dpalmqvist:
## Why

CLAUDE.md notes the empirical `classic` win on Qwen3.5/3.6 is partly because its caches are trim-able and incur no GDN rollback per verify. When a `dflash`/`eagle` request has consistently high acceptance, rejection is rare enough that always paying full rollback cost is wasteful.

## Scope

- Track per-request rejection rate (windowed)
- Skip `_GDNStateCapture` replay opportunistically when the rolling rejection rate is below a threshold
- Add a fast-path that defers replay until a rejection actually occurs, paying it as a one-shot fix-up

## Acceptance criteria

- `dflash`/`eagle` benchmarks on Qwen3.5-27B-4bit match or exceed `classic` throughput
- No correctness regression vs the eager-rollback path on the existing speculative tests
- Rollback skip rate exposed in `PrefetchStats`-style metrics

## Files

- `engine/dflash/decoder.py`
- `engine/eagle/decoder.py`
- `engine/gdn_rollback.py`

dpalmqvist:
## Decision: won't do — closing

Closing as won't-do: this optimizes decoders that don't work well enough to keep.

### Why

This issue proposes amortizing the `_GDNStateCapture` rollback cost in `dflash`/`eagle` on high-acceptance requests. But per CLAUDE.md's own measured numbers, those decoders sit at **~2% acceptance (dflash)** and **~6% (eagle)** on Qwen3.5 — the regime this optimization targets ("consistently high acceptance, rejection is rare") essentially never occurs for them. Rejection isn't rare; it's the common case.

The acceptance criterion — *"match or exceed `classic` throughput"* — isn't reachable by shaving rollback overhead when the base acceptance is 2–6%. The throughput gap is dominated by the low acceptance itself (feature-space drift, MASK convention mismatch, compounding feature-chain error), not by rollback bookkeeping.

On top of that, #278 already leans toward keeping `dflash` experimental indefinitely / wontfix absent published reference draft models, and `classic`/`mtp` dominate on the targets where speculative actually helps. Optimizing a path that's slated never to promote is polishing a dead end.

If `dflash`/`eagle` ever get a genuine acceptance breakthrough (new training recipe, fixed MASK convention), the rollback-amortization idea is worth revisiting then — but it's premature now. Closing as won't-do.

PR #387, opened 2026-05-27
fix(eagle): use buffer-based GDNStateCapture API after extraction refactor

dpalmqvist:
## Summary
- `EagleDecoder` was using the pre-refactor `GDNStateCapture` API (`_GDNStateCapture(model)`, `capture.clear()`, `capture.rollback(cache, accepted, trim)`) — none of these exist on the post-refactor class in `gdn_rollback.py`. A hybrid linear-attention target (Qwen3.5/3.6) running EAGLE would crash on the first partial-acceptance verify.
- Fix mirrors DFlash: `_GDNStateCapture.for_model(target)` to get capture + buffer, `use_buffer(buffer)` to register, `buffer.clear()` per step, `capture.rollback_single(buffer, cache, accepted, trim)` on rejection.
- Test stubs updated to expose the real API (`for_model`/`use_buffer`/`rollback_single`) so future drift surfaces as a stub mismatch instead of going unnoticed.

The bug was masked because every existing `TestEagleDecoderGDNPath` case monkey-patches `_GDNStateCapture` with a fake mirroring the *old* API. Hybrid-target EAGLE has no real-API coverage in the unit suite — that gap remains, but the stub now at least models the right method names.

Surfaced while triaging #360 (GDN rollback amortization on high-acceptance requests) — separate issue, but #360 work would have hit this crash immediately.

## Test plan
- [x] `uv run pytest tests/test_eagle.py tests/test_gdn_rollback.py tests/test_dflash.py tests/test_speculative.py tests/test_speculative_hybrid.py` — 130 passed
- [x] `uv run ruff check` + `uv run ruff format --check`
- [ ] Run EAGLE against a real Qwen3.5/3.6 target end-to-end (no local checkpoint handy — would still hit the ~5% acceptance ceiling documented in CLAUDE.md, but rollback path would no longer crash)

### DIFF
--- a/olmlx/engine/eagle/decoder.py
+++ b/olmlx/engine/eagle/decoder.py
@@ -46,6 +46,7 @@ from olmlx.engine.dflash.decoder import (
 from olmlx.engine.gdn_rollback import (
     _HAS_GDN,
     find_gdn_class as _find_gdn_class,
+    GDNBuffer as _GDNBuffer,
     GDNStateCapture as _GDNStateCapture,
     get_model_layers as _get_layers,
 )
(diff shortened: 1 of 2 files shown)


dpalmqvist:
@claude review this PR

PR #388, opened 2026-05-28
DFlash: multi-window training (#382)

### LINKED ISSUE #382: DFlash: multi-window training (multiple masked blocks per sequence)
## Summary

Currently the DFlash trainer uses **one random window per sequence per step** (`olmlx/engine/dflash/prepare.py` module docstring). The paper ([arxiv:2602.06036](https://arxiv.org/abs/2602.06036)) trains with a "fixed number of masked blocks per sequence, anchors resampled each epoch."

Each target forward pass produces hidden states for the full sequence, but we only train on one `block_size`-length window — wasting most of the computed hidden states. Training on multiple windows per sequence would amortize the cost of the target forward pass and produce a better-trained draft in the same wall-clock time.

## What would change

- Instead of picking a single random pivot `p` per sequence, sample `K` non-overlapping windows (e.g. 4-8 windows per sequence)
- Compute loss as the sum (or mean) over all windows in the batch
- The gradient update uses the same target forward output for all windows — no extra target compute

## Considerations

- **Memory**: Each additional window adds a `(B, block_size, vocab)` logits tensor. For `block_size=16` and a 250k vocab, that's ~16 MB per window per batch — manageable for 4-8 windows.
- **Distillation**: KL terms scale linearly with window count. The distill path already captures target logits for all positions, so the per-window logit slice is cheap.
- **Per-position weighting**: The `position_decay_gamma` weighting should apply independently per window (each window treats its first position as `k=1`).
- **Pad avoidance**: Each window must independently be checked for pad contamination; not all K windows will be valid. The window count `K` should be a target, not a guarantee.
- **CLI**: `--train-windows-per-step N` (default 1 for legacy behavior).

## Related

- Original gap report: #317
- Fix PR for remaining Gap 6: #381

dpalmqvist:
## Summary
- Adds `--train-windows-per-step N` to `olmlx dflash prepare`. K=1 (default) is bit-exact with the prior single-window behaviour; K > 1 trains on N non-overlapping masked windows per batch in a single optimizer step, reusing one target forward.
- Spec: `docs/superpowers/specs/2026-05-27-dflash-multi-window-training-design.md`
- Plan: `docs/superpowers/plans/2026-05-28-dflash-multi-window-training.md`

## Design
- New `_select_pivots` helper places K pivots via slot-and-jitter inside the batch's shared unpadded prefix. Each pivot's `[p, p+block_size]` span fits inside its slot, so adjacent pivots are at least `block_size + 1` apart and their spans never overlap (proof in the helper docstring; verified across 200 seeds in tests).
- K=1 delegates to the existing `_select_pivot` and preserves the legacy `random.randint` call exactly — same RNG draw under a fixed seed, same monkey-patch surface for the inference-convention test.
- Multi-window loss closure runs `_draft_loss` per window with a fresh draft cache, sums per-window losses, and divides by `len(windows)`. One `optimizer.update` per batch.
- When the shared prefix is too short to fit K non-overlapping windows, fewer are used and the batch still produces a real gradient update (K is a target, not a guarantee). `--steps` continues to count optimizer updates, not windows.

## Closes
Closes #382.

## Test plan
- [x] `uv run pytest tests/test_dflash_prepare.py tests/test_dflash.py tests/test_dflash_distill.py tests/test_dflash_precompute.py tests/test_cli.py` — 291 PASS.
- [x] `uv run ruff check olmlx/ tests/` clean.
- [x] `uv run ruff format --check olmlx/ tests/` clean.
- [x] K=1 bit-exact regression test (`TestK1BitExactness`) pins the pre-refactor loss baseline: `[4.2952, 4.3009, 4.2603, 4.2794, 4.1174]`.
- [x] K=4 non-overlap invariant tested across 200 seeds.
- [x] K=4 short-range downgrade (pad-aware): emits the "shared prefix too short" debug log, no skip increment.
- [x] K=4 all-pad batches still trigger the consecutive-skip abort path.
- [x] K=4 × `--distill`: target logits captured once, sliced K times.
- [x] K=4 × precomputed tuples: monkey-patches `_capture_target_outputs` to fail if it is called — guarantees the precomputed path is actually hit.
- [x] `uv run olmlx dflash prepare --help` shows the new flag with the documented help text.
- [ ] (Out of scope for this PR) Empirical sweep on a real 27–35B target to pick a sweet-spot default K.


dpalmqvist:
@claude review this PR please

PR #391, opened 2026-05-28
Cross-request radix prefix cache (#365)

### LINKED ISSUE #365: Cross-request radix prefix cache (in-memory tier)
## Why

The existing prompt cache stores trim-able caches to disk and reuses them within a request. An in-memory radix tree above that disk tier catches sibling branches (same system prompt + first user turn → different next turn) without a disk round trip, and gives non-trim-able cache types (gated-delta linear-attention; sliding-window/rotating caches) somewhere to live without wasting disk I/O.

Closes the door on the wasted-I/O failure mode described in issue #343.

## Scope

- Radix tree above `PromptCacheStore`, LRU + RAM budget
- SGLang-style node sharing across sibling continuations
- Disk persistence gated on the in-memory tier (non-trim-able caches stay in RAM, never spill)
- Metric: in-memory hit rate, eviction count

## Out of scope

- Distributed cache sharing across distributed workers (follow-up)
- Compressed prompt cache (LLMLingua-style)

## Acceptance criteria

- Hit-rate metric exposed
- Claude-Code-style workload (large system prompt, branching tool calls) shows ≥50% prefill saved on the system-prompt prefix
- No quality regression on the existing prompt-cache tests

## Files

- new `engine/prompt_cache/`
- `engine/inference.py`
- `config.py`

dpalmqvist:
## Summary

Closes #365. Adds an in-memory token-prefix radix index above the existing `PromptCacheStore` so a `cache_id` miss falls back to a longest-prefix sibling lookup. On hit, the matched entry is re-keyed to the new `cache_id` (takeover — no KV copy) and the existing trim+extend path handles alignment. Catches the Claude-Code branching workload where many sessions share a large system prompt.

## What's new

- `engine/prompt_cache/` package: `state.py` (`CachedPromptState`), `radix.py` (`PrefixCacheIndex` trie), `store.py` (`PromptCacheStore` with radix + metrics + RAM budget), `metrics.py` (`CacheMetrics`).
- `_setup_prompt_cache` falls back to `find_by_prefix` + `takeover` on cache_id miss, gated by `OLMLX_PROMPT_CACHE_RADIX` and `supports_cache_persistence`.
- Trie returns the deepest reachable terminal — including subtree terminals past the divergence point — so true siblings (system prompt shared, different next turn) hit.
- Soft RAM byte budget evicts beyond `max_slots`. Disk metrics (`evictions_disk`, `bytes_on_disk`) tracked in `_cleanup_disk`.
- Eight metric counters surface on `/api/ps` per loaded model via a new `cache_metrics` field.

## Settings

| Name | Default | Purpose |
| --- | --- | --- |
| `OLMLX_PROMPT_CACHE_RADIX` | `true` | Kill switch for the sibling fallback. |
| `OLMLX_PROMPT_CACHE_RADIX_MIN_PREFIX_TOKENS` | `256` | Below this, skip takeover and fresh-prefill. |
| `OLMLX_PROMPT_CACHE_RAM_BUDGET_GB` | `8.0` | Soft byte budget; `_max_slots` remains the hard cap. |

## Out of scope (per spec)

- Concurrent two-stream sibling branching (one stream wins via takeover; the other falls back to fresh prefill).
- Rescuing non-persistable caches (#284 SSM ArraysCache, #343 RotatingKVCache / ChunkedKVCache).
- Distributed cache sharing across workers.

## Test Plan

- [x] 3302 tests pass, 0 fail (`uv run pytest`).
- [x] 28 new tests covering trie, metrics, store integration, sibling-prefix fallback, RAM budget, `/api/ps` cache_metrics surfacing.
- [x] Regression: multi-model `/api/ps` no longer shadows the outer `model_store` local (added by reviewer feedback during PR prep).
- [x] `ruff check` + `ruff format --check` clean.
- [x] CLI smoke (`uv run olmlx --help`) imports cleanly.
- [ ] Manual: run Claude Code against the server with a large system prompt across multiple sessions; verify `/api/ps` reports `radix_hits > 0` and `cache_id_hits` reflects expected reuse.

## Design + plan docs

- Spec: `docs/superpowers/specs/2026-05-28-radix-prompt-cache-design.md`
- Plan: `docs/superpowers/plans/2026-05-28-radix-prompt-cache.md`


dpalmqvist:
Thanks for the review — both findings addressed:

- **Disk restores bypass RAM budget** (medium): `async_get` now calls `_enforce_ram_budget` after the in-memory insert and guards the original disk-file unlink on `cache_id in self._entries`. The sync `_load_from_disk` path already went through `self.set()` (which enforces the budget), but its unconditional unlink had the same dual-tier-loss bug when the budget evicted the just-loaded entry — now guarded too. Test in `tests/test_kv_disk_cache.py::TestAsyncDiskCache::test_async_get_disk_restore_enforces_ram_budget`. Commit `7a923bc`.

- **Trie one-cache-id-per-terminal** (low): Trie terminals now hold a `set[str]` of cache_ids; `insert` adds, `remove` discards, prune fires when the set is empty AND no children. Keeps the store/index invariant: every entry in `_entries` stays reachable through the trie even when distinct entries share identical `state.tokens`. Test in `tests/test_prompt_cache_radix.py::TestPrefixCacheIndexRemove::test_duplicate_path_keeps_both_terminals`. Commit `db3bbe7`.

3304 tests pass, ruff clean.

dpalmqvist:
Addressed in `4b20946`. The trie's `find_longest_prefix` now takes a `min_depth` parameter and short-circuits to `(None, 0)` immediately when the descent depth is below threshold — no subtree DFS for queries that only share a BOS (or any short prefix) with the stored entries. `find_by_prefix` passes `min_prefix_tokens` through. Regression test in `test_find_by_prefix_short_circuits_below_threshold`. 3305 tests pass, ruff clean.

dpalmqvist:
Fixed in `f868116`. `_evict_all_to_disk_sync` and `async_evict_all_to_disk` now increment `evictions_ram` by the number of cleared entries before resetting `_entries`/`_radix`/`bytes_in_ram`. So memory-pressure events surface as a counter bump in `/api/ps cache_metrics` instead of looking like nothing happened. Regression test in `test_bulk_eviction_counts_toward_evictions_ram`. 3306 tests pass, ruff clean.

dpalmqvist:
@claude review the changes on this PR

dpalmqvist:
Thanks for the careful look. All three addressed in `22fb469`:

- **Finding 1**: `find_by_prefix` no longer carries the dead `depth < min_prefix_tokens` clause. Comment explains the invariant.
- **Finding 2**: `_cleanup_disk` now always updates `bytes_on_disk` and only runs the eviction loop when `disk_max_bytes` is set. So `OLMLX_PROMPT_CACHE_DISK=true` without a cap surfaces an accurate disk footprint on `/api/ps`. Regression test in `test_cleanup_disk_updates_bytes_without_size_cap`.
- **Finding 3**: `takeover`'s dead sync disk write is now an `assert evicted_id is None` — the pop-then-insert invariant prevents an LRU eviction, and a future break will surface loudly rather than silently stalling the event loop.

3307 tests pass, ruff clean.

To your direct question on what's left: nothing important. Two findings were observability/cleanup, one was a defensive branch that couldn't fire. No correctness, security, or perf concerns remain. The follow-up items the spec explicitly deferred (KV cache cloning for concurrent siblings, disk-spilled entries indexed in the radix, rescuing #284/#343) are out of scope for this PR.

dpalmqvist:
Addressed in `5e9fab5` (after a rebase onto `origin/main`, which had advanced with #389).

`bytes_on_disk` now stays consistent across every disk-state-changing path: `remove`, `clear_disk`, `clear`, `_read_from_disk`'s error branch, `_load_from_disk`'s restore, and `async_get`'s restore. A new `_refresh_disk_bytes` helper does a cheap directory scan after single-file unlinks (negligible cost for slot-cap-sized caches); bulk wipes (`clear`/`clear_disk`) zero the counter directly. Regression tests in `test_bytes_on_disk_decrements_on_remove` and `test_bytes_on_disk_zero_after_clear`.

3336 tests pass, ruff clean.

dpalmqvist:
Addressed in `9bbf143` (after rebase onto `origin/main` — picks up #392).

- **Finding 1** (`clear_disk` zeroing despite unlink failures): now calls `_refresh_disk_bytes()` instead of setting to 0, so any files that survived an `OSError` stay reflected in `bytes_on_disk`. Regression test in `test_clear_disk_recomputes_metric_after_unlink_failure`.
- **Finding 2** (stale spec): updated the `PrefixCacheIndex` section of `docs/superpowers/specs/2026-05-28-radix-prompt-cache-design.md` to match the shipped trie — `terminal_cache_ids: set[str]`, set-union insert, set-discard remove, subtree-DFS lookup with `min_depth` short-circuit. The spec now describes the actual behaviour rather than the bug that was fixed during PR review.

3349 tests pass, ruff clean.

**Nothing important left.** Three review rounds, all low-severity, all observability/documentation cleanups. The feature has been called correct twice (aider's last summary: "no high-signal bugs"; Claude bot: "the feature logic is correct"). Out-of-scope items per the spec: KV cloning for concurrent siblings, disk-spilled entries in the radix, rescuing #284/#343 caches.

dpalmqvist:
Both fixed in `5742cdf`:

- **`_refresh_disk_bytes` on the event loop**: introduced a small `_unlink_and_refresh(file_path)` helper that does the unlink and the directory scan in one shot, called via `asyncio.to_thread`. Both `async_get` unlink sites use it now — no more sync I/O on the loop.
- **Stale `async_find_by_prefix` in spec**: updated three references (method signature, `_setup_prompt_cache` snippet, test list) to the shipped sync `find_by_prefix`. Added a one-liner explaining why it's sync (lookup is pure in-memory radix walk + dict get — no I/O to offload).

3349 tests pass, ruff clean.

Issue #223, opened 2026-04-14
Severe memory thrashing when loading large models while prior models remain resident


dpalmqvist:
## Summary
During a benchmark sweep that loads multiple large MLX models in sequence, the server drops to essentially zero throughput (observed: **4 tokens in 190 seconds, ~0.02 tok/s**) and appears stalled for minutes at a time. The trigger is a new model load while earlier models are still in VRAM — LRU eviction + prompt-cache eviction does not free memory fast enough, so Metal falls back to swap.

## Observed sequence
Running the API bench sweep (models loaded in order, most recent at bottom):
1. `mlx-community/Qwen3-4B-4bit` (~7 GB) — completed normally
2. `mlx-community/gemma-4-26b-a4b-it-4bit` (~27 GB) — completed normally
3. `mlx-community/Qwen3.5-35B-A3B-4bit` (~36 GB) — hung (separate issue #222)
4. Server restarted manually
5. `mlx-community/Qwen3.5-27B-4bit` (~48 GB) — extreme slowdown on load; first generation after warmup logged `Generation complete: 12 prompt tokens (0.1 tok/s), 4 tokens generated (0.0 tok/s), 190.40s total`

Relevant log excerpt (model 5, first cell after warmup):
```
[INFO] ... Generation complete: 12 prompt tokens (0.1 tok/s), 4 tokens generated (0.0 tok/s), 190.40s total
[WARNING] olmlx.engine.inference: [...] Memory pressure high, evicting prompt caches to free GPU memory
[INFO] olmlx.engine.turboquant_cache: Created TurboQuant KV cache: 16/64 cache entries quantized, 4-bit, head_dim=256
[INFO] olmlx.engine.inference: Prompt cache init: creating fresh cache for 22 tokens
<silence for 6+ minutes>
```
Unlike issue #222, `/api/version` remains responsive here — only inference is crawling.

## Why it matters for benchmarking
Any bench that covers more than one large model back-to-back will spend most of its time in this thrashing state instead of producing meaningful throughput numbers. Workaround is to restart the server between models.

## Suggested investigation
- Verify `ModelManager` LRU actually unloads prior models before a large load, and that `mx.metal.clear_cache()` (or equivalent) is called at the right point.
- Consider an "exclusive load" mode: if the incoming model's estimated footprint plus resident models would exceed `OLMLX_MEMORY_LIMIT_FRACTION`, unload everything first rather than partially.
[...]

Issue #315, opened 2026-05-14
Event loop blocked in _evict_lru_if_needed during executor.shutdown


dpalmqvist:
## Background

PR #301 fixed the resource leak on LRU eviction and keep-alive expiry by routing all lifecycle exits through `ModelManager._close_loaded_model`, which calls `executor.shutdown(wait=True)` on the FlashWeightStore (32 threads) and Prefetcher (16 threads) pools.

`_expire_stale` was restructured to pop expired entries under `self._lock`, then close them after releasing it — so concurrent `ensure_loaded()` callers aren't stalled.

`_evict_lru_if_needed` could not be decoupled the same way: it's a synchronous method called while `ensure_loaded()` holds `self._lock`. So `_close_loaded_model()` (and its `executor.shutdown(wait=True)`) runs while the lock is held. Every other coroutine on the event loop is stalled until the pool drains.

In practice the shutdown of idle pools is near-instant — flagged by multiple round-4/5/9 reviews on #301 as a known issue but non-blocking for the leak fix.

## Why it's worth fixing

- 48 threads being joined synchronously on the event loop is a real latency spike if any IO is in-flight at the moment of eviction.
- Eviction happens on the path that callers most need to be responsive (a new `ensure_loaded()` triggered the eviction in the first place).
- The pre-PR baseline only ran `speculative_decoder.close()` here (GDN patch teardown, no executor drain), so this is a new source of blocking introduced by #301.

## Suggested approach

Either:
- Pop the evicted entry inside the lock, dispatch `_close_loaded_model` via `asyncio.to_thread` (or a deferred-cleanup queue) after releasing the lock — coordinating with the subsequent Metal allocator flush so the new load doesn't allocate against unfreed memory.
- Or change `_evict_lru_if_needed` to be `async` and await `asyncio.to_thread(self._close_loaded_model, evicted)` from inside the lock — the lock is still held, but at least other coroutines can run during the thread's work.

Both options need careful Metal allocator coordination: the incoming load can't allocate until the prior close has drained.

## References

- PR #301 round-4 reviewer note: "noting it but not blocking on it"
- PR #301 round-9 reviewer note: "Worth tracking as a follow-up issue (especially with 32 weight-store + 16 prefetch threads now shut down on eviction)"

PR #392, opened 2026-05-29
fix(model_manager): evict idle models + reserve inference headroom under memory pressure (#223)

dpalmqvist:
## Summary

Closes the two gaps left open by #328 on issue #223 (severe memory thrashing when loading large models while prior models remain resident).

PR #328 freed GPU memory on eviction and added a pre-load hygiene pass, but:
- Every memory-pressure responder still shed **prompt caches only — never an idle model**.
- Eviction stayed **purely count-based** (`_pop_lru_evictees` keys off `max_loaded_models`, not memory).

So with `max_loaded_models > 1` (the "prior models remain resident" bench-sweep case), resident model *weights* were never freed under pressure → Metal swapped to a ~0.02 tok/s crawl. Separately, the admission check was weights-only, so a model that loads just under the limit can still swap once the KV cache allocates on top during decode.

## Changes

1. **Idle-model eviction under pressure** (`ensure_loaded` hygiene pass): after flushing prompt caches, if pressure persists, evict idle (`active_refs == 0`) models LRU-first via the new `_pop_one_idle_lru` + existing `_close_evictees` until pressure clears or nothing idle remains — the "exclusive load" the issue asks for. Models in active use are never evicted; the load then proceeds with the existing warning. Default `max_loaded_models=1` is unaffected (count eviction already unloads the prior model there), so this only engages for multi-resident configs.

2. **Headroom-aware admission** — new `OLMLX_INFERENCE_HEADROOM_FRACTION` (default `0.0`): reserves space below `memory_limit_fraction` for the KV cache / activations. Effective weight budget = `memory_limit_fraction - inference_headroom_fraction`. **Default `0.0` reproduces the legacy weights-only check** — opt-in to address the single-model case (a model that loads near the limit then swaps mid-generation).

## Tests (TDD — written first, watched fail, then implemented)

- `test_evicts_idle_model_when_pressure_persists`
- `test_does_not_evict_active_model_under_pressure`
- `test_inference_headroom_fraction_tightens_admission`
- `test_inference_headroom_default_does_not_reject`

`test_model_manager.py` + `test_config.py`: **273 passed**. Inference/lifecycle/memory-utils: **270 passed, 3 skipped**. `ruff check` + `format --check` clean. CLAUDE.md Memory-safety note updated.

## Note

The issue's *literal* logged repro (step 5) was a **single** ~48 GB model after a restart — that's "model too big for inference headroom," which the new `OLMLX_INFERENCE_HEADROOM_FRACTION` knob addresses but only when set. The default-on win here is the multi-resident idle-model eviction. Headroom is left opt-in (default `0.0`) deliberately, to avoid rejecting models that work fine for short contexts.


dpalmqvist:
## ✅ Manual verification (real server, end-to-end) — PASS

Drove the running server (`uv run olmlx serve`, isolated temp `models.json`, small local models) through the Ollama `/api/generate` + `/api/ps` surface. Thresholds were calibrated against the 64 GB host so the pressure logic trips **without any real swap** (the OOM this fix prevents). Qwen3-4B-4bit measured at a 2159 MB Metal footprint, used as the calibration anchor.

### Headroom-aware admission

| Phase | Config | Result |
|---|---|---|
| 1 | `headroom=0.0`, `limit=0.90` | 4B → **HTTP 200**, loads; `/api/ps` shows it |
| 2 | `headroom=0.88`, `limit=0.90` (→ effective 0.02) | *same* 4B → **HTTP 503**, `/api/ps` empty |

Only the headroom changed between phases. Phase 2 error:

> Model 'small-d:latest' requires ~2159 MB but the memory limit is 1310 MB (2% of 65536 MB system RAM) **(includes a 88% inference headroom reserve; lower OLMLX_INFERENCE_HEADROOM_FRACTION to reclaim it)**. Use a smaller/more quantized model, or increase OLMLX_MEMORY_LIMIT_FRACTION (current: 0.9).

The headroom note only renders when `headroom > 0` (absent in Phase 1).

### Pressure-driven idle-model eviction

`memory_limit_fraction=0.0347` (centers 2159 MB inside the pressure band), **`max_loaded_models=3`** so count-eviction cannot fire at 1 resident model:

1. Load `small-d` → `/api/ps` = `['small-d:latest']`
2. Load `tiny-a` → server log:
   ```
   [WARNING] Metal memory under pressure before loading tiny-a:latest; flushing prompt caches to free GPU memory
   [WARNING] Metal pressure persists; evicting idle model small-d:latest to make room for tiny-a:latest
   [INFO]    Loading model tiny-a:latest from mlx-community/Qwen2.5-Coder-0.5B-Instruct-4bit
   ```
3. `/api/ps` → `['tiny-a:latest']` — `small-d` evicted, `tiny-a` resident. Both requests HTTP 200.

This went through the new `_pop_one_idle_lru` path, not `_pop_lru_evictees` (only 1 resident, cap 3).

### 🔍 Probe — eviction is pressure-driven, not count-driven

Same `max_loaded_models=3` but a roomy `limit=0.90`: loaded both models → `/api/ps` = `['small-d:latest', 'tiny-a:latest']`, **zero** pressure/eviction log lines. Eviction does not fire merely because a 2nd model loaded — only under real pressure.

### Notes
- The pressure band is inherently ~10% wide (admission `1.0×limit` vs pressure `0.9×limit`) — that's why the eviction demo needed calibration. It's a property of the existing pressure threshold, not this change; in real use the trigger is a genuinely oversized resident set well past the threshold.
- Did **not** exercise the active-model-protection path (`active_refs > 0` skipped) live — needs concurrent in-flight load + request orchestration. Covered by unit test `test_does_not_evict_active_model_under_pressure`.
- `/api/ps` `size`/`size_vram` reports on-disk dir size (~7.9 GB for the 4B) while the admission log reports actual Metal footprint (2159 MB) — pre-existing, unrelated, but they measure different things.

olmlx/config.py
```diff
@@ -93,6 +93,14 @@ class Settings(BaseSettings):
     default_keep_alive: str = "5m"
     max_loaded_models: int = 1
     memory_limit_fraction: Annotated[float, Field(gt=0, le=1.0)] = 0.75
+    # Fraction of system RAM reserved below ``memory_limit_fraction`` for the
+    # KV cache and activations that allocate on top of model weights during
+    # decode.  The model-load admission check uses an effective weight budget
+    # of ``memory_limit_fraction - inference_headroom_fraction``.  Default 0.0
+    # preserves the legacy weights-only check; raise it (e.g. 0.1) on machines
+    # where a model that loads near the limit then swaps mid-generation
```
dpalmqvist, line 101: Fixed in efdb7e1. Added a startup `model_validator` (`validate_inference_headroom`) that rejects `inference_headroom_fraction >= memory_limit_fraction` with an actionable message, so a zero/negative effective budget fails fast at startup instead of silently 503-ing every model load. Covered by `test_inference_headroom_rejects_when_ge_limit` / `_when_equal_limit` / `_below_limit_ok`.

olmlx/engine/model_manager.py
```diff
@@ -1377,7 +1395,40 @@ async def ensure_loaded(
...
+                # evict idle (active_refs == 0) models LRU-first until it
+                # clears or nothing idle remains — the "exclusive load" the
+                # issue asks for.  Models in active use are left alone (they
+                # cannot be freed); the load then proceeds with a warning.
+                pressure = memory_utils.is_memory_pressure_high(
+                    settings.memory_limit_fraction
+                )
+                while pressure:
+                    async with self._lock:
+                        idle = self._pop_one_idle_lru()
```
dpalmqvist: Fixed in efdb7e1. `_pop_one_idle_lru` now takes an `exclude` arg and the pressure-eviction loop passes `normalized`, so the model this caller is about to load is never popped — if a concurrent caller loaded it during the lock-release window (active_refs still 0), it's left intact and our second lock section returns it instead of loading a duplicate. Covered by `test_pop_one_idle_lru_excludes_named_model` / `_returns_none_when_only_excluded`.

Note: the broader pop/close window (closing some *other* idle model a different caller just received before it bumps active_refs) is pre-existing in the count-based `_pop_lru_evictees` and shared by the #315 pop/close-split design — not introduced here, so I scoped this fix to the self-eviction + duplicate-load case you flagged.

olmlx/engine/model_manager.py
```diff
@@ -1377,7 +1403,44 @@ async def ensure_loaded(
...
+                # what push Metal into swap once a new large model loads on
+                # top — the sub-token/sec thrash in issue #223.  Count-based
+                # eviction (_pop_lru_evictees) won't touch them because the
+                # count is still under the limit.  So if pressure persists,
+                # evict idle (active_refs == 0) models LRU-first until it
+                # clears or nothing idle remains — the "exclusive load" the
+                # issue asks for.  Models in active use are left alone (they
+                # cannot be freed); the load then proceeds with a warning.
+                pressure = memory_utils.is_memory_pressure_high(
+                    settings.memory_limit_fraction
```
dpalmqvist: Fixed in 5508eca. Added `Settings.effective_load_budget_fraction` (`max(0, memory_limit_fraction - inference_headroom_fraction)`) as a single source of truth and routed both the admission check and all three pre-load pressure checks through it, so an idle model in the `[effective_budget, 0.9·limit]` gap now triggers eviction before the load is attempted. Default headroom 0.0 → effective == raw limit, so existing behaviour is unchanged. Covered by `test_pressure_check_uses_effective_budget_fraction` (asserts the trigger uses 0.45, not 0.75) and the `effective_load_budget_fraction` property tests.

I left the inference-time KV-flush trigger (`inference.py`) on the raw `memory_limit_fraction` deliberately: that path frees KV for an already-admitted model where the full limit is the real ceiling — headroom is specifically an admission-time reserve for the KV that will grow.

olmlx/engine/model_manager.py
```diff
@@ -1377,7 +1410,40 @@ async def ensure_loaded(
...
+                # top — the sub-token/sec thrash in issue #223.  Count-based
+                # eviction (_pop_lru_evictees) won't touch them because the
+                # count is still under the limit.  So if pressure persists,
+                # evict idle (active_refs == 0) models LRU-first until it
+                # clears or nothing idle remains — the "exclusive load" the
+                # issue asks for.  Models in active use are left alone (they
+                # cannot be freed); the load then proceeds with a warning.
+                pressure = memory_utils.is_memory_pressure_high(budget)
+                while pressure:
+                    async with self._lock:
```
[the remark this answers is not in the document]
dpalmqvist: Confirmed real, and tracked as a dedicated follow-up: #393. It's a pre-existing race class — `ensure_loaded()` returns before the caller's `_inference_ref` increments `active_refs`, and the inference paths await (`_acquire_inference_lock`, `_setup_prompt_cache`) in between — shared by the count-based `_pop_lru_evictees` and the background `_expire_stale`, not introduced here.

I prototyped the fix (pin-on-load: `ensure_loaded(pin=True)` increments `active_refs` under the lock; `_inference_ref(adopt=True)` adopts+releases it). The core mechanism is ~40 lines and clean, but closing it fully requires try/finally release-guards across all 7 inference consumers (the awaits sit before `_inference_ref`), and a slipped release leaks a pin → a permanently un-evictable model (worse than the race). Given that hot-path surface + failure mode, it's landing as its own reviewed PR (#393) rather than expanding this memory-pressure fix. This PR's actual #223 thrash fix + the three addressed findings are unaffected.

PR #394, opened 2026-05-29
fix(model_manager): pin-on-load to close the model handoff race (#393)

### LINKED ISSUE #393: Close the model handoff race: pin-on-load so eviction can't close an in-use model
## Background

Split out from PR #392 (memory-pressure idle-model eviction, issue #223). The aider review on that PR flagged a real concurrency hazard, confirmed against the code:

`ensure_loaded()` returns a `LoadedModel` to a caller, but the caller does not increment `active_refs` until later — inside `_inference_ref`. The inference paths `await` **before** that point (`_acquire_inference_lock`, `await _setup_prompt_cache`, `_kv_cache_preflight_check`), so during those awaits a *concurrent* `ensure_loaded` can run its eviction (count-based `_pop_lru_evictees`, the new pressure-based `_pop_one_idle_lru`, or the background `_expire_stale`), select this model as idle (`active_refs == 0`), pop it and `_close_evictees()` it — closing/nulling resources the first caller is about to use.

This is **pre-existing** (documented as "safe" in `inference.py:_inference_ref` and `model_manager.py:_expire_stale`) and shared by all three eviction paths. Real-world blast radius is mostly **Flash/Flash-MoE** models (closing weight-store FDs mid-stream); plain models capture the model object by reference and largely survive eviction-nulling.

## Fix (prototyped on `proto/pin-on-handoff`)

**Core mechanism (done, ~40 lines, backward-compatible, 5 tests passing, full suite green):**
- `LoadedModel.acquire_ref()` / `release_ref()`
- `ensure_loaded(..., pin=False)` — when `pin=True`, increments `active_refs` **under the lock** at all 3 return points, so the returned model is pinned atomically (no window)
- `_inference_ref(..., adopt=False)` — when `adopt=True`, adopts the existing pin (no re-increment) and releases on exit

Both params default off → existing behavior unchanged.

**Remaining integration (the bulk + the risk):** update all 7 inference consumers (`generate_completion`, `generate_chat`, `_stream_completion`, `_full_completion`, `generate_embeddings`, `generate_transcription`, anthropic `/v1/messages`) to `ensure_loaded(pin=True)` + `_inference_ref(adopt=True)`, each with a try/finally that releases the adopted pin on **every** error path that bypasses the `with` (lock-acquire timeout, prompt-cache `MemoryError`, etc.). Projected ~150–300 lines, mostly body re-indents in hot-path functions.

### Hazards to handle in the dedicated PR
- **Leaked pin = permanently un-evictable model** (`"All loaded models are in use"`) if any release path is missed — worse than the race. TDD every release path.
- **Never-started streaming async-generator** can't run its `finally`; the pin handoff for streaming needs explicit handling (wrapper that releases on close, or accept reliance on the asyncio async-gen finalizer — decide explicitly).
- Keep the count-based `_pop_lru_evictees` and `_expire_stale` consistent (they share the predicate).

## Acceptance
- A model returned by `ensure_loaded(pin=True)` cannot be evicted/closed by a concurrent load during the await window (test with a forced concurrent eviction).
- Exactly-once release on all paths: normal generation, error before `_inference_ref`, client disconnect mid-stream, never-started generator.
- No `active_refs` leak under a load/evict stress loop.

dpalmqvist:
Closes #393. Split out of #392 (memory-pressure eviction, #223) — see #393 for the full background and the aider review thread that surfaced this.

## The race

`ensure_loaded()` returns a `LoadedModel` to a caller, but the caller doesn't increment `active_refs` until later (inside `_inference_ref`). The inference paths `await` **before** that point (`_acquire_inference_lock`, `await _setup_prompt_cache`, `_kv_cache_preflight_check`), so during those awaits a concurrent `ensure_loaded` can run its eviction, classify this model as idle (`active_refs == 0`), pop it and close its resources — breaking the first caller (notably Flash/Flash-MoE, whose weight-store FDs get closed mid-stream). Pre-existing and shared by `_pop_lru_evictees` and `_expire_stale`.

## This PR — core mechanism (backward-compatible)

- `LoadedModel.acquire_ref()` / `release_ref()`
- `ensure_loaded(..., pin=False)` — `pin=True` increments `active_refs` **under the lock** at all 3 return points → the returned model is pinned atomically, no window
- `_inference_ref(..., adopt=False)` — `adopt=True` adopts the existing pin (no re-increment), releases on exit

Both params default off, so every existing call site is unchanged. 5 new tests; full suite green (510 passed).

## Remaining (draft → ready): integrate the 7 consumers

Each inference consumer must switch to `ensure_loaded(pin=True)` + `_inference_ref(adopt=True)` with a try/finally that releases the adopted pin on **every** error path that bypasses the `with`:

- [ ] `generate_completion` (stream + full delegation)
- [ ] `generate_chat`
- [ ] `_stream_completion` (pre-`with` raises: lock-acquire timeout, deferred-cleanup, sync)
- [ ] `_full_completion`
- [ ] `generate_embeddings`
- [ ] `generate_transcription`
- [ ] anthropic `/v1/messages`
- [ ] **never-started streaming generator** handling (its `finally` can't run) — wrap or accept reliance on the asyncio async-gen finalizer; decide explicitly
- [ ] stress test: load/evict loop asserts no `active_refs` leak

**Failure mode to guard against:** a missed release leaks a pin → that model becomes permanently un-evictable (`"All loaded models are in use"`). Every release path gets a test first.

Draft until the integration + leak-stress tests land.

### DIFF
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -1284,18 +1284,16 @@ def _inference_ref(
 
     When *adopt* is True the caller obtained ``lm`` via
     ``ensure_loaded(..., pin=True)``, which already incremented ``active_refs``
-    under the lock — so this context manager does NOT increment again; it
-    merely *adopts* that existing pin and releases it on exit. This closes the
-    handoff window: the model is pinned continuously from inside
-    ``ensure_loaded`` (under the lock) through the awaits that precede this
-    block (inference-lock acquisition, async prompt-cache setup) until exit.
-    The caller is responsible for releasing the pin (via ``lm.release_ref()``)
-    on any error path that bypasses this context manager.
+    under the lock. This context manager does NOT increment on entry and does
+    NOT release on exit; the caller's outer try/finally owns the single pin
+    release. This avoids races where an exception path here and the caller's
+    cleanup could double-release. Expiry refresh still runs on exit so the
+    model's deadline reflects the inference that just happened.
 
-    With *adopt* False (the legacy path) this increments on entry as before;
-    a small unprotected window then exists between ``ensure_loaded`` and here,
-    but that path is only used where the caller does not hold the model across
-    awaits.
+    With *adopt* False (the legacy path) this increments on entry and releases
+    on exit as before; a small unprotected window then exists between
+    ``ensure_loaded`` and here, but that path is only used where the caller
+    does not hold the model across awaits.
 
     Bug #118: Python ``+=`` on int is not atomic. Concurrent async tasks can
     race on ``active_refs``. Use the model's ``_active_refs_lock`` to protect
@@ -1310,7 +1308,8 @@ def _inference_ref(
     try:
         yield
     finally:
-        lm.release_ref()
+        if not adopt:
+            lm.release_ref()
         # Refresh expiry so the model doesn't expire immediately after inference.
         # Honour per-request keep_alive when available (fix #338).
         # Falls back to settings.default_keep_alive when no per-request value
@@ -1326,6 +1325,31 @@ def _inference_ref(
             lm.expires_at = time.time() + ka
         else:
             lm.expires_at = None
+
+
+async def _release_pin_after_gen(
+    gen: AsyncGenerator[dict, None], lm: LoadedModel
+) -> AsyncGenerator[dict, None]:
+    """Forward ``gen`` and release ``lm``'s pin on any generator exit.
+
+    Used by the streaming entry points to release the pin acquired by
+    ``ensure_loaded(pin=True)`` once the underlying generator finishes
+    (normally, via exception, or via aclose on client disconnect). The
+    inner generator's ``_inference_ref(adopt=True)`` does NOT release —
+    only this wrapper does — so the pin is released exactly once.
+
+    The finally explicitly closes ``gen`` so the wrapped generator's own
+    finally blocks (e.g. the inference-lock release in
+    ``_stream_completion``) fire when the client disconnects mid-stream.
+    """
+    try:
+        async for chunk in gen:
+            yield chunk
+    finally:
+        try:
+            await gen.aclose()
+        finally:
+            lm.release_ref()
 
 
 def _build_generate_kwargs(options: dict | None, is_vlm: bool = False) -> dict:
@@ -1976,7 +2000,7 @@ async def generate_completion(
     stats = TimingStats()
 
     with Timer() as load_timer:
-        lm = await manager.ensure_loaded(model_name, keep_alive)
+        lm = await manager.ensure_loaded(model_name, keep_alive, pin=True)
     stats.load_duration = load_timer.duration_ns
 
     # /api/generate defaults thinking OFF when unspecified (None), unlike the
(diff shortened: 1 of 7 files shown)


dpalmqvist:
Fixed in \`b0f4e6e\`. The pin is now acquired BEFORE the \`_probe_cache_capabilities\` await, with a try/except that releases the pin on probe failure so the outer cleanup's \`active_refs == 0\` pop branch still fires normally.

Regression test in \`tests/test_model_manager.py::TestModelManager::test_ensure_loaded_pin_survives_unload_during_probe\`: races an \`ensure_loaded(pin=True)\` against a concurrent \`unload\` while the probe is stalled on an asyncio event, asserts \`unload\` raises \`ActiveRequestsError\` and the model survives. Confirmed failing before the fix, passing after.

3355 tests pass, ruff clean.

dpalmqvist:
Wired the pin/adopt mechanism into all five inference paths in \`c94fbc1\`. The handoff race is now closed for every request type.

**Touchpoints**:
- \`generate_completion\` / \`generate_chat\`: \`ensure_loaded(pin=True)\` → \`_full_completion\`/\`_stream_completion\` get \`adopt_pin=True\`. Non-stream releases via \`finally\`; stream is wrapped in \`_release_pin_after_gen\` which propagates \`aclose\` to the inner generator (so the inference lock + cache cleanup still run) and releases the pin exactly once.
- \`generate_embeddings\` / \`generate_transcription\`: \`ensure_loaded(pin=True)\` → \`_inference_ref(adopt=True)\` → function-level \`try/finally\` releases.
- Anthropic count_tokens (\`routers/anthropic.py:816\`): same pattern.

**Contract change**: \`_inference_ref(adopt=True)\` no longer releases the pin on exit — the caller's outer \`try/finally\` owns the single release. This avoids the double-release race where \`_inference_ref\`'s finally and the caller's exception handler would both try to release on an early failure. Updated \`test_inference_ref_adopt_does_not_double_count\` to reflect the new contract.

**Regression test**: \`tests/test_prompt_cache.py::TestPinReleasedAcrossStreamingHandoff\` — streams to the first \`cache_info\` chunk, asserts \`active_refs == 1\` mid-stream, calls \`aclose()\`, asserts \`active_refs == 0\` after. Caught the bug where my first wrapper didn't propagate \`aclose\` to the inner gen and left the inference lock held.

3356 tests pass, ruff clean. Known residual: if a stream generator is created but never iterated, the pin would leak — but FastAPI's \`StreamingResponse\` always iterates immediately, so this isn't reachable in practice.

dpalmqvist:
Fixed in \`2d43ea6\`. The outer try in \`generate_completion\` and \`generate_chat\` now wraps from RIGHT AFTER \`ensure_loaded(pin=True)\` through the dispatch, so an exception in the chat-template or kwargs setup releases the pin via \`except BaseException: lm.release_ref(); raise\` instead of leaking it.

The three exit paths share a single \`pin_released_or_transferred\` flag:
- Stream: set after the wrapper is constructed (ownership transferred to \`_release_pin_after_gen\`).
- Non-stream: set inside the inner \`finally\` after manual release.
- Outer except: releases only when the flag is False.

Regression test in \`TestPinReleasedAcrossStreamingHandoff::test_pin_released_when_setup_raises_before_dispatch\`: stubs \`apply_chat_template\` to raise RuntimeError, asserts the exception propagates AND \`active_refs == 0\` after.

3357 tests pass, ruff clean.

Issue #396, opened 2026-05-29
Qwen3-Next-80B-A3B: no prompt cache reuse (hybrid SWA + Gated-DeltaNet not unblocked by checkpoint port)


dpalmqvist:
## Summary

`mlx-community/Qwen3-Next-80B-A3B-Instruct-4bit` (and the rest of the Qwen3-Next family) shows zero cross-request prompt-cache reuse and a visible sawtooth memory-pressure pattern: full prefill on every request, KV-cache growth then release each cycle.

This is distinct from issues #284 (Qwen3.5 ArraysCache Metal-stream crash) and #343 (RotatingKVCache non-trimmable) because Qwen3-Next combines **both** failure modes in the same model — 36 Gated-Attention-with-SWA layers (`RotatingKVCache`) interleaved with 12 Gated-DeltaNet recurrent layers (`ArraysCache`). The checkpoint-mechanism port planned to address #284 and #343 covers each family individually but does not compose for the mixed SWA + SSM layout.

## Affected models

- `mlx-community/Qwen3-Next-80B-A3B-Instruct-4bit` (and other quants)
- Any future Qwen3-Next-derived model (Thinking variant, fine-tunes)

## Observed behavior

Visible from system memory pressure during repeated agentic-style requests sharing a large prefix: each request prefills the full prompt, KV cache for the SWA layers grows over the request, the cache is dropped on completion, the next request repeats from zero. The non-wired RAM footprint cycles by several GB per request on top of the ~42 GB wired weights.

In olmlx terms: `_probe_cache_capabilities` correctly classifies this model as `supports_cache_persistence = False` because its layer list contains both `RotatingKVCache` (per #343 fold) and `ArraysCache` (per #284 guard).

## Upstream context

mlx-lm issue [ml-explore/mlx-lm#980](https://github.com/ml-explore/mlx-lm/issues/980) covers the same systemic gap. It was closed after [PR #1072](https://github.com/ml-explore/mlx-lm/pull/1072) (\"Batch generation refactoring and various fixes\") + [PR #911](https://github.com/ml-explore/mlx-lm/pull/911) (\"Better caching in the server\") landed — a server-side prompt-trie + checkpoint mechanism that fixes Qwen3.5-9B and Nemotron-H. Those PRs do **not** address the Qwen3-Next composition specifically:

[...]

PR #397, opened 2026-05-29
Message-boundary checkpoint prompt cache (fixes #284, #343; tracks #396)

### LINKED ISSUE #284: GPU stream error with Qwen3.5: VLM detection + streaming+tools+prompt cache

### LINKED ISSUE #343: RotatingKVCache models get near-zero prompt-cache reuse in real chat flow

dpalmqvist:
## Summary

Ports mlx-lm's message-boundary checkpoint mechanism into olmlx's `PromptCacheStore`, enabling cross-request prompt-cache reuse for hybrid models the existing flat path couldn't safely persist:

- **Pure SSM / `ArraysCache` layouts** (Qwen3.5 text tower, Nemotron-H, Jamba) — closes #284. Eager `mx.eval` before snapshot materializes the lazy `gated_delta_kernel` outputs so the stored state is safe to reuse from any worker thread.
- **Pure sliding-window / `RotatingKVCache` layouts** (Gemma 3, GPT-OSS) — closes #343. The strict-prefix lookup never needs trim, sidestepping the post-wrap non-trimmability that previously discarded every multi-turn cache.

**Mixed `RotatingKVCache + ArraysCache` layouts** (Qwen3-Next, #396) are explicitly excluded by the probe and continue to do full prefill per request, pending the composition work tracked in that issue.

## What this changes

- New `olmlx/engine/prompt_cache/checkpoint.py`: `SegmentedPrompt`, `Segment`, `snapshot_cache_for_persistence()`.
- `PromptCacheStore`: new `insert_checkpoint()` / `fetch_nearest()` API and `PrefixCacheIndex.find_strict_prefix()`.
- `CachedPromptState`: gains `cache_type` and `is_checkpoint` (backcompat defaults preserve existing call sites).
- `engine/inference.py`: new `tokenize_segmented_chat()`, `_drive_segmented_prefill()`, `_setup_via_checkpoint_path()`. `_setup_prompt_cache` accepts new optional `messages` / `tokenizer` kwargs and routes through the checkpoint path when `lm.uses_checkpoint_persistence`. `_store_prompt_cache_after_generation` early-returns for checkpoint-path requests.
- `engine/model_manager.py`: `LoadedModel.uses_checkpoint_persistence` field; `_PERSISTABLE_CACHE_CLASSES_WITH_CHECKPOINT`, `_EXCLUDED_MIXED_LAYER_PAIRS`, `_cache_supports_checkpoint_persistence`. Probe sets the flag and OR-ins to `supports_cache_persistence`. Five existing `TestProbeCacheCapabilities` tests updated to reflect the new contract.
- `CLAUDE.md`: updated to describe the new path.
- Implementation plan committed at `docs/superpowers/plans/2026-05-29-prompt-cache-checkpoint-port.md`.

## Plan deviation

Phase 3's router plumbing (4 router updates) was skipped — `generate_chat` already has `messages` and the tokenizer, so segmentation is computed in the engine rather than each router. Routers are untouched.

## Notable issues caught in final review and fixed in `ed07a85`

- **Critical KV-depth bug in `_drive_segmented_prefill`**: the last-segment snapshot would store `tokens=flat[:N]` while the cache KV state only covered `N-1` positions (the trailing token is reserved for `stream_generate`'s decode seed). On a future warm-start with an extended prompt, `fetch_nearest` would return `already_covered=N` and the drive loop would skip past the missing `flat[N-1]`, corrupting attention. Fixed by guarding the snapshot with `if boundary < len(flat)`. Regression test `test_drive_does_not_snapshot_last_segment_boundary` covers it.
- Stale comments in `_setup_prompt_cache` and `_store_prompt_cache_after_generation` describing pre-checkpoint behavior were updated.
- Redundant `not memory_too_high` guard in the checkpoint dispatch removed.

## Test plan

- [x] Unit tests:
  - `tests/test_prompt_cache_checkpoint.py` (5 tests)
  - `tests/test_prompt_cache_store_multi.py` (5 tests)
  - `tests/test_prompt_cache_radix.py` (33 tests — includes 3 new for `find_strict_prefix`)
  - `tests/test_inference_checkpoint_drive.py` (7 tests including last-segment regression)
  - `tests/test_probe_cache_capabilities_v2.py` (7 tests — 1 default + 6 parametrized layouts including #396 exclusion)
- [x] Full suite: `uv run pytest -q` → 3385 passed, 4 skipped.
- [x] `uv run ruff check olmlx tests && uv run ruff format --check olmlx tests` → clean.
- [ ] **Deferred** — real-model end-to-end benchmark on Qwen3.5-9B (or similar small hybrid) measuring TTFT improvement on multi-turn shared-prefix requests. Plan's Task 6.1 — needs a hybrid model loaded on the runner. Recommend manual verification before relying on the new path in production.

## Related

- #284 — ArraysCache cross-thread Metal-stream crash (closed by this PR for non-mixed layouts)
- #343 — RotatingKVCache near-zero reuse from non-trimmability (closed by this PR for non-mixed layouts)
- #365 — radix prefix cache (existing infrastructure that this builds on)
- #396 — Qwen3-Next mixed SWA+GDN composition gap (NOT addressed; explicitly excluded)
- ml-explore/mlx-lm#980 — upstream tracking
- ml-explore/mlx-lm#1072 / #911 — upstream server-side fix that inspired this olmlx-side port


dpalmqvist:
## Benchmark results (updated after `ea34310` + `57ce638`)

Initial benchmark on `mlx-community/Qwen3.5-0.8B-MLX-4bit` surfaced two real issues that have now been fixed:

### Issue 1: Segmentation didn't fire for modern chat templates

`tokenize_segmented_chat`'s prefix-tokenization strategy failed for templates that reject partial prefixes via `{{ raise_exception('No user query found in messages') }}` — i.e., Qwen3, Llama 3, Mistral, Gemma. Every checkpoint-path request silently fell back to a fresh cache.

**Fix (`ea34310`):** replaced with EOS-token boundary detection. Scans the full tokenization for `tokenizer.eos_token_id` (which IS the end-of-message marker in modern chat templates — `<|im_end|>` for Qwen, `<|eot_id|>` for Llama 3, `<end_of_turn>` for Gemma). One segment per detected EOM marker. Template kwargs now propagated through the call chain.

### Issue 2: `usage.prompt_tokens` reported the suffix, not the full prompt

With the cache path handing a 1-token suffix to mlx-lm, `result.prompt_tokens` (and thus OpenAI `usage.prompt_tokens` / Ollama `prompt_eval_count`) reported `1`. Pre-existing latent bug — also affected the flat path on cache hits.

**Fix (`57ce638`):** override `stats.prompt_eval_count` with `len(full_prompt_tokens)` after the inner generate returns. Reports the full request size to callers, as the spec requires.

### Updated bench numbers

Same three-request shared-prefix scenario, `mlx-community/Qwen3.5-0.8B-MLX-4bit`:

| Request | Wall | `usage.prompt_tokens` | Notes |
|---|---|---|---|
| 1 (cold) | 1351 ms | 136 | full prefill |
| 2 (warm) | 225 ms | 139 | system prompt cached |
| 3 (warm) | 235 ms | 142 | system prompt cached |

`/api/ps` after the run:

```json
{
  "cache_id_hits": 0,
  "cache_id_misses": 0,
  "radix_hits": 2,
  "radix_misses": 0,
  "evictions_ram": 0,
  "evictions_disk": 0,
  "bytes_in_ram": 3145728,
  "bytes_on_disk": 0
}
```

2 radix hits on requests 2 + 3, 3 MB of KV cache resident in RAM. The checkpoint path is firing as designed.

Wall-time improvement is modest on a 0.8B model (prefill is cheap anyway) but the per-token accounting confirms 130+ tokens of prefill being skipped on hot calls — the savings scale with prompt size and model size. Recommend running this against Qwen3.5-27B or similar for a more dramatic wall-time delta.

Test suite: 515 passed, 3 skipped, 0 failures. Ruff clean.

dpalmqvist:
## Benchmark — `mlx-community/Qwen3.5-27B-4bit`

Same 3-request shared-prefix scenario as the 0.8B run, with a longer system prompt (487 tokens vs 128) and a much larger model:

| Request | Wall | `usage.prompt_tokens` | Notes |
|---|---|---|---|
| 1 cold (load + prefill) | 10 660 ms | 487 | full cold prefill |
| 2 warm | 2 606 ms | 490 | system cached, 7 new |
| 3 warm | 2 609 ms | 493 | system cached, 10 new |

**Speedup: 4.1× on warm requests.**

`/api/ps`:
```json
{
  "cache_id_hits": 0,
  "cache_id_misses": 0,
  "radix_hits": 2,
  "radix_misses": 0,
  "evictions_ram": 0,
  "evictions_disk": 0,
  "bytes_in_ram": 33554432,
  "bytes_on_disk": 0
}
```

32 MB of KV state checkpointed at the system-message boundary, surviving across the two follow-up requests with no eviction.

Wall-time decomposition for warm requests is roughly 280ms suffix prefill (~7-10 new tokens, drive function) + 2 050ms decode (32 tokens × ~16 tok/s). The prefill savings is what was previously a ~4 second cold prefill of the 480-token system prompt. At longer context — agentic workloads with a multi-KB system prompt — the wall-time delta scales linearly with the saved prefill.

dpalmqvist:
Addressed both aider-code-review findings in `ac21d43`:

**[high/bug]** `_setup_via_checkpoint_path:2390` — fixed. When `fetch_nearest` returns a stored entry, the snapshot now computes `eager_eval` from `_cache_list_contains_lazy_state(cached_state.cache)` rather than hard-coding `False`. Pre-PR disk entries were stored before `snapshot_cache_for_persistence` existed, so we can't assume the lazy `gated_delta_kernel` graph was already materialised. Eager eval on the loading thread closes the #284 Metal-stream crash path for stale entries.

**[medium/bug]** `tokenize_segmented_chat:1750` — fixed. Now normalises the three return-shape variants `apply_chat_template` can produce (flat list, batch-of-1 nested list, `BatchEncoding`-style mapping with `input_ids`), mirroring the pattern already used in `_apply_chat_template_text`. Unrecognised shapes return an empty single segment so the caller's `segmented.flatten() != prompt_tokens` guard cleanly routes to the flat path without raising.

Three regression tests added (`test_tokenize_segmented_chat_handles_batchencoding_dict_return`, `..._handles_nested_list_return`, `..._returns_empty_segment_for_unrecognised_shape`).

Test suite: 465 passed across `tests/test_inference_checkpoint_drive.py tests/test_inference.py tests/test_model_manager.py`. Ruff clean.

dpalmqvist:
@claude review this PR

dpalmqvist:
@claude review it again

PR #404, opened 2026-05-30
fix(prompt_cache): bound checkpoint-drive chunking to two model calls per request

dpalmqvist:
## Summary

- `_drive_segmented_prefill` no longer runs one `model(...)` call per uncovered segment. It now drives in **at most two chunks** — `[already_covered .. deepest_interior_boundary]` then `[deepest_interior_boundary .. trailing-token]` — and snapshots only at the deepest interior boundary.
- Fixes the verbatim-paragraph-repetition regression that appeared on the **second request** of any chat session served by hybrid-SSM MoE models (Qwen3.6-35B-A3B-6bit, Qwen3-Next MoE variants) after #397 enabled the checkpoint path for them.

## Root cause

mlx-lm's `gated_delta_kernel` (GatedDeltaNet linear-attention layers in Qwen3.5 / Qwen3-Next / Qwen3.5-MoE text towers) is **not exactly chunking-invariant**: feeding `[A, B, C]` produces a slightly different recurrent state than feeding `[A]`, then `[B]`, then `[C]` separately. The pre-fix per-segment drive made the number of chunk boundaries grow linearly with conversation depth — every extra turn added another chunk boundary inside the GDN forward.

On dense full-precision targets the drift was invisible. On 6-bit MoE targets it crossed expert-routing thresholds by the second turn and the model derailed into verbatim paragraph repetition (the user-reported symptom).

The snapshot/restore round-trip itself contributes **zero** drift — empirically verified by comparing chunked-then-snapshotted vs chunked-then-not-snapshotted layer states on a real Qwen3.5 model. All of the drift comes from the kernel's chunking sensitivity, not the snapshot deepcopy. So the fix targets the right place: cut the number of chunk boundaries the drive imposes.

## Behavior change

| Request shape | Pre-fix `model()` calls | Post-fix `model()` calls | Snapshots taken |
|---|---|---|---|
| Cold start, 2 segments | 2 | 2 | 1 (unchanged) |
| Warm-start, 4 segments (sys covered) | 3 | 2 | 1 (was 2; deepest only) |
| Warm-start, 6 segments (sys covered) | 5 | 2 | 1 (was 4; deepest only) |
| Warm-start, only final segment uncovered | 1 | 1 | 0 (unchanged) |

Earlier requests' boundary snapshots remain in the store and still serve shallower branched lookups — the sequentially-extending chat case only consults the deepest snapshot anyway, so no warm-start coverage is lost for the common workload.

## Measurements

Qwen3.5-0.8B-MLX-4bit, 79-token, 6-segment workload:

- L0 GDN recurrent-state drift (warm-start drive vs fresh single-chunk prefill): **0.0066 → 0.0025** (~2.7× reduction).
- KVCache layers showed zero chunking sensitivity in either path (concat is associative).
- Generation comparison (warm-start vs fresh `stream_generate`): both paths produce coherent output that doesn't derail. Pre-fix, warm-start drifted noticeably mid-paragraph.

Larger models (more layers, more turns, MoE routing, 6-bit quant) compound the per-boundary drift more aggressively — the practical impact on the user's reported scenario should be substantial.

## Tests

- All 21 existing `test_inference_checkpoint_drive.py` tests pass unchanged (the 2-segment cases they cover have identical chunking shape under the new logic).
- Two new regression tests lock in the new shape:
  - `test_drive_uses_two_chunks_and_one_snapshot_for_multi_segment_request` — 4-segment warm-start asserts exactly 2 model calls and exactly 1 snapshot (at the deepest interior boundary, with the correct role).
  - `test_drive_no_snapshot_when_only_final_segment_remains` — boundary case where `already_covered` skips into the final segment: 1 chunk, 0 snapshots.
- Full `tests/test_inference.py`: 229 passed, 3 slow-skipped.
- All `prompt_cache*` test files together: 165 passed.

## Test plan

- [ ] Pull the branch on the reporter's setup and run the same Opencode scenario with `mlx-community/Qwen3.6-35B-A3B-6bit` for two turns; confirm the second request no longer repeats paragraphs.
- [ ] Smoke-test a few non-hybrid models (Qwen3-8B dense, Gemma) to confirm no regression in their (trim-path) caching.
- [ ] Sanity-check `/api/ps` `cache_metrics` after a multi-turn chat: `radix_hits` should still increment on the second request (warm-start still works), and `_entries` should grow by one snapshot per request rather than (N-1) per request.


dpalmqvist:
@claude review this PR

PR #410, opened 2026-05-31
feat(turboquant): implement capacity cap for KV cache side buffers during trim

dpalmqvist:
Implements a capacity cap in TurboQuantKVCache.trim() that shrinks the side buffers when the offset drops below 50% of the current capacity to bound memory usage during long sessions with heavy trimming.


dpalmqvist:
@claude review this PR

PR #414, opened 2026-06-01
fix: multi-turn tool use for Devstral and sliding-window (SWA) models

dpalmqvist:
## Summary

Fixes two distinct **multi-turn tool-use** failures, found by a new verification harness that drives a real two-round tool conversation against each registered model.

### 1. Devstral / minimal-Mistral templates (HTTP 500 on turn 2)
These templates only allow `user`/`system`/`assistant` roles and `raise_exception` on anything else, so feeding a tool result back as `role:"tool"` crashed `apply_chat_template`:

```
Chat template failed: Only user, system and assistant roles are supported!
```

Fix: new `TemplateCaps.handles_tool_role` detection + `_convert_tool_messages_to_user_text`, which folds tool calls + results into `user`-message text for templates that can't render the tool role. The rewrite is unified ahead of the VLM/text split: `uses_tool_responses` → Gemma `tool_responses`; else `not handles_tool_role` → user-text fold; else pass through (Qwen/Llama/MiniMax unchanged).

### 2. Pure-`RotatingKVCache` (sliding-window) models — gpt-oss, Step-3.5, Gemma 3
With `tools` present, these emitted **coherent-but-unrelated output and skipped tool calls** (e.g. gpt-oss reasoning about "a LaTeX document"). 

Root cause (confirmed by isolation — *not* turboquant, flash_moe, or the prompt): the checkpoint path's **two-chunk prefill** — which exists only to bound GatedDeltaNet/`ArraysCache` drift — **corrupts sliding-window attention** when split at an interior message boundary. Tool definitions add the `system`/`developer` segments that create that boundary, which is why it only triggered with tools (single-segment / no-tools prompts were fine).

Fix: `_is_pure_rotating_cache` (RotatingKVCache present, no ArraysCache) → prefill in a **single** `model()` call with one end-of-prompt snapshot so warm-start reuse still works. ArraysCache / mixed Rotating+Arrays keep the two-chunk path.

## Validation
- **gpt-oss, caching ON** (the real failing config): 2-round tool conversation now correct — `get_user_location` → `get_weather` → "…Paris is 18 °C and sunny." Single-chunk branch fires (incl. warm-start). Was garbage before.
- Devstral: validated against the real chat template (tokenizer-only) — crash before, clean prompt after.
- Step-3.5: same cache class → covered by the fix. Not loaded to validate (its 293 GB flash-MoE read crashes the external drive — separate hardware issue).
- TDD throughout; ruff clean; **3418 unit tests pass**.

## Notes
- `prompt_cache: false` is no longer required for gpt-oss — the code fix handles it. `Qwen3-Coder-Next` / `Qwen3.6-35B-A3B-6bit` keep their overrides (mixed/GatedDeltaNet class, out of scope here).
- `scripts/verify_multiturn_tools.py` is reusable: `python scripts/verify_multiturn_tools.py [--model <id>]`.


dpalmqvist:
@claude review this PR

PR #415, opened 2026-06-01
fix(prompt_cache): run segmented prefill on generation_stream

dpalmqvist:
## Problem

Qwen3-Next-family MoE-quantized targets (`Qwen3-Coder-Next-4bit`, `Qwen3.6-35B-A3B-6bit`) produced **coherent but completely off-prompt output** — raw pretraining-data dumps — on multi-message prompts once the prompt exceeded ~16k tokens, whenever the cross-request prompt cache was enabled. Each had been pinned to `prompt_cache: false` as a workaround.

## Root cause

Not the message-boundary chunking (the previously-suspected GatedDeltaNet "chunking non-invariance"). The real cause is a **Metal stream mismatch**: `_drive_segmented_prefill._run` prefilled the cache on `mx.default_stream`, then handed a 1-token suffix to `stream_generate`, which **decodes on mlx-lm's dedicated `generation_stream`**. That cross-stream materialization of the lazily-built `gated_delta_kernel` recurrent state corrupts MoE expert routing at scale (same hazard family as #284). Single-message requests were unaffected because they skip the drive entirely and let `stream_generate` prefill on `generation_stream`.

This was isolated with a probe matrix at 16k tokens on `Qwen3-Coder-Next-4bit`:

| prefill path | stream | result |
|---|---|---|
| mlx-lm native (no checkpoint path) | generation_stream | clean |
| one big `model()` call | default | derail |
| 512 sub-chunks | default | derail |
| 512 sub-chunks | generation_stream + clear_cache | clean |
| **shipped 2-chunk msg-boundary logic** | **generation_stream + clear_cache** | **clean** |

The last row is decisive — the existing chunking is fine once the stream is right.

## Fix

- Run the prefill forward on `generation_stream` (the stream decode uses). This is the actual fix and is universal to all checkpoint-path models.
- Sub-chunk the GatedDeltaNet/`ArraysCache` prefill at `_PREFILL_CHUNK` (2048) with `mx.eval` + `mx.clear_cache()` per sub-chunk, matching mlx-lm's prefill loop. This also bounds the separate ~70k-token single-forward Metal OOM (123 GB on `Qwen3.6-35B-A3B-6bit`).
- Pure-rotating SWA models (gpt-oss, Step-3.5, Gemma 3) keep their single-call behavior (windowed-attention sensitivity), now also on `generation_stream`.

## Tests

- `test_drive_prefill_runs_on_generation_stream` — asserts prefill enters `generation_stream`, not the default stream (RED before fix).
- `test_drive_subchunks_long_prefill` — asserts long spans are sub-chunked at `_PREFILL_CHUNK`.
- Verified clean end-to-end at 16k on the real shipping path for `Qwen3-Coder-Next-4bit` (4-turn conversation, all on-topic).
- Full checkpoint-drive / prompt-cache / inference suites pass; ruff clean.

## Note

The per-model `prompt_cache: false` workarounds live in the user's runtime `~/.olmlx/models.json` (not in the repo) and have been removed locally for both affected models now that the fix is in.


dpalmqvist:
@claude review this PR

PR #416, opened 2026-06-01
fix(prompt_cache): correct decode tok/s on cache-hit timing

dpalmqvist:
## Problem

Generation throughput logged on a prompt-cache hit was physically impossible:

```
Generation complete: 40491 prompt tokens (24095.8 tok/s), 69 tokens generated (49275.2 tok/s), 1.68s total
```

No LLM decodes 69 tokens in ~1.4 ms. The prefill rate was also overstated.

## Root cause

On a cache hit the streaming path hands mlx-lm only the **uncached suffix**, so `token.prompt_tps` is measured over that suffix. But the streaming loop deliberately overrides `stats.prompt_eval_count` to the **full** prompt size (for transparency to the caller). `_derive_timing_stats` then computed `raw_prompt_ns = full_count / suffix_rate`, pretending the entire prompt was prefilled at the suffix's rate. That inflated the prefill duration far past wall-clock, so the proportional clamp handed ~all of `eval_timer` to prefill and left decode a ~1 ms sliver — producing the bogus 49,275 tok/s, with the prefill rate degenerating into `full_count / eval_timer` (24,095).

The non-streaming path never hit this: it sets `prompt_eval_count = result.prompt_tokens` (the actually-prefilled count, consistent with the rate).

## Fix

Add a `prefilled_count` argument to `_derive_timing_stats` so the prefill **duration** is apportioned against the count the rate was measured over (the suffix), while the API still reports the full prompt size. Decode keeps mlx-lm's real measured rate; the logged prompt rate becomes the genuine prefill throughput. The non-streaming caller passes the same value explicitly for parity.

## Testing

- Added `test_cache_hit_uses_prefilled_count_not_full_prompt` reproducing the inflated-prefill → bogus-decode-rate path (fails before, passes after).
- 14/14 timing tests, 325/325 inference + prompt-cache tests pass.
- ruff check + format clean.

### DIFF
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -602,6 +602,7 @@ def _derive_timing_stats(
     prompt_tps: float,
     gen_tps: float,
     eval_timer_ns: int,
+    prefilled_count: int | None = None,
 ) -> tuple[float, float]:
     """Populate ``stats.prompt_eval_duration`` and ``stats.eval_duration`` from
     mlx-lm's measured prefill/decode rates, with conservative fallbacks for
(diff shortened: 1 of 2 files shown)


dpalmqvist:
@claude Review this change

dpalmqvist:
@claude  Merge the PR

PR #418, opened 2026-06-02
feat: Prometheus /metrics endpoint (#371)

### LINKED ISSUE #371: Prometheus metrics endpoint (/metrics)
## Why

The bench harness already collects most of what we'd want to expose: TTFT, decode tok/s, Flash + MoE hit ratios, speculative acceptance per strategy, prompt-cache hit rate. Surfacing them at `/metrics` makes operating a long-lived instance (especially with experimental knobs on) tractable without re-running benches.

## Scope

- `prometheus_client` integration; mount at `GET /metrics`
- Counters: requests, errors, tokens in/out, speculative accepts/rejects, cache hits/misses
- Histograms: TTFT, inter-token latency, request duration
- Gauges: loaded models, KV cache bytes, Flash cache occupancy
- Surface existing `ExpertCacheStats`, `PrefetchStats`, prompt-cache stats

## Out of scope

- Pushgateway integration
- Multi-instance aggregation (single-user server)

## Acceptance criteria

- Scraped values match `olmlx bench run` reports on the same workload within tolerance
- No measurable latency regression from collection
- Cardinality bounded (no per-request labels that explode the series count)

## Files

- `app.py`
- new `utils/metrics.py`
- all engine stats objects (annotation only)

dpalmqvist:
Closes #371.

Adds a Prometheus-scrapable `GET /metrics` endpoint exposing the operational metrics olmlx already collects, so a long-lived instance can be monitored without re-running `olmlx bench run`.

## Design

Three collection layers in a single `olmlx/utils/metrics.py` sharing one private `CollectorRegistry` (isolated from the global default so tests stay clean):

1. **HTTP middleware** (`MetricsMiddleware`) — `olmlx_http_requests_total{path,method,status}`, `olmlx_http_request_duration_seconds`, `olmlx_http_requests_in_flight`. `path` is the matched **route template** (not raw URL) to bound cardinality. Also sets a `surface_var` ContextVar (ollama/openai/anthropic/audio).
2. **Engine inference instrumentation** — recorded at the two shared `inference.py` seams (`_stream_completion` done-chunk, `_full_completion_inner` return) plus the error path, reading the finalized `TimingStats`: `olmlx_inference_{requests,errors,tokens}_total`, `olmlx_inference_ttft_seconds`, `olmlx_inference_decode_tokens_per_second`, `olmlx_inference_request_duration_seconds`. Speculative proposed/accepted recorded in `speculative_stream._log_stats`, labelled by strategy. Because these read the same `TimingStats` the HTTP responses use, scraped TTFT/decode-tok-s/token counts match `bench run` within tolerance.
3. **Lazy `OlmlxStatsCollector`** (registered in `lifespan`) — reads `ModelManager` at scrape time, so there is **no per-request cost**: point-in-time gauges (loaded models, KV bytes) plus process-global cumulative counters (prompt-cache / flash expert / prefetch hit-miss-eviction) folded through a `CounterAccumulator` so they stay **monotonic across model eviction/reload**.

All labels are bounded (model, surface, strategy, route template) — no per-request labels. Collector is unregistered on shutdown so repeated `create_app()` in tests doesn't raise a duplicate-collector error.

## Acceptance criteria (#371)

- ✅ Scraped values match `bench run` — inference metrics derive from the same `TimingStats`; a unit test pins `_decode_tps(stats) == PromptResult.tokens_per_second`.
- ✅ No measurable latency regression — gauges are pull-at-scrape; push paths are cheap counter incs / histogram observes.
- ✅ Cardinality bounded — bounded label sets only.

Out of scope (per issue): Pushgateway, multi-instance aggregation.

## Tests

New: `tests/test_metrics.py` (12), `tests/test_routers_metrics.py` (2), `tests/test_inference_metrics.py` (2), `tests/test_speculative_metrics.py` (1). Covers the helpers, the monotonic accumulator across a simulated evict/reload, the lazy collector with stub stats objects (and the safe no-store case), the endpoint content-type/body, middleware counting + surface, and the bench-parity assertion.

Full suite: **3488 passed, 1 skipped**. `ruff check`/`format` clean; `pyright` clean on the new modules.

Design spec and implementation plan committed under `docs/superpowers/`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,10 +40,12 @@ olmlx/
 │   ├── anthropic.py    # /v1/messages — Anthropic Messages API
 │   ├── openai.py       # /v1/chat/completions, /v1/completions, /v1/models, /v1/embeddings
 │   ├── audio.py        # /v1/audio/transcriptions — Whisper STT
+│   ├── metrics.py      # GET /metrics — Prometheus exposition
 │   └── *.py            # /api/{chat,generate,tags,show,ps,pull,copy,create,delete,embed,blobs,version} (Ollama)
 ├── schemas/            # Pydantic request/response models per API surface
 └── utils/
     ├── streaming.py    # async_mlx_stream, safe_ndjson_stream
+    ├── metrics.py      # Prometheus registry, helpers, lazy OlmlxStatsCollector
     └── timing.py       # Timer, TimingStats
 ```
 
@@ -84,6 +86,7 @@ olmlx/
 - **Audio transcription**: OpenAI-compatible `/v1/audio/transcriptions` via mlx-whisper. Whisper is a first-class `ModelManager` kind: `_detect_model_kind` matches `model_type == "whisper"` or mlx-whisper dims (`n_mels`/`n_audio_state`); `LoadedModel.is_whisper` guards LLM-only prompt-cache/KV-quant paths. `generate_transcription` injects the managed model into `mlx_whisper.transcribe.ModelHolder` (the module-level singleton — race-free under the serialized lock) and runs transcribe in a worker thread. Formats: `json`, `verbose_json`, `text`, `srt`, `vtt`. `ForceJSONMiddleware` skips `multipart/form-data` so upload boundary survives. **Requires ffmpeg on PATH**. Streaming and diarization out of scope.
 - **Terminal chat**: `olmlx chat` runs in-process via `ModelManager`/`generate_chat()` — no HTTP. Connects to external MCP servers (stdio/SSE) with a full agent loop. MCP config in Claude Desktop format at `~/.olmlx/mcp.json`.
 - **Model load timeout**: `OLMLX_MODEL_LOAD_TIMEOUT` → `ModelLoadTimeoutError` (HTTP 504).
+- **Prometheus metrics** (`utils/metrics.py`, `routers/metrics.py`): `GET /metrics` exposes operational metrics from a private `CollectorRegistry` (not the global default, so tests stay isolated). Three layers: (1) `MetricsMiddleware` records `olmlx_http_*` request/duration/in-flight and sets `surface_var` (ollama/openai/anthropic/audio, derived from path); (2) engine instrumentation reads the finalized `TimingStats` at the two shared `inference.py` seams (`_stream_completion` done-chunk, `_full_completion_inner` return) plus the error path, emitting `olmlx_inference_*` requests/errors/tokens/ttft/decode-tok-s — values are read from the same `TimingStats` the HTTP responses use, so they match `bench run`; speculative proposed/accepted recorded in `speculative_stream._log_stats`; (3) a lazy `OlmlxStatsCollector` registered in `lifespan` reads `ModelManager` at scrape time for point-in-time gauges (loaded models, KV bytes) and process-global cumulative counters (prompt-cache / flash / prefetch) folded through `CounterAccumulator` so they survive model eviction/reload (counters stay monotonic). Labels bounded (model, surface, strategy, route template); no per-request labels. The collector is unregistered on shutdown so repeated `create_app()` in tests doesn't raise a duplicate-collector error.
 
 ## Development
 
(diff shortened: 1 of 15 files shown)


dpalmqvist:
The aider review summary reports no high-signal issues. Its raw reasoning floated one concern worth recording: that `MetricsMiddleware` resets `surface_var` when `call_next` returns — before the streaming response body is consumed — which would mislabel streaming inference metrics as `"unknown"`.

Verified this is **not** a bug. Starlette's `BaseHTTPMiddleware` runs the inner app in a sub-task that copies the context, so the middleware's `finally` reset happens in its own context and does not clobber the value the streaming body sees in the copied context. This is the same mechanism `RequestIDMiddleware` already documents and relies on for request-ID logging during streaming.

Added a regression test (`test_surface_var_survives_into_streaming_body`) that drives a `StreamingResponse` through `MetricsMiddleware` and asserts the body reads the correct surface during iteration.

dpalmqvist:
Addressed the review notes in ee2eada:

1. **Streaming error path skipped recording when `full_prompt_tokens is None`** — fixed. Error recording is no longer nested under the cache-invalidation condition; it now fires for any in-flight real exception (`sys.exc_info()`), so early failures (e.g. prefill OOM, which leaves `full_prompt_tokens=None`) are counted. `GeneratorExit`/`CancelledError` are excluded so normal client disconnects aren't miscounted as errors.

2. **`OlmlxStatsCollector` manager typed `Any`** — replaced with a minimal `_ManagerProtocol` (`get_loaded()` only), avoiding a heavy/circular import of `ModelManager` while restoring type safety.

3. **`CounterAccumulator` reset logic** — verified it does not inflate: folding the live value when `live < last` adds at most the new store's real events. The only inaccuracy is a slight *under*-count if a reloaded store passes the prior total before the first scrape observes the reset. Documented this in the docstring rather than changing the (correct) logic.

Full suite: 3489 passed, 1 skipped.

dpalmqvist:
Addressed the second round of notes in b74ae95:

1. **Non-streaming error path** — `_full_completion`'s `finally` now records `olmlx_inference_errors_total` for any in-flight real exception (excluding `GeneratorExit`/`CancelledError`), symmetric with the streaming path. `_full_completion_inner` still records success exactly once on its normal return, so there's no double count.

2. **Expert-stats `{}` fallback** — the collector now falls back to direct attribute reads (`cache_hits`/`cache_misses`/`load_failures`/`load_calls`) when a stats object lacks `snapshot()`, instead of discarding live values. Added `test_collector_reads_expert_stats_without_snapshot_method`.

3. **Blocking scrape** — `GET /metrics` is now a sync (`def`) endpoint, so FastAPI runs `generate_latest` + the collector walk in its threadpool rather than on the event loop.

Full suite: 3490 passed, 1 skipped.

dpalmqvist:
Addressed two of the three notes in 3ad655a; pushing back on the third.

**Done:**
- **Collector scrape robustness** — `OlmlxStatsCollector.collect()` now emits each model inside a per-model `try/except`, so a malformed stats object on one model can't fail the whole scrape (which would 500 `/metrics`). Added `test_collector_scrape_survives_one_malformed_model`.
- **Brittle test** — relaxed `test_seams_call_observe_inference` from an exact `observe_inference` call count to a presence check (`>= 2` + `error=True`), so adding seams won't break it.

**Pushback — cache-persistence failure after successful generation:**
I'm intentionally leaving this as a recorded success. Per the design spec, `olmlx_inference_errors_total` records *mid-generation* failures; `_store_prompt_cache_after_generation` runs *after* the generation has completed (tokens produced, success already recorded). The generation genuinely succeeded — only an opportunistic KV-cache persist failed. The user-visible request failure (the resulting 5xx) is already captured by `olmlx_http_requests_total{status=...}`, so the two signals stay distinct rather than conflating a post-generation infrastructure error into the generation error rate. Happy to revisit if you'd prefer the generation counter to track request outcome instead of generation outcome.

Full suite: 3490 passed, 1 skipped (local runs have an intermittent native MLX/Metal SIGABRT at teardown unrelated to this change; CI is green).

Issue #372, opened 2026-05-25
OpenTelemetry tracing


dpalmqvist:
## Why

Per-request spans for prefill, draft, verify, SSD prefetch, and disk-cache I/O make slow-request triage tractable, especially with Flash + speculative combos where the bottleneck shifts request-by-request. Logs alone don't show the relative cost of each phase.

## Scope

- Optional OTLP exporter (off by default; opt-in via env)
- Spans around `prefill`, per-step `step`, Flash prefetch, disk cache reads/writes, MCP tool calls (in terminal chat)
- Correlation id propagated to log records
- Trace attributes: model, strategy, accepted/rejected counts

## Out of scope

- Trace propagation across distributed workers (follow-up)
- Auto-instrumentation of the HTTP layer (FastAPI default is fine)

## Acceptance criteria

- A single chat request renders a coherent trace in Jaeger/Tempo with prefill/decode/verify sub-spans
- No measurable latency regression when tracing is disabled (compile-time/import-time only)

## Files

- `engine/inference.py`
- `engine/speculative.py`, `engine/dflash/decoder.py`, `engine/eagle/decoder.py`
- `engine/flash/*`
- `routers/*`

dpalmqvist:
Implemented and shipped in #420 (commit bb091ca, merged 2026-06-03). The PR referenced this issue as `(#372)` in its title rather than with a `Closes` keyword, so it was never auto-closed. `olmlx/utils/tracing.py` exists and the feature is documented in CLAUDE.md. Closing as done.

PR #420, opened 2026-06-03
feat(tracing): OpenTelemetry per-phase spans (OLMLX_TRACING) (#372)

dpalmqvist:
## Summary

Adds opt-in OpenTelemetry tracing that emits per-phase spans for each request, exportable to any OTLP collector, gated behind a single `OLMLX_TRACING` toggle with **zero cost when disabled** (no `opentelemetry` import on the off path).

- **`olmlx/utils/tracing.py`** — the only module that touches `opentelemetry` (lazily). `span(name, **attrs)` returns a shared no-op singleton when disabled; `init_tracing`/`shutdown_tracing` wired into `lifespan`; cross-thread `current_context`/`attach_context`/`detach_context` helpers; a `logging.Filter` stamping `trace_id`/`span_id` onto log records.
- **Span tree:** root `http.request` (`RootSpanMiddleware`) → `inference` → `prefill`/`decode`, plus `spec.{prefill,step,verify}`, `flash.{prefetch,weight_load}`, `cache.disk_{read,write}`, and `mcp.tool_call` (terminal chat).
- **Config:** `OLMLX_TRACING` (default off) + optional `[otel]` extra. All endpoint/protocol/sampler/service-name config via native `OTEL_*` env vars.
- **Docs:** CLAUDE.md design entry + USER_MANUAL "Distributed Tracing" section (install, env vars, span inventory, Jaeger quick-start).

### Notes / deviations from the plan
- AC #2 (zero-import off path) is guarded by a **subprocess** test; the in-process Task-2 assertion was changed to "imports no *new* otel modules" because installing `opentelemetry-api` activates an unrelated library's optional OTel integration at pytest-collection time (it does not reflect olmlx's off path).
- `RootSpanMiddleware` derives `surface` from the request path rather than `surface_var` — `BaseHTTPMiddleware` copies context into an inner sub-task, so a ContextVar set by an inner middleware isn't visible to the outer root span after `call_next`.
- Streaming `inference` span is held open across the generator via `_trace_inference_gen`, gated on `tracing.enabled()` so the off-path generator topology is unchanged.

### Follow-ups (deferred, low-risk scoping)
- `spec.step`/`spec.prefill` on `PromptLookupDecoder` and the experimental `dflash`/`eagle`/`self` decoders (each a bespoke class — no shared base; wrapping their linear `step` bodies means indenting hot decode loops). Classic `SpeculativeDecoder` is fully instrumented; PLD has `spec.verify`.
- Optional per-turn `chat.turn` span; distributed-worker trace propagation.

## Test Plan
- [x] `uv run pytest tests/test_tracing.py tests/test_tracing_integration.py` — 16 passed, no skips (otel deps in `dev` group)
- [x] AC #2 subprocess guard: `test_off_path_never_imports_opentelemetry`
- [x] Regression slice (app/middleware/streaming/inference/speculative/prompt_cache/chat/config/metrics/flash/kv_disk): 1865 passed, 1 skipped
- [x] `ruff check .` + `ruff format --check .` clean
- [ ] Manual: `OLMLX_TRACING=true OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 uv run olmlx` against a Jaeger all-in-one, confirm a request renders the span tree


olmlx/engine/inference.py
```diff
@@ -3473,50 +3513,64 @@ async def _stream_completion(
...
+                    "repetition_penalty",
+                    "seed",
+                }
+                # Greedy-compatible defaults: temperature=0, top_p=1.0, top_k=0
+                _greedy_defaults = {"temperature": 0, "top_p": 1.0, "top_k": 0}
+                _dropped = {
+                    k
+                    for k in gen_kwargs
+                    if k in _sampling_keys
+                    and gen_kwargs[k] is not None
```
dpalmqvist, line 3540: Verified — the prefill span wraps only synchronous stream construction; the actual prefill forward runs lazily in the `CancellableStream` worker thread (streaming) or fused inside the single `_generate_sync` `to_thread` call (non-streaming), so a coroutine-side span boundary can't enclose it without restructuring the hot loop. Fixed per your second suggestion (fold into decode): `ttft_ns` (= `stats.prompt_eval_duration`) and `cache_hit` are now set on the `decode` span, where the measured prefill duration is available; the `prefill` span remains a setup-phase marker carrying `prompt_tokens`. Streaming `cache_hit` is derived from mlx re-prefilling fewer tokens than the full prompt (prefix served from cache). eebc86d.

olmlx/utils/tracing.py
```diff
@@ -0,0 +1,249 @@
...
+
+    _PROVIDER = provider
+    _TRACER = provider.get_tracer("olmlx")
+    _ENABLED = True
+    _LOG_FILTER = _install_log_filter()  # Task 5
+    logger.info("OpenTelemetry tracing enabled (service.name=%s)", service_name)
+
+
+def install_test_provider(provider: Any) -> None:
+    """Test hook: enable tracing against a caller-supplied TracerProvider
```
dpalmqvist, line 179: Verified — `install_test_provider` overwrote `_LOG_FILTER` without removing the prior filter, so a second call without shutdown would leave duplicate stamps and `shutdown_tracing` would only remove the latest. Fixed: it now uninstalls any existing filter before installing, making repeat calls idempotent. Added `test_install_test_provider_twice_no_duplicate_log_filter` to guard it. eebc86d.

Issue #421, opened 2026-06-03
Cross-request KV cache reuse for speculative decoding (decoder-owned persistent caches)


dpalmqvist:
## Summary

Reuse KV cache across requests for speculative decoding so each agent turn prefills only the new suffix, instead of re-prefilling the entire growing conversation every turn.

This is **Part 2** of the work in `docs/superpowers/specs/2026-06-03-speculative-prefill-cancellation-and-cache-reuse-design.md`. Part 1 (cancellable speculative prefill — fixes the GPU-pinned freeze / deferred-cleanup 503) is done; this issue tracks the larger perf change.

## Motivation

Prompt cache is gated off whenever speculative is on (`inference.py:4328`, `not lm.is_speculative`): speculative decoders own their `_target_cache`/`_draft_cache`, rebuilt fresh every `prefill()` and never persisted. For a long agentic session (Claude-Code-style) against `unsloth/Qwen3.6-27B-MLX-8bit`, every turn re-prefills ~69k tokens.

Observed in a real run (`/tmp/olmlx.log`):

| prompt toks | generated | wall | decode tok/s |
|------------:|----------:|-----:|-------------:|
| 69,079 | 155 | 21 min | 0.2 |
| 69,306 | 95 | 21 min | 0.1 |
| 69,491 | 3,846 | 33 min | 3.2 |
| 48,020 | 142 | 9 min | 0.5 |

Outputs are tiny; prefill is ~99% of wall-clock and is repaid in full each turn. Speculative's ~1.3–1.9× *decode* speedup is irrelevant when prefill dominates. With cross-request reuse, each turn would prefill only the new message + tool results.

## Approach (B — decoder-owned persistent caches)

Chosen over (A) full shared-store integration and (C) target-only reuse.

- A small LRU `_SpecCacheStore` on the `SpeculativeDecoder` instance holding `{tokens, target_snap, draft_snap}` entries. Single decoder instance per model, serialized under the inference lock → no locking needed.
- **Lookup = longest token-prefix match** (linear scan; no `cache_id` dependency, robust to unstable client ids).
- Rewritten `prefill(prompt, *, segmented, cancel_event)`:
  - **Trimmable target** (`KVCache`): deepcopy snapshots into working caches, `trim_prompt_cache` both to the common prefix.
  - **Non-trimmable** (Qwen3.6 / `ArraysCache`): reuse only on a *full* prefix match (strict extension — the chat-append workload); else fresh prefill.
[...]

PR #422, opened 2026-06-03
fix(speculative): make prefill cancellable to stop GPU-pinned freeze

dpalmqvist:
## Problem

A long agentic run against `unsloth/Qwen3.6-27B-MLX-8bit` with classic speculative decoding appeared to "freeze" while pinning the GPU at ~90%, memory flat. From `/tmp/olmlx.log`:

```
17:01:41  request starts → prefilling ~69k tokens (speculative, prompt cache disabled)
  ...      client disconnects mid-prefill (cancel_event set; 10s join expires)
17:08:12  WARNING: "Inference thread still alive after cleanup attempts — deferring..."
17:08:15  next request arrives → "Waiting for deferred GPU cleanup"
17:08:45  next request → 503: "deferred GPU cleanup did not complete within 30.0s"
17:09:51  original thread finally exits (~8 min of non-cancellable prefill)
```

**Root cause:** `decoder.prefill()` ran to completion ignoring `cancel_event`. The decode loop checks `cancel_event` between steps (`speculative_stream.py`), but the prefill helpers `_chunked_prefill` / `_prefill_last_logit` had no checkpoint — so a disconnect during a multi-minute prefill kept the GPU and inference lock busy until prefill finished, 503-ing the next request via deferred cleanup.

## Fix

- New `PrefillCancelled` exception in `speculative.py`.
- `_chunked_prefill` / `_prefill_last_logit` check `cancel_event` at each `_PREFILL_CHUNK` (2048) boundary and raise it.
- `SpeculativeDecoder.prefill` and `PromptLookupDecoder.prefill` thread it through; `dflash`/`eagle` accept the kwarg but prefill in a single forward (documented no-op — experimental, off the default path).
- `speculative_stream_generate` passes the `cancel_event` it already receives from `CancellableStream` and catches `PrefillCancelled` for a clean stream exit (no token yielded). The disconnect → `drain_and_join()` → `cancel()` wiring already existed, so no `inference.py` change was needed.

**Result:** post-disconnect GPU work drops from the whole prefill to at most one sub-chunk (~sub-second–1s); the lock releases promptly and the next request proceeds — no deferred-cleanup 503.

## Tests (TDD)

New `tests/test_speculative_prefill_cancel.py` (written first, watched fail on `ImportError: PrefillCancelled`, then pass):

- `_chunked_prefill` raises when cancelled up front (no forward runs) and stops within one chunk when cancel fires mid-prefill.
- No-cancel path unchanged (backward compat).
- `_prefill_last_logit` and `SpeculativeDecoder.prefill` raise on cancel.
- `speculative_stream_generate` returns cleanly (no tokens) when prefill is cancelled.

Full speculative suite: **193 passed** (updated one existing stub decoder to match the new protocol signature). `ruff check` + `format` clean.

## Not in this PR

Cross-request KV cache reuse under speculative — the underlying reason every agent turn re-prefills ~69k tokens (prompt cache is gated off when speculative is on). Tracked in #421; design doc added under `docs/superpowers/specs/2026-06-03-speculative-prefill-cancellation-and-cache-reuse-design.md`.


dpalmqvist:
@claude review this pr

dpalmqvist:
@claude check the updates and fixes added since your last review

PR #423, opened 2026-06-03
fix(speculative): add cancel_event to SelfSpeculativeDecoder.prefill

dpalmqvist:
## Summary

`SpeculativeDecoderProtocol` declares `prefill(prompt, cancel_event=...)` and `speculative_stream_generate` calls **every** decoder with the kwarg. Part 1 (#422) added it to the classic/flash/dflash/eagle/pld decoders but **missed `SelfSpeculativeDecoder`** — so any `self_speculative` request reaching the stream raised:

```
TypeError: prefill() got an unexpected keyword argument 'cancel_event'
```

Discovered while reviewing #421 (Part 2 cache-reuse design) and scoping `self_speculative` out of cache reuse.

## Fix

- Add `cancel_event: threading.Event | None = None` to `SelfSpeculativeDecoder.prefill`.
- Thread it into the module-local `_prefill_last_logit`, honoring it at the two forward boundaries (before prefix, before last token) so a disconnected request skips the second large forward.
- Granularity is coarser than the classic sub-chunk loop because self-speculative prefills in a single forward — consistent with the single-forward shape of dflash/eagle.

## Testing (TDD)

New `tests/test_self_speculative_prefill_cancel.py`:
- Reproducer: the exact `speculative_stream_generate` call (`prefill(prompt, cancel_event=...)`) — failed with the precise `TypeError` before the fix.
- Backward-compat: positional-only call unchanged.
- Cancellation honored: an already-set event raises `PrefillCancelled` before the prefix forward.

Local run: 3 new pass; 11 pass with the existing `test_speculative_prefill_cancel.py`; `ruff check` + `format` clean.

Related: #422 (Part 1), #421 (Part 2 design).


dpalmqvist:
@claude review the changes in this PR

PR #424, opened 2026-06-03
feat(speculative): cross-request KV cache reuse (#421)

### LINKED ISSUE #421: Cross-request KV cache reuse for speculative decoding (decoder-owned persistent caches)

dpalmqvist:
## Summary

Closes #421 — **Part 2** of the speculative-prefill design (Part 1, cancellation, landed in #422/#423).

Speculative decoding gated prompt caching off (`inference.py`, `not lm.is_speculative`), so every agent turn re-prefilled the entire growing conversation. For a Claude-Code-style session against `unsloth/Qwen3.6-27B-MLX-8bit`, prefill of ~69k tokens was ~99% of wall-clock and repaid in full each turn — dwarfing speculative's ~1.3–1.9× *decode* speedup.

Each `classic`/`pld` decoder now owns a tiny LRU `_SpecCacheStore` of `{tokens → KV snapshot}` keyed by **longest token-prefix match** (linear scan; no `cache_id`). The speculative branch in `inference.py` builds a full-token-aligned `SegmentedPrompt` and threads it through `async_speculative_stream` → `speculative_stream_generate` → `decoder.prefill`. `prefill` deep-copies a matching snapshot into the working caches and drives only the uncovered suffix, snapshotting the deepest interior message boundary for the next turn.

## Correctness landmines (from the design)

- **Single stream.** Prefill **and** decode both run on the **default stream** — `_drive_spec_prefill` *replicates* (not calls) `inference.py`'s `_drive_segmented_prefill`, which forces `generation_stream`. A split reintroduces the #284/#396 GatedDeltaNet / MoE-routing corruption on Qwen3-Next-family targets.
- **Snapshot hygiene.** Snapshots are deepcopied + eager-eval'd (`ArraysCache`) so they survive the cross-worker-thread reuse; `step()`'s in-place mutations never touch the store (deepcopy-on-reuse).
- **Reuse decision** (`_spec_reuse_decision`): **trimmable** (`KVCache`) trims to the common prefix (backs up one on exact match for the seeding logit); **non-trimmable** (hybrid `ArraysCache`) reuses only on a strict full-prefix extension, else fresh prefill.
- Pure-`RotatingKVCache` targets (gpt-oss / Step-3.5) skip reuse (legacy fresh prefill) since an interior split corrupts windowed attention.

## Config

- `OLMLX_SPECULATIVE_CACHE_SLOTS` (default **2**; `0` = kill switch). `classic` + `pld` only; `dflash`/`eagle`/`self`/flash-speculative always fresh-prefill.

## v1 limitations

- In-memory only (no disk spill; not counted against `OLMLX_MEMORY_LIMIT_FRACTION` — 2 slots bound it; documented because the target is a 27B).
- ~30 lines of chunking/snapshot logic intentionally duplicated from `_drive_segmented_prefill` (the call sites diverge — two models, last-logit capture, default vs generation stream).

## Tests (TDD)

New `tests/test_speculative_cache_reuse.py` (30 tests): `_SpecCacheStore` longest-prefix/LRU/`0`-disables; `_spec_reuse_decision` trimmable vs hybrid vs exact-match; end-to-end classic + PLD strict-extension reuse asserting covered-token count **and token-identical output vs fresh prefill** (regression guard against cross-stream/chunking corruption); deepcopy isolation; stream plumbing. Plus config tests for the new knob.

Full speculative + inference + prompt-cache + streaming + router suites green; `ruff check`/`format` clean.


dpalmqvist:
@claude review this PR

dpalmqvist:
Thanks @claude — addressed in 905300b (docs/comments only, no behavior change). Dispositions:

**1. `list.remove(best)` fragility (medium)** — Verified this does *not* manifest, so no change. `list.remove` short-circuits on object identity (`PyObject_RichCompareBool` returns true for `Py_EQ` on identical objects), and `insert` dedups so entry `tokens` are unique — earlier entries fail the `tokens` comparison first and the target is found by identity, so the opaque `payload` is never compared by `==`. Reproduced with a payload whose `__eq__` returns an `mx.array`: `find()` does not raise. Since the suggested change is behavior-identical I can't write a failing test for it, so I left the code as-is and recorded the reasoning in the commit.

**2. Double deepcopy on reuse / untracked memory peak (medium)** — Accurate; it's the unavoidable cost of copy-on-reuse isolation. Documented it: the `OLMLX_SPECULATIVE_CACHE_SLOTS` config docstring now states each slot is a live materialized KV deepcopy and a reuse hit briefly holds an extra working copy on top of the resident slots (uncounted against `OLMLX_MEMORY_LIMIT_FRACTION`), plus inline comments at the reuse sites.

**3. `is_trimmable` on fresh cache, used after replace (low)** — Added a one-line comment in both decoders: cache layer types are invariant per model, so classifying the fresh cache is valid for the restored snapshot.

**4. LRU promotion on rejected reuse (low)** — Documented in `_SpecCacheStore.find`'s docstring: promotion happens on any prefix overlap, even a later-rejected near-miss; immaterial at the default slot counts, revisit if slots are raised substantially.

On the two automated (aider) findings you concurred with — the string-based cache-type classification and the cancellation-check duplication — I'm leaving both as-is intentionally: the name-allowlist matches the established convention across `model_manager.py`/`inference.py` (a new cache type degrades safely to fresh prefill, never corrupts), and the driver duplication is a deliberate design choice (the two call sites diverge on models/last-logit/stream — noted in the design doc's "known limitations"). Both are reasonable follow-ups if a marker-based classification or a shared driver is wanted later.

Issue #429, opened 2026-06-04
Lift prompt-cache / grammar / speculative gating on the VLM (image) path


dpalmqvist:
## Why

Three engine features are gated off whenever a request carries images, which matters more now that multimodal-first models (Gemma 4) are first-class:

1. **Prompt caching** — non-streaming VLM is gated off because `mlx_vlm.generate` doesn't accept `prompt_cache` (per CLAUDE.md). Multi-turn vision chats re-prefill the (often large) image prefix every turn.
2. **Structured outputs / grammar** — xgrammar `logits_processors` aren't threaded through the VLM path, so JSON-mode / JSON-Schema is silently skipped with images.
3. **Speculative decoding** — skipped with images (`speculative decoding skipped: request includes images`).

Each is an independent improvement; filing as one tracking issue with three sub-tasks since they share the VLM-path plumbing.

## Scope

- [ ] **Prompt cache on VLM**: thread a `prompt_cache` through the VLM forward (either via an mlx_vlm API that accepts it, or by driving the language-model submodule directly post-vision-embed) so cross-turn image-prefix reuse works. Respect the existing radix/`cache_id` machinery.
- [ ] **Grammar on VLM**: thread `logits_processors` into the VLM decode loop so `response_format` / `format` apply with images present.
- [ ] **Speculative on VLM**: scope feasibility — most drafts are text-only, so this may be wontfix or limited to PLD over the text suffix. Document the decision.

## Out of scope

- Distributed VLM.
- Video/audio modalities (tracked separately) — though the prompt-cache fix should generalize.

## Acceptance criteria

- A 3-turn vision conversation reuses the image-prefix KV across turns (verify via `/api/ps` cache metrics), with output identical to the uncached path.
- `response_format={"type":"json_schema",...}` produces schema-valid JSON on an image request.
- The speculative-with-images decision is documented in CLAUDE.md (enabled, scoped, or explicitly out of scope).

## Files

- `engine/inference.py` (VLM branch gating: prompt-cache, grammar, speculative skips)
- `engine/model_manager.py` (VLM generate path)
- `engine/prompt_cache/*`, `engine/grammar.py`
- CLAUDE.md (update the VLM-gating design note)

dpalmqvist:
Done in #437 (merged as 40c99d8).

All three sub-tasks + acceptance criteria:
[...]

PR #437, opened 2026-06-07
feat(vlm): lift prompt-cache / grammar / speculative gating on the image path (#429)

dpalmqvist:
## Summary

Lifts the three engine features that were gated off whenever a request carried images, closing #429. The key discovery: **mlx_vlm 0.4.4 natively supports `prompt_cache`, `logits_processors`, and a `PromptCacheState` cross-turn reuse mechanism** — the olmlx gating notes were stale, so most of this is removing dead gates and wiring mlx_vlm's own machinery.

- **Grammar on VLM** — lift the `is_vlm` early-return in `_install_grammar_processor`; `_resolve_model_vocab_size` now descends into `model.language_model` to size the xgrammar bitmask. `logits_processors` already flows through `gen_kwargs` to `mlx_vlm.generate_step`.
- **Prompt caching on VLM** — new `VlmPromptCacheStore`, a tiny per-`cache_id` LRU of mlx_vlm `PromptCacheState` objects (modeled on the speculative `_SpecCacheStore`), created per VLM `LoadedModel`. mlx_vlm owns prefix matching / cache trimming / image-in-prefix detection and updates the state post-generation; `_setup_vlm_prompt_cache` only fetch-or-creates it. Works on **both** streaming and non-streaming. Reuse counters (`vlm_cache_hits`/`misses`/`tokens_reused`) surface on `/api/ps`. Tunable via `OLMLX_VLM_PROMPT_CACHE_SLOTS` (default 2, 0=off).
- **Non-streaming VLM** now drains `mlx_vlm.stream_generate` instead of `mlx_vlm.generate`, so `prompt_cache_state` + `logits_processors` thread through and token counts are real (fixes the long-standing `prompt_eval_count`/`eval_count == 0` gap).
- **Removed the image-dropping `input_ids` hack** — the old text-path VLM branch passed text-only token ids to mlx_vlm, which treated them as authoritative and silently dropped the image on fresh requests.
- **Speculative + VLM** documented as out of scope (text-only drafts; target multimodal forward not threaded).

**v1 limits (documented in CLAUDE.md):** VLM cache is in-memory only — no radix-takeover, disk spill, or KV-quant; distributed VLM out of scope.

## Test Plan

- [x] Unit: `tests/test_vlm_prompt_cache_store.py` (LRU semantics + metrics), `tests/test_grammar_vlm_vocab.py` (VLM vocab resolution)
- [x] Regression: text path unchanged — 1099 passed across grammar/prompt_cache/inference/completion/status/model_manager suites; ruff clean
- [ ] **Live (`real_model`, requires Gemma 4 VLM downloaded):** `tests/live/test_vlm_cache_grammar.py` — run `uv run pytest tests/live/ -v -m real_model`. Covers all #429 acceptance criteria: grammar json_schema on an image request, 3-turn image-prefix KV reuse (`/api/ps`), cached==uncached greedy output, non-zero non-streaming token counts.


dpalmqvist:
**Live `real_model` suite run locally — all green (6/6):**

```
tests/live/test_vlm_cache_grammar.py::test_grammar_json_schema_on_image_request PASSED
tests/live/test_vlm_cache_grammar.py::test_non_streaming_vlm_reports_token_counts PASSED
tests/live/test_vlm_cache_grammar.py::test_multi_turn_vision_reuses_image_prefix_kv PASSED
tests/live/test_vlm_cache_grammar.py::test_cached_and_uncached_outputs_match PASSED
tests/live/test_vlm_tools_images.py::test_openai_vlm_tools_with_image_produces_tool_call PASSED
tests/live/test_vlm_tools_images.py::test_ollama_vlm_tools_with_image_produces_tool_call PASSED
```

The first run surfaced a real integration bug (fixed in `f637d8e`): the VLM cache branches set `full_prompt_tokens` to the **text-tokenizer** count, which misses image-patch expansion (~19 vs ~256), and both completion paths override `stats.prompt_eval_count` with `len(full_prompt_tokens)` — undercounting `prompt_tokens` for image requests. The text-path override is correct (mlx-lm under-reports on cache hits), but mlx_vlm's `GenerationResult` already reports the accurate full size (reused + new) on hits and misses, so the VLM branches now leave `full_prompt_tokens=None` and let the engine count stand. The same bug also silently affected the existing `test_vlm_tools_images.py` tools tests (which assert `prompt_tokens > 200`); both pass after the fix.

olmlx/engine/inference.py
```diff
@@ -3486,7 +3535,23 @@ async def _stream_completion(
...
+            # VLM: attach an mlx_vlm PromptCacheState. ``prompt`` stays the full
+            # str — mlx_vlm tokenizes it and reuses the KV prefix internally.
+            read, creation = _setup_vlm_prompt_cache(
+                lm, prompt_tokens, gen_kwargs, cache_id=cache_id
+            )
+            cs = _CacheSetupResult(
+                prompt=prompt,
+                cache_read_tokens=read,
+                cache_creation_tokens=creation,
+                # None: the text-tokenizer count misses image-patch expansion;
```
dpalmqvist, line 3548: Thanks — I looked into this closely and don't think it's a correctness bug; declining the suggested removal.

`mlx_vlm` only calls `prompt_cache_state.update(all_ids, tracked_cache)` on **natural completion** (generate.py:784, after the decode loop). So an interrupted generation (client disconnect / error) leaves `state.token_ids` at the last *successful* value, while `state.cache` may hold extra appended KV beyond it.

On the next reuse, `mlx_vlm` self-heals: it computes `prefix_len = find_prefix_length(new_ids)` against that valid `token_ids`, then explicitly trims the cache — *"Trim cache to prefix_len in case it includes generated tokens"* (generate.py:688) → `if cached_len > prefix_len: c.keys = c.keys[:, :, :prefix_len, :]`. Positions `[0:prefix_len]` are the shared prefix and are never rewritten (KVCache appends only), so the partial KV is discarded and output stays exact. The 3-turn `test_cached_and_uncached_outputs_match` live test exercises this reuse+trim path and produces byte-identical output to the uncached run.

Two more points:
- `VlmPromptCacheStore` has no `remove` method (`get`/`insert`/`clear` only), so the suggested call wouldn't compile as written.
- Removing the entry on every incomplete generation would *defeat* cross-turn reuse for exactly the agentic-cancellation case (clients that cancel mid-stream, cf. #422) — strictly worse latency with no correctness gain, since the trim already makes reuse safe.

Transient memory: the partial cache is bounded (2-slot LRU, re-trimmed on the next reuse of that `cache_id`), so no unbounded growth.

dpalmqvist:
**`/code-review` (high effort, 4 finder angles + verify) — 1 fix, rest refuted.**

✅ **Fixed (`97c8d75`):** the under-pressure pre-load memory-hygiene flush in `ModelManager.ensure_loaded` evicted each resident model's *text* `prompt_cache_store` but never cleared `vlm_prompt_cache_store`, leaving image-prefix KV (live mlx arrays) resident — partially defeating the #223 thrash-avoidance flush when loading a new model on top of VLM turns. Now cleared in the same loop.

Refuted after verifying against the code / passing live tests:
- **Stale `PromptCacheState` after aborted generation → corruption.** mlx_vlm only calls `state.update()` on natural completion, and on the next reuse it trims the cache to `find_prefix_length` against the last-good `token_ids` (*"Trim cache to prefix_len in case it includes generated tokens"*). Self-healing; positions `[0:prefix_len]` are never rewritten. (Same as the aider thread.)
- **`mx.clear_cache()` frees live cache arrays → use-after-free.** `clear_cache()` only releases the allocator's unused buffer pool, not refcounted live `mx.array`s held by `PromptCacheState.cache`.
- **Cache-hit underreports `prompt_tokens`.** mlx_vlm sets `prompt_tokens = reused_prefix_len + input_ids.size` = full size on hits and misses; the `full_prompt_tokens=None` change is what lets that accurate count stand (verified by `test_non_streaming_vlm_reports_token_counts`).
- **Grammar masks wrong vocab → invalid JSON on VLM.** `test_grammar_json_schema_on_image_request` produces schema-valid JSON; vocab is sized via the model's `lm_head` and the grammar uses the same text tokenizer mlx_vlm uses.
- **LRU eviction → #284 cross-thread use-after-free.** Eviction runs under the inference lock during the next request's setup, after the prior request's `mx.synchronize()`; the arrays are quiescent.

Known minor (not fixed): the VLM `cache_read`/`cache_creation` counts on `/api/ps` are a text-tokenizer estimate that undercounts image-patch reuse — documented as approximate in `_setup_vlm_prompt_cache` (mlx_vlm doesn't expose its internal `reused_prefix_len`); cosmetic on this single-user local server.

PR #439, opened 2026-06-08
fix(bench): kill worker process group on timeout to stop orphaned-server RAM leak

dpalmqvist:
## Problem

Running `olmlx bench run --prompt-set all` across a model ladder drove olmlx RAM to **~61 GB** on a 64 GB machine, with a cascade of scenario failures (by the 8B model, every scenario failed). Throughput on small models was fine, so this looked like a leak under load.

## Root cause (bench harness, not the engine)

`olmlx/bench/runner.py::_run_worker` spawned each scenario via `subprocess.run(..., timeout=worker_timeout)`. The worker itself `Popen`s a uvicorn **server** child. On `worker_timeout`, CPython **SIGKILLs the worker** — uncatchable, so the worker's `finally` block (which would terminate its server) never runs. The **server orphans with the model still resident in RAM**.

57-prompt quality scenarios on ≥4B models exceed the old 600s timeout, so each timeout leaked a server. Orphans are detached, so they accumulate across scenarios *and* across models in the ladder → tens of GB → later loads starve and fail.

Ruled out before fixing:
- **Per-request leak / recent VLM changes** — refuted: a single text server is flat at **0.92 GB across 40 requests**; the VLM prompt-cache/grammar code is gated on `is_vlm` and never touches the text path.

## Fix

- Launch the worker with `start_new_session=True` and reap the **whole process group** (`os.killpg`) in a `finally` on every path (timeout / crash / clean exit), so the server can never outlive the worker. Switched from `subprocess.run` to `Popen` + `communicate(timeout=)` to own the kill.
- Raise the default `OLMLX_BENCH_WORKER_TIMEOUT` **600s → 1800s** — the 50-prompt quality set on mid/large models legitimately runs many minutes per scenario and was reliably timing out (and losing its results) at 600s.

## Verification

- A worker forced to time out mid-generation now leaves **zero** orphaned servers (was: orphan holding the model).
- A 4B `--prompt-set all` stress run with a forced **90s** timeout (deliberately recreating the failure condition) held at **max 1 concurrent server / ~4 GB peak** across many forced timeouts — previously each timeout added ~2.5 GB of permanent orphan.

## Tests

`tests/test_bench_runner.py`: process group is SIGKILLed on timeout, worker runs in its own session, group is reaped on clean exit. `tests/test_bench_quality_wiring.py`: raised-default + override assertions. All bench tests green (`57 passed`).

## Mid-tier quality runs (real large-model confirmation)

Re-ran the mid-tier quality suite (baseline scenario, raised timeout) on this branch. All completed **50/50 prompts, no timeouts, no failures**, with memory bounded the whole time:

| Model | Quality | Throughput | Notes |
|---|---|---|---|
| gemma-4-26b-a4b-it-4bit | 38/40 (95%) | 69.9 tok/s | MoE 4B active |
| Qwen3.5-27B-4bit (dense) | 38/40 (95%) | 15.6 tok/s | dense — slowest decode |
| Qwen3.6-35B-A3B-6bit | 35/40 (88%) | 57.7 tok/s | MoE 3B active, `turboquant:4` KV |

(40 = graded GSM8K+MMLU+HumanEval subset; 10 `code_exec` prompts excluded without `--enable-code-exec`.)

Memory during these three runs: **peak olmlx RSS 27.4 GB** (the 6bit 35B + tq4), **max 1 concurrent server**, **0 orphans** — bounded at a single model's footprint vs. the previous 61 GB cascade.

### DIFF
--- a/olmlx/bench/runner.py
+++ b/olmlx/bench/runner.py
@@ -6,6 +6,7 @@ import json
 import logging
 import math
 import os
+import signal
 import socket
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


dpalmqvist:
Re: the aider-code-review note on `Popen` raising → `UnboundLocalError`: this is a false positive. `proc = subprocess.Popen(...)` (runner.py:471) is **outside and before** the `try:` (line 479) whose `finally` calls `_terminate_process_group(proc)`. If `Popen` raises, the exception propagates out of `_run_worker` before the `try` is ever entered, so the `finally` does not run and `proc` is never referenced unbound. (The `proc = None` guard in `_run_server_scenario` is needed there only because that function's `Popen` sits *inside* its `try`.) No change needed.

PR #443, opened 2026-06-08
feat(audio): OpenAI-compatible /v1/audio/speech TTS endpoint (#367)

### LINKED ISSUE #367: TTS endpoint (/v1/audio/speech)
## Why

Pairs with the Whisper transcription endpoint to give olmlx full OpenAI audio parity. mlx Kokoro and Parakeet TTS models are available and run well on Apple Silicon.

## Scope

- `POST /v1/audio/speech` in the shared audio router
- Voice selection
- Streaming PCM/MP3
- TTS model class in the model registry (loadable via `olmlx models pull`)

## Out of scope

- Voice cloning UI
- Custom voice training

## Acceptance criteria

- OpenAI SDK `client.audio.speech.create()` works without changes
- Both Kokoro and Parakeet load and produce audible output
- Streaming mode delivers first chunk in < 1s on M3 for short utterances

## Files

- `routers/audio.py` (shared with the Whisper endpoint)
- engine TTS path under `engine/`

## Depends on

The Whisper transcription endpoint (router scaffolding lives there).

dpalmqvist:
## Summary

Adds an OpenAI-compatible text-to-speech endpoint, `POST /v1/audio/speech`, backed by mlx-audio (Kokoro-82M). Closes #367. It mirrors the existing Whisper transcription path: TTS is a first-class `ModelManager` kind, an engine async-generator (`generate_speech`) drives mlx-audio under the serialized inference lock, and the route lives in the shared `routers/audio.py`.

- **Model kind** — `_detect_model_kind` recognizes Kokoro by its config signature (`istftnet` + `plbert`; Kokoro ships no `model_type`). `LoadedModel.is_tts` guards the LLM-only prompt-cache / KV-quant / grammar paths, mirroring `is_whisper`. Loaded via `mlx_audio.tts.utils.load_model`.
- **Engine** — `generate_speech` runs mlx-audio's *synchronous* `model.generate` in one worker thread (all MLX work on a single thread, per the #284 stream hazards) and bridges float-audio segments (24 kHz mono) to the event loop via a queue.
- **Voices** — the six OpenAI names map to Kokoro ids (`alloy→af_alloy`, `echo→am_echo`, `fable→bm_fable`, `onyx→am_onyx`, `nova→af_nova`, `shimmer→af_sky`); native Kokoro ids pass through; unknown → 422.
- **Formats** — all six: `pcm` raw-streamed, `wav` pure-Python buffered (exact RIFF header), `mp3`/`opus`/`aac`/`flac` piped through ffmpeg (already required) with guaranteed subprocess reaping.
- **Streaming + non-streaming** — `StreamingResponse` serves both SDK call styles. The router primes the **PCM source** generator so a non-TTS model surfaces as a clean 400 before the 200 stream starts (priming the ffmpeg-encoded stream wouldn't catch it — ffmpeg emits its header before reading stdin).
- **Validation** — empty input → 422, oversize (`OLMLX_TTS_MAX_INPUT_CHARS`, default 8192) → 413, bad format → 422, unknown voice → 422, non-TTS model → 400.
- **Deps** — `mlx-audio` + `misaki[en]` (Kokoro's English G2P; auto-downloads the spaCy `en_core_web_sm` model on first generation).

**Out of scope:** voice cloning/training, the SSE `stream_format` event API, distributed TTS, KV-quant/prompt-cache for TTS, non-Kokoro TTS models (path is generic but only Kokoro is wired/tested in v1). The issue listed "Parakeet" but that's an STT model, not TTS — dropped.

## Test Plan

- [x] Full non-live suite: **4329 passed, 3 skipped** (`pytest -m "not real_model"`)
- [x] Unit coverage: voice resolution, audio encoding (pcm/wav + real ffmpeg mp3), `generate_speech` (segments, non-TTS reject, early-disconnect teardown), router (all formats, 422/413/400 paths), model-kind detection, TTS load branch + guards
- [x] **Live test against real Kokoro** via the OpenAI SDK over in-process ASGI: all 6 formats + streaming first-chunk — **7 passed** (`tests/live/test_tts_speech.py`, `real_model`)
- [x] `ruff check` + `ruff format --check` clean across the tree
- [ ] Reviewer: confirm `misaki[en]` resolves cleanly in a fresh `uv sync` on a clean machine


dpalmqvist:
@claude review this pr please

dpalmqvist:
Thanks for the review — dispositions below.

**Fixed**
- **wav `pcm += chunk` (O(n²))** — fixed in 3bdb652: collect chunks in a list and `b"".join()` once.

**Evaluated, not changing (with reasoning)**
- **Worker catches "BaseException-derived" failures / `isinstance(item, Exception)` (high+medium)** — the worker uses `except Exception`, which does *not* catch `BaseException` (KeyboardInterrupt/SystemExit propagate); and `asyncio.to_thread` doesn't inject `CancelledError` into the thread. Queued items are always `np.ndarray` segments (never exception instances) and the sentinel is a private `object()`, so the identity + `isinstance` checks can't collide. No reachable hang here.
- **Streaming priming only covers the first chunk (medium)** — this is inherent to any streaming response: once the 200 + first bytes are sent, a later generation error can only fail mid-stream. The priming exists specifically to convert the *pre-stream* failure (a non-TTS model, which raises on the first pull before any bytes) into a clean 400; that path is covered by `test_speech_non_tts_model_400`. A mid-generation failure after bytes have flowed isn't expressible as an HTTP status in any streaming design.
- **istftnet+plbert detection is load-bearing (low)** — agreed it's a contract; it's already documented at the detection site and in CLAUDE.md ("Text-to-speech" bullet), mirroring how the Whisper kind is keyed off its dims. Kept as-is.
- **Live async fixture (medium)** — `@pytest.fixture` + `async def sdk_client` is byte-identical to the existing `tests/live/test_responses_sdk.py` convention (the repo runs `asyncio_mode = auto`), and the live suite passes 7/7 against real Kokoro. Matching the established pattern intentionally.

dpalmqvist:
Dispositions on the review (fixes in d8991fc):

**Fixed**
- **#2 unbounded queue** — `generate_speech` now caps in-flight segments with a `threading.Semaphore(4)`: the worker blocks on a free slot before producing, the consumer releases one per segment taken, and teardown releases once to unblock a worker parked in `acquire()`. Verified by `test_generate_speech_releases_ref_on_early_close` (100-segment producer, early close) and a real-Kokoro live run (7/7, no deadlock).
- **#3 orphaned feeder exception** — `ffmpeg_encode`'s `finally` now always awaits `feeder` (cancelling only if still running) and swallows `CancelledError`/`BrokenPipeError`/`ConnectionResetError`, so the broken-pipe case no longer logs "Task exception was never retrieved".
- **#4 missing 413 test** — added `test_speech_oversize_413`.

**Evaluated, not changing**
- **#1 `except ValueError` too narrow / `ModelLoadTimeoutError` → 500** — there's a global `@app.exception_handler(ModelLoadTimeoutError)` → 504 at `app.py:318`. `ModelLoadTimeoutError` is a `TimeoutError`, not a `ValueError`, so it propagates past the local `except ValueError` (which only maps the non-TTS case to 400) to that global handler → 504. This matches how `/v1/audio/transcriptions` relies on the same global handler — it doesn't catch the timeout locally either. So the timeout already maps to 504, not 500.
- **#5 hardcoded `_TTS_SAMPLE_RATE`** — correct for Kokoro (the v1/only wired TTS model) and documented as a Kokoro assumption in CLAUDE.md. Threading `result.sample_rate` through the queue is the right move when a second TTS model at a different rate is added; left as a v1 simplification rather than speculative generality.

Also: the fresh aider-review pass on the prior SHA reported "No high-signal issues found."

olmlx/routers/audio.py
```diff
@@ -137,3 +140,89 @@ async def create_transcription(
...
+            async for chunk in _pcm_chunks():
+                parts.append(chunk)
+        except ValueError as exc:  # non-TTS model etc.
+            raise HTTPException(status_code=400, detail=str(exc)) from exc
+        return Response(
+            audio_utils.wav_bytes(b"".join(parts), sample_rate=_TTS_SAMPLE_RATE),
+            media_type=media_type,
+        )
+
+    # Prime the first PCM chunk so model-load / non-TTS errors (ValueError)
```
dpalmqvist, line 216: Fixed in f203a86: pre-check `shutil.which("ffmpeg")` for compressed formats and return a clean 500 with an install hint before streaming starts (ffmpeg is only spawned once StreamingResponse iterates, after the 200). Added test_speech_missing_ffmpeg_500.

olmlx/utils/audio.py
```diff
@@ -0,0 +1,148 @@
...
+            data = await proc.stdout.read(65536)
+            if not data:
+                break
+            yield data
+        # A broken pipe here means ffmpeg died mid-stream while the feeder was
+        # still writing stdin; swallow it so the returncode + stderr diagnostic
+        # below is what surfaces, rather than the less useful BrokenPipeError.
+        try:
+            await feeder
+        except (BrokenPipeError, ConnectionResetError):
```
dpalmqvist, line 132: Fixed in f203a86: the ffmpeg_encode finally now swallows ANY feeder exception (cancellation, broken pipe, or an upstream error re-raised from the propagating body) so proc reaping is never skipped. The real error still propagates from the body.

olmlx/utils/audio.py
```diff
@@ -0,0 +1,151 @@
...
+            err = (await proc.stderr.read()).decode(errors="replace")
+            raise RuntimeError(f"ffmpeg encode failed ({proc.returncode}): {err}")
+    finally:
+        # Always await the feeder (cancel only if still running) so its
+        # result/exception is retrieved — if it already died with a broken
+        # pipe (ffmpeg gone), not awaiting it logs "Task exception was never
+        # retrieved" on generator close.
+        if not feeder.done():
+            feeder.cancel()
+        # Swallow ANY feeder exception here (cancellation, broken pipe, or an
```
dpalmqvist, line 151: Addressed in cc47c24, taking the "always logged" path rather than swallowing: the body's `await feeder` now logs an unexpected feed error with exc_info and re-raises. Swallowing (the other suggested option) would silently truncate the audio with no signal; once the 200 is sent a real mid-generation error can only surface as a broken stream, so propagating it is intentional. The common case (non-TTS model) never reaches here — it's caught during PCM-source priming before ffmpeg starts. proc reaping is already guaranteed by the finally.

olmlx/routers/audio.py
```diff
@@ -137,3 +141,103 @@ async def create_transcription(
...
+        ):
+            yield audio_utils.float_to_pcm16(seg)
+
+    media_type = audio_utils.FORMAT_MEDIA_TYPES[fmt]
+
+    # WAV is buffered so the header sizes are exact (small utterances).
+    if fmt == "wav":
+        # Collect into a list and join once — `pcm += chunk` in the loop is
+        # O(n^2) (each += copies the whole accumulated buffer).
+        parts: list[bytes] = []
```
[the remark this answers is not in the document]
dpalmqvist: The premise here isn't accurate. `ensure_loaded` already resolves per-model `keep_alive` when the request value is `None` (model_manager.py: `effective_keep_alive = keep_alive; if effective_keep_alive is None and model_config.keep_alive is not None: effective_keep_alive = model_config.keep_alive`), then falls back to the global default. So a TTS model's `models.json` keep_alive IS applied. Passing `keep_alive=None` from `_pcm_chunks` is correct: `/v1/audio/speech` (like `/v1/audio/transcriptions`, which also doesn't forward it) is an OpenAI surface with no keep_alive field, so there's nothing client-supplied to forward — and per-model config still wins over the global default via the fallback above. Matches the Whisper path intentionally.

The two summary-only notes are also non-actionable: the module-level `_TTS_QUEUE_MAX_SEGMENTS` constant is conventional placement, and the finally's extra `slots.release()` over-increments a per-call semaphore that's discarded the moment the generator finishes (no reuse, no practical impact) — it exists to unblock a worker parked in `acquire()` during teardown.

PR #457, opened 2026-06-08
fix: address code-review findings #454 (security, error handling, module splits)

### LINKED ISSUE #454: Code review: security, maintainability, and error handling findings
## Code review findings

A comprehensive review of the olmlx codebase across key modules (`engine/inference.py`, `engine/model_manager.py`, `cli.py`, `app.py`, `utils/streaming.py`, `routers/`, `chat/`) identified the following findings, ordered by severity.

---

### 1. [HIGH] Distributed secret exposed via SSH command line

**File:** `olmlx/cli.py:1543`

The distributed secret is injected into the SSH remote command via `shlex.quote()`, which protects against shell injection but leaves the secret visible in `ps aux` on both the coordinator and worker machines for the lifetime of the worker process:

```python
f"printf %s {shlex.quote(settings.distributed_secret)} > $SECRET_FILE",
```

Additionally, the `chmod 600` runs *after* the secret is written, so there is a window where the temp file is world-readable on the remote host.

**Recommendation:** Use SSH `SendEnv` / `AcceptEnv`, a pre-shared keyfile, or `ssh -F` with a dedicated config that sets environment variables server-side, rather than embedding the secret in the command string.

---

### 2. [HIGH] Plist writes all `OLMLX_*` environment variables including secrets

**File:** `olmlx/cli.py:1602-1618`

`_build_plist()` forwards **all** `OLMLX_*` environment variables into the launchd plist:

```python
for key, value in os.environ.items():
    if key.startswith("OLMLX_"):
        env_vars[key] = value
```

This captures `OLMLX_DISTRIBUTED_SECRET` if set. The plist is written to `~/Library/LaunchAgents/com.olmlx.plist`, which is readable by any local process. Plist storage of secrets is not a safe credential store.

**Recommendation:** Filter out known sensitive keys (`*_SECRET`, `*_KEY`, `*_TOKEN`, `*_PASSWORD`) from the plist environment variables, or require secrets to be set via a separate mechanism (keychain, config file with restricted permissions).

---

### 3. [MEDIUM] `inference.py` and `model_manager.py` are dangerously large

- **`olmlx/engine/inference.py`**: 5,277 lines
- **`olmlx/engine/model_manager.py`**: 3,864 lines

These two files together contain ~9,141 lines and encompass: lock management, inference orchestration, preflight checks, prompt caching, tokenization, chat template application, grammar setup, speculative dispatch, model loading/unloading, cache capability probing, Flash/MoE setup, TTS/Whisper loading, reranker setup, and more.

**Risk:** High cognitive load for contributors, difficult to review PRs, higher regression risk from unrelated changes touching the same file.

**Recommendation:** Split into focused modules. For `inference.py`, extract at minimum: prompt caching logic, chat template application, grammar/logits processor setup, and speculative dispatch. For `model_manager.py`, extract: speculative draft loading, Flash/MoE setup, TTS/Whisper/reranker loading into their own modules.

---

### 4. [LOW] Swallowed exception in metrics unregistration during shutdown

**File:** `olmlx/app.py:140-141`

```python
try:
    metrics_mod.REGISTRY.unregister(app.state.metrics_collector)
except Exception:
    pass
```

If the collector is already unregistered or some other failure occurs, the error is silently discarded with no log message. This could mask a real bug during test teardown or repeated `create_app()` calls.

**Recommendation:** Log at debug level before swallowing, or narrow the exception type.

---

### 5. [LOW] `BaseException` in CLI tool paths suppresses `KeyboardInterrupt`

**Files:** `olmlx/chat/builtin_tools.py:269,433`

The bash tool and search tool use `except BaseException:` to kill subprocesses before re-raising. Since these are user-facing CLI code paths (not lock-holding async context managers like the inference code), catching `BaseException` means `KeyboardInterrupt` is intercepted, the subprocess is killed, and the exception propagates — this is **functionally correct** (Ctrl+C still exits) but unnecessarily broad. A stray `SystemExit` from a background thread could also trigger the kill path.

**Recommendation:** Narrow to `except (Exception, KeyboardInterrupt):` or `except (asyncio.CancelledError, Exception):` so the intent is clear and `SystemExit` isnt intercepted.

---

### 6. [INFO] Pervasive broad `except Exception:` patterns

Across the codebase there are 41+ `except Exception:` catch-all sites (13 in `inference.py`, 11 in `model_manager.py`, 9 in `utils/streaming.py`, 6 in `cli.py`, plus dozens more). Most are in cleanup/finally paths where the intent is log-and-continue, which is reasonable. However, a few lack sufficient logging to distinguish between expected errors (e.g., a closed connection) and unexpected bugs (e.g., an `AttributeError` from a typo in a refactored field name), making debugging harder.

Notable low-logging sites:
- `olmlx/utils/streaming.py:447` — queue put failure breaks silently from the worker loop
- `olmlx/engine/model_manager.py:1200` — prompt cache eviction failure in an already-contentious TOCTOU path

**Recommendation:** In non-cleanup contexts, log the exception type and message at `debug` or `warning` level so operators can spot regressions.

---

### 7. [INFO] No test coverage for plist secret filtering or SSH secret handling

**Files:** `olmlx/cli.py:1602-1618` (plist), `olmlx/cli.py:1539-1547` (SSH secret)

The distributed tests (`tests/test_distributed.py`) test `_launch_distributed_workers` but dont assert on the secret transport mechanism. There are no tests for `_build_plist` filtering sensitive environment variables.

**Recommendation:** Add unit tests verifying that `_build_plist` filters out secret-bearing keys and that the SSH command does not include the raw secret.

---

### Overall assessment

The codebase is well-structured for a complex ML inference server. Error handling in streaming/threading paths is thorough with proper `finally` blocks, lock release, and deferred cleanup. The main areas needing attention are the two security findings (items 1 and 2 above) and the long-term maintainability risk from the oversized core modules (item 3).

dpalmqvist:
Addresses the findings in #454. Scope was confirmed with the issue author: the triage "Worth doing" set, the module splits (#3), and the BaseException narrowing (#5). The SSH `SendEnv` rework (#1) was explicitly deferred.

## Security / error handling
- **#2 [HIGH]** — `_build_plist()` no longer forwards credential-bearing `OLMLX_*` env vars into the cleartext launchd plist (`~/Library/LaunchAgents/com.olmlx.plist`). Filters keys ending in `_SECRET`/`_TOKEN`/`_KEY`/`_PASSWORD`/… plus `SECRET`/`PASSWORD`/`CREDENTIAL` substrings. Suffix matching is deliberate so the many `*_TOKENS` config keys (`OLMLX_SPECULATIVE_TOKENS`, `OLMLX_PROMPT_CACHE_MAX_TOKENS`, …) are **not** mistaken for credentials.
- **#7** — added `test_plist_filters_secret_env_vars` covering both the filtered and the preserved keys.
- **#4 [LOW]** — `app.py` shutdown now debug-logs the metrics-collector unregister failure instead of silently swallowing it.
- **#6 [INFO]** — `utils/streaming.py` worker now debug-logs the queue-put failure before breaking the loop.
- **#5** — narrowed the two `except BaseException` cleanup-and-reraise sites in `chat/builtin_tools.py` to `(Exception, KeyboardInterrupt, asyncio.CancelledError)` so a stray `SystemExit` isn't intercepted.

## #3 — module splits (maintainability)
Three move-and-re-export extractions. Each new module holds the moved code; the original re-exports the names so all call sites, `chat/session.py`, and tests are unchanged (only a handful of `mock.patch` targets were repointed to the new module).

| New module | Contents | Lines moved |
|---|---|---|
| `engine/logits_processors.py` | grammar installer, freq/presence penalty processors, model-vocab resolver, gpt-oss channel filter | ~300 |
| `engine/chat_templating.py` | text + VLM `apply_chat_template`, tool injection, thinking resolution, segmented tokenization, tool-message role rewrites | ~714 |
| `engine/speculative_loaders.py` | `SpeculativeLoaderMixin` — draft loaders for classic/dflash/eagle/mtp/pld/self + draft-path resolver + vocab check (mixed into `ModelManager`) | ~700 |

Result: `inference.py` 5277 → 4293 (−18.6%), `model_manager.py` 3864 → 3139 (−18.8%).

The speculative mixin avoids a circular import by never importing `model_manager` at top level; the one shared free helper (`_load_with_model_type_fallback`) is pulled in lazily at call time.

### Deliberately deferred
- **#1** — SSH `SendEnv`/keyfile secret transport (author opted out for this PR).
- The inference.py prompt-cache **orchestration** and model_manager flash/MoE setup were left in place. They're tightly coupled to the inference lock + Metal-stream/deepcopy machinery (the documented #284/#396 hazards), so splitting them carries real regression risk that outweighs the maintainability win here. Better as a focused follow-up.

## Testing
- Full suite: **4449 passed, 4 skipped**.
- `ruff check` + `ruff format --check` clean on all touched files.

Closes #454


olmlx/chat/builtin_tools.py
```diff
@@ -266,7 +266,10 @@ async def _run_search(cmd: list[str]) -> tuple[str, str, int | None]:
                 proc.communicate(),
                 timeout=30,
             )
-        except BaseException:
+        except (Exception, KeyboardInterrupt, asyncio.CancelledError):
```
dpalmqvist, line 269: Intentional, per issue #454 finding #5, which explicitly asked to narrow this `except BaseException` (the issue author classified it as cosmetic). The genuine cleanup triggers — `asyncio.CancelledError` (task cancellation/timeout) and `KeyboardInterrupt` (Ctrl-C) — are both still caught and re-raise after killing the child. The only now-uncaught case is a `SystemExit` raised *while suspended at the `await proc.communicate()`*, which is vanishingly rare in this CLI tool path and is OS-reaped on interpreter shutdown anyway. Keeping the narrowed form as specified.

olmlx/chat/builtin_tools.py
```diff
@@ -430,7 +433,10 @@ async def _handle_bash(args: dict) -> str | ToolError:
             tool_name="bash",
             is_user_error=False,
         )
-    except BaseException:
+    except (Exception, KeyboardInterrupt, asyncio.CancelledError):
```
dpalmqvist, line 436: Intentional, per issue #454 finding #5, which explicitly asked to narrow this `except BaseException` (the issue author classified it as cosmetic). The genuine cleanup triggers — `asyncio.CancelledError` (task cancellation/timeout) and `KeyboardInterrupt` (Ctrl-C) — are both still caught and re-raise after killing the child. The only now-uncaught case is a `SystemExit` raised *while suspended at the `await proc.communicate()`*, which is vanishingly rare in this CLI tool path and is OS-reaped on interpreter shutdown anyway. Keeping the narrowed form as specified.

dpalmqvist:
Thanks aider — reviewed all five. None are actionable for this PR; rationale below.

1. **`logits_processors.py:13` — grammar import coupling (medium):** no *new* coupling. `inference.py` already imported `GrammarSpec`/`make_processor`/`unwrap_mlx_tokenizer` from `engine.grammar` at top level before this PR; the extraction just relocates those imports. Nothing imports `logits_processors` solely for the gpt-oss filter — it's only consumed via `inference.py`, which pulls in `grammar` regardless. No startup cost added.

2. **`speculative_loaders.py:25` — imports `settings`/`SpeculativeConfig` (medium):** `model_manager` already imports both (`settings` from `config`, `SpeculativeConfig` from `registry`) at module load, so the mixin adds no new import coupling. The extraction's goal was to break the *circular* dependency (mixin ↔ model_manager), which is achieved — `speculative_loaders` never imports `model_manager` at top level; the one shared helper is pulled in lazily at call time.

3. **`speculative_loaders.py:64` — `_resolve_draft_path` relative path (high/bug):** intended, documented behavior — moved verbatim, not introduced here. A non-absolute string is an **HF repo id** (e.g. `my-org/dflash-draft`), routed through `store.ensure_downloaded`; the docstring explains that short-circuiting relative paths to local dirs would silently swap the operator's intended remote artifact. Only the *absolute*-path branch is a local dir, and it does fail fast (`FileNotFoundError`). A bad repo id surfaces a clear HF error downstream. No missing validation.

4. **`cli.py:1616` — TOKEN only matched as suffix (medium/security):** deliberate. `_TOKEN`/`_KEY` are matched as **suffixes** specifically to avoid false positives on this codebase's many legitimate `*_TOKENS`/`*_KEY...` config keys (`OLMLX_SPECULATIVE_TOKENS`, `OLMLX_PROMPT_CACHE_MAX_TOKENS`, the new `OLMLX_DEFAULT_MAX_TOKENS`, …). Matching `TOKEN` as a substring would wrongly strip those from the plist. `OLMLX_HF_TOKEN` ends in `_TOKEN`, so it's caught by design, not by accident. `SECRET`/`PASSWORD`/`CREDENTIAL` are matched as substrings since they have no such collisions.

5. **`builtin_tools.py:267` — explicit timeout case (low):** current behavior is already correct — `asyncio.TimeoutError` is an `Exception`, so the child is killed and re-raised. The bash variant already has an explicit `except asyncio.TimeoutError`. Leaving the search variant as-is to keep the diff scoped to the requested `BaseException` narrowing (#454 #5).

Issue #465, opened 2026-06-09
prompt_cache: _estimate_state_bytes silently returns 0 for unknown cache layouts


dpalmqvist:
## Problem

`_estimate_state_bytes` (`olmlx/engine/prompt_cache/store.py:~35-100`) sizes cache layers via three strategies: `.keys`/`.values` attributes, a `__dict__` walk (for TurboQuant side buffers), and a `.state` fallback (ArraysCache).

A future cache class exposing none of these contributes **0 bytes silently**. `metrics.bytes_in_ram` then undercounts and `_enforce_ram_budget` never fires, so the store can blow past `ram_budget_bytes` without any signal — the failure shows up later as Metal memory pressure with no pointer back to the cause.

## Fix

When a layer contributes 0 bytes while the entry overall is non-trivial, emit a one-time `logger.warning` naming the layer class, e.g.:

```python
if layer_bytes == 0:
    logger.warning(
        "prompt cache byte estimator does not understand %s; "
        "RAM budget accounting will undercount", type(layer).__name__
    )
```

(one-time per class to avoid log spam). Alternatively, formalize a small sizing protocol that cache classes implement, and warn on classes that don't.

_Found during a full-repo code review._

PR #476, opened 2026-06-09
fix(grammar): drop compile-cache entries even when model close partially fails (#464)

### LINKED ISSUE #464: grammar: compile cache keyed on id(tokenizer) can serve a stale grammar after a partially-failed unload
## Problem

The xgrammar compile cache is keyed on `id(tokenizer)` (`olmlx/engine/grammar.py:90-98`). The documented safety argument is that `drop_for_tokenizer` is wired into `ModelManager._close_loaded_model`, so entries die with the tokenizer.

But `unload()` **absorbs** close failures (`olmlx/engine/model_manager.py:1629-1632`):

```python
try:
    await asyncio.to_thread(self._close_loaded_model, lm)
except ExceptionGroup:
    pass  # already logged per-resource inside _close_loaded_model
```

If `_close_loaded_model` fails before/around the `drop_for_tokenizer` step, the stale cache entry persists with the dead tokenizer's `id()`. CPython freely reuses addresses, so a subsequently loaded tokenizer can collide with the key and `compile_for_tokenizer` would return a `CompiledGrammar` built for a **different vocabulary** — silent wrong-grammar constraint, very hard to debug.

Low probability, but the failure mode is silent output corruption rather than an error.

## Fix options

1. Key on a `weakref.ref(tokenizer)` / use `weakref.WeakValueDictionary`-style ownership so entries die with the object regardless of unload success.
2. Move the compile cache onto `LoadedModel` itself (per-model lifetime, dies with the model, no global keying at all). Probably the cleanest given caches are already per-model elsewhere (`vlm_prompt_cache_store` etc.).

dpalmqvist:
## Problem

The xgrammar compile cache is keyed on `id(tokenizer)`, with the documented safety argument that `drop_for_tokenizer` (wired into `ModelManager._close_loaded_model`) removes entries while the tokenizer is still alive. But `unload()` absorbs close failures, and `_close_loaded_model` had one **unguarded** step before the drop: `vlm_prompt_cache_store.clear()`. A raise there:

- skipped `drop_for_tokenizer` (and the whisper `ModelHolder` cleanup), leaving the stale id-keyed entry behind, and
- escaped as a bare exception that `unload()`'s `except ExceptionGroup` doesn't absorb (contract violation).

Since CPython freely recycles addresses, a later tokenizer could collide with the dead tokenizer's `id()` and `compile_for_tokenizer` would serve a `CompiledGrammar` built for a **different vocabulary** — silent wrong-grammar output, not an error.

## Fix (two layers)

1. **Weakref-validated ownership** (`engine/grammar.py`): both `_compile_cache` and `_compiler_cache` entries now store `(weakref.ref(tokenizer), value)`; lookups serve an entry only when the referent **is the same live object**. A stale entry can therefore never be served to a colliding tokenizer, regardless of unload success. Dead entries are swept on the next cache insert so a missed drop can't leak them forever. Non-weakref-able tokenizers (exotic stand-ins only — HF tokenizers are always weakref-able) skip caching rather than risk staleness.
2. **Deterministic drop ordering** (`engine/model_manager.py`): `drop_for_tokenizer` now runs **first** in `_close_loaded_model` (it has no ordering dependency on the other resources), so no earlier close failure can skip it; and the VLM prompt-cache clear is guarded like every other step, restoring the documented "each close attempted independently, always raises ExceptionGroup" contract.

Key format is unchanged (`(id(tokenizer), spec-hash)`), so `drop_for_tokenizer` semantics and all existing call sites are untouched.

## Testing (TDD)

New tests written first and confirmed failing for the right reasons pre-fix:

- `tests/test_model_manager.py::TestCloseLoadedModelGrammarDrop` — grammar drop still runs (and the failure folds into the `ExceptionGroup`) when `vlm_prompt_cache_store.clear()` raises, and when *every* guarded step raises.
- `tests/test_grammar.py::TestStaleEntryProtection` — a stale compile/compiler entry under a colliding id (live different owner, or dead owner) is never served and is replaced by a correctly-owned entry; dead entries are swept on insert.

Ran:

- `uv run pytest tests/ -k "grammar or model_manager"` — 351 passed, 1 skipped
- `uv run pytest tests/test_model_manager_coverage.py tests/test_inference_grammar.py tests/test_grammar_vlm_vocab.py` — 74 passed
- `uv run ruff check .` / `uv run ruff format .` — clean

CLAUDE.md structured-outputs bullet updated to describe the new cache-ownership model.

Fixes #464


olmlx/engine/grammar.py
```diff
@@ -88,23 +89,53 @@ def cache_key(self) -> str:
...
+def _make_ref(tokenizer: Any) -> weakref.ref[Any] | None:
+    """A weakref to *tokenizer*, or ``None`` if it isn't weakref-able.
+
+    HF tokenizers are plain Python classes and always weakref-able; the
+    fallback only matters for exotic stand-ins, which then simply skip
+    caching (correctness over speed).
+    """
+    try:
+        return weakref.ref(tokenizer)
+    except TypeError:
```
dpalmqvist, line 133: **[low/observability]** When `weakref.ref(tokenizer)` raises `TypeError`, both caches are silently disabled: every structured-output request with that tokenizer recompiles the grammar *and* rebuilds the `GrammarCompiler` (full-vocab walk), with no log line pointing at the cause. A log-once-per-class warning here (same pattern as the prompt-cache estimator warning from #465) would make the degradation diagnosable.

olmlx/engine/grammar.py
```diff
@@ -113,12 +144,15 @@ def _get_compiler(tokenizer: Any, vocab_size: int) -> Any:
         )
     key = id(tokenizer)
     with _compiler_cache_lock:
-        compiler = _compiler_cache.get(key)
-        if compiler is not None:
-            return compiler
+        entry = _compiler_cache.get(key)
+        if entry is not None and entry[0]() is tokenizer:
```
dpalmqvist: **[low/cleanup]** The ownership check `entry is not None and entry[0]() is tokenizer` is repeated three times (here and twice in `compile_for_tokenizer`, lines 183/196). A tiny helper, e.g. `_live_value(entry, tokenizer)`, would give the subtle deref-then-identity pattern one source of truth.

Issue #460, opened 2026-06-09
dflash: step() lacks the try/reset wrapper EAGLE/MTP have — corrupt cache state survives a mid-step failure


dpalmqvist:
## Problem

`EagleDecoder.step()` wraps its body in a try/reset (`olmlx/engine/eagle/decoder.py:423-437`), with a comment spelling out why:

> A mid-step exception in the draft loop (MLX Metal error during `.item()`) or the verify forward (OOM, shape mismatch) would otherwise leave both KV caches in a partially-modified state and the GDN capture's `_gdn_inputs` list with leftover entries from the failed verify — the next `step()` would either consume corrupt caches or trip the cardinality check inside `rollback` in a confusing way.

`MTPDecoder.step()` has the same wrapper. **`DFlashDecoder.step()` (`olmlx/engine/dflash/decoder.py:428-582`) does not**, despite having the identical structure — including a non-trimmable-target rollback (`self._capture.rollback_single(...)` at lines 561-566) that can raise.

If any exception escapes mid-step (draft forward, verify forward, rollback), the target/draft caches and the GDN capture buffer are left partially modified, and `speculative_stream_generate` will call `step()` again on the corrupt state.

## Fix

Apply the same pattern as EAGLE: split the body into `_step_impl()` and wrap:

```python
def step(self) -> tuple[list[int], int]:
    try:
        return self._step_impl()
    except Exception:
        self.reset()
        raise
```

Per CLAUDE.md TDD: a test that makes the rollback (or verify forward) raise and asserts the decoder resets (caches cleared, `_capture` closed) should come first.

_Found during a full-repo code review. Related architectural follow-up: a shared decoder base class would have prevented this divergence (separate issue)._

PR #482, opened 2026-06-10
refactor(speculative): extract SpecDecoderBase shared decoder base class (#467)

### LINKED ISSUE #467: speculative: extract a shared decoder base class (try/reset, patch lifecycle, cancel_event contract)
## Problem

The five decoders (`classic`, `pld`, `dflash`, `eagle`, `mtp`) share a protocol (`prefill`/`step`/`reset` + `verify_draft_greedy`) but no base class, and they have **already drifted**:

- Exception safety: EAGLE and MTP wrap `step()` in try/reset; DFlash doesn't (#460 — a real bug produced by exactly this duplication).
- `cancel_event`: classic/pld honor it via shared `_chunked_prefill`, MTP at sub-chunk boundaries, dflash/eagle accept-and-ignore — each wired by hand.
- `_patch_model`/`_unpatch_model` + `_GDNStateCapture` install/teardown: three near-identical copies (dflash, eagle, mtp), with naming drift (`_hidden_capture` vs `_hidden_storage`).
- Tracing: classic fully instrumented, PLD `spec.verify` only, the rest none (already noted in CLAUDE.md as a follow-up blocked on "each a bespoke class with no shared base").
- MTP's decoder "mirrors EagleDecoder" — i.e. the trim arithmetic and capture lifecycle exist as a third copy.

## Proposal

A small `SpecDecoderBase` owning the mechanical parts, not the algorithms:

- `step()` → try/reset wrapper around an abstract `_step_impl()` (makes #460-class bugs impossible).
- `prefill(prompt, *, cancel_event=None, segmented=None)` as the canonical signature with accept-and-ignore defaults, so `speculative_stream_generate` stops special-casing signatures.
- Hook/capture lifecycle: `_patch_model` install + idempotency, `_GDNStateCapture` create/close, `reset()` teardown ordering.
- Optional: the `spec.prefill`/`spec.step`/`spec.verify` tracing seams, which then come for free for all strategies.

Keep `classic`/`pld`'s `_SpecCacheStore` reuse as a mixin or subclass-only feature — no need to generalize it.

This is deliberately shallow: the verify/draft algorithms stay in the subclasses; only the lifecycle and exception-safety scaffolding moves up.

dpalmqvist:
## Summary

Implements #467: a small `SpecDecoderBase` (`engine/spec_decoder_base.py`) that owns the mechanical parts every speculative decoder had hand-rolled (and drifted on), while the verify/draft algorithms stay in the subclasses. All six decoders (classic, pld, dflash, eagle, mtp, self) now inherit it.

### What the base owns

- **`step()` → try/reset around abstract `_step_impl()`** — makes the #460 bug class impossible. Classic, PLD, and self-speculative previously had **no** exception boundary and could leave partially-modified KV caches behind; now a mid-step failure forces a clean teardown (re-`prefill` to recover).
- **`prefill()` → reset + try/reset around abstract `_prefill_impl()`** with the canonical signature `prefill(prompt, *, segmented=None, cancel_event=None)`. `speculative_stream_generate` now passes both kwargs unconditionally — the per-decoder signature special-casing in the bridge is gone. DFlash's prefill gains error cleanup it never had (eagle/mtp's hand-rolled wrappers are subsumed).
- **Ordered `reset()` teardown**: GDN capture close *first* (it holds `_GDN_PATCH_LOCK`) → `_unpatch_model` → draft unbind → subclass `_reset_state()`, each flag cleared in a `finally` so double-calls never double-unpatch.
- **Hook lifecycle helpers**: `_install_layer_hooks` / `_bind_draft` / `_install_gdn_capture` replace the three near-identical copies in dflash/eagle/mtp (and the `_hidden_capture` vs `_hidden_storage` naming drift no longer hides the duplication).
- **Tracing seams**: `spec.prefill` / `spec.step` / `spec.verify` now emit for **every** strategy — closes the CLAUDE.md follow-up that was blocked on "each a bespoke class with no shared base". `spec.step` records proposed/accepted via the stats-counter delta, so the semantics match what each strategy actually proposed.
- **Shared stats counters + `stats_summary()`** with a `_stats_extra()` hook for the strategy-specific keys (`lambda`, `block_size`, `ema_acceptance_rate`).

### Moves / re-exports

- `_LayerHook` / `_patch_model` / `_unpatch_model` and `verify_draft_greedy` move to `spec_decoder_base`; re-exported from `dflash.decoder` / `speculative` so existing importers (cli, prepare, scripts, tests) keep working.
- Classic/PLD keep their overridden `close()` (their GDN capture and `_SpecCacheStore` are decoder-lifetime, created in `__init__`); for the hook decoders `close()` is `reset()` (unchanged semantics, now inherited).

### Behavior changes (intentional)

- Classic/PLD/self `step()` and all `prefill()`s now reset on exception (the point of the issue).
- Self-speculative `close()` now resets per-request state instead of being a state-preserving no-op (uniform lifecycle; nothing reads decoder state after close — test updated to pin the new semantics).
- Tests that monkeypatched `_GDNStateCapture` / `verify_draft_greedy` on the eagle/mtp modules now patch the base-module seams.

## Test plan

- New `tests/test_spec_decoder_base.py` (TDD: written first, watched fail): base try/reset semantics, canonical signature forwarding, teardown ordering + idempotency, stats merging, tracing spans, all-six-subclass check, bridge-always-passes-kwargs check, and per-decoder reset-on-exception tests for the three decoders that previously lacked the boundary.
- Full suite: **4524 passed, 4 skipped**; ruff check + format clean.

Closes #467


olmlx/engine/spec_decoder_base.py
```diff
@@ -0,0 +1,405 @@
...
+                        "proposed": self._stats_proposed - proposed_before,
+                        "accepted": len(accepted),
+                    }
+                )
+                return accepted, num_draft
+        except Exception:
+            self.reset()
+            raise
+
+    def reset(self) -> None:
```
dpalmqvist, line 285: **reset() does not deliver its documented contract.** `prefill()`/`step()`'s except-paths say "`reset()` swallows nested teardown errors per step so the caller sees the original error", but each teardown step here is `try/finally` with no `except` — a raising `capture.close()` propagates, **replaces the original inference error**, and skips `_unpatch_model`/`unbind`/`_reset_state()` (hooks stay installed on the shared target). Pre-existing in eagle/mtp, but the base now codifies the false contract for all six decoders. Fix: swallow-and-log each teardown step (keep the `finally` flag-clears) so reset() never raises — `close()`/`__del__` also rely on that.

olmlx/engine/spec_decoder_base.py
```diff
@@ -0,0 +1,405 @@
...
+    def _reset_state(self) -> None:
+        """Clear strategy-specific per-request state (caches, seeds,
+        pending tokens). Called at the end of every ``reset()``; must be
+        idempotent and must not raise."""
+
+    # ----- shared helpers ----------------------------------------------------
+
+    def _install_layer_hooks(self, layer_ids: list[int], storage: list[Any]) -> None:
+        """Install hidden-capture hooks on the target; ``reset()`` unpatches."""
+        _patch_model(self._target, layer_ids, storage)
```
dpalmqvist: **Set `self._patched = True` _before_ `_patch_model`, not after.** `_patch_model` can raise mid-loop with some layers already wrapped (e.g. a corrupt dflash `target_layer_ids` like `[5, 999]` — the `all()` check short-circuits at the first unwrapped valid id and never bounds-checks the later one). With the flag still False, `reset()` skips `_unpatch_model` and the `_LayerHook` proxies leak onto the shared target model for every subsequent request. `_unpatch_model` is idempotent and scans all layers, so flag-first is strictly safer. (Pre-existing ordering hoisted from dflash/eagle/mtp, but the base is the right place to fix it once.)

olmlx/engine/speculative.py
```diff
@@ -603,11 +567,13 @@ def __init__(
     def close(self) -> None:
```
dpalmqvist, line 563: **Classic/PLD `close()` should end with `reset()` and guard the capture close.** (a) No try/finally: if `_gdn_capture.close()` raises, `_cache_store.clear()` is skipped — and `ModelManager._close_loaded_model` deliberately preserves the decoder reference on close failure, pinning multi-GB snapshots. (b) Unlike the base `close()==reset()` (and the PR's own change to self-speculative), the last request's working KV caches survive `close()` here. A trailing `self.reset()` is safe: `unload()` raises `ActiveRequestsError` on in-flight requests and eviction skips `active_refs > 0`, and classic/PLD never set the base's request-lifetime hook fields. Same for `PromptLookupDecoder.close()` below.

olmlx/engine/spec_decoder_base.py
```diff
@@ -0,0 +1,405 @@
...
+        self._capture, self._capture_buffer = GDNStateCapture.for_model(self._target)
+
+    def _verify_greedy(
+        self, draft_tokens: list[int], target_logits: mx.array
+    ) -> list[int]:
+        """:func:`verify_draft_greedy` under a ``spec.verify`` span."""
+        with _tracing.span("spec.verify", strategy=self._strategy()):
+            return verify_draft_greedy(draft_tokens, target_logits)
+
+    def _strategy(self) -> str:
```
dpalmqvist: **`_strategy()` re-runs import machinery + a dict lookup up to twice per decode step** (step span + verify span), eagerly even with tracing disabled, since `span()` only short-circuits after kwargs are built. Impact is noise next to the MLX forward (~100-300ns vs ms), but the label is invariant for the instance — resolve it once in `__init__` (`self._strategy_label`) and use that in the spans.

olmlx/engine/speculative.py
```diff
@@ -499,6 +462,7 @@ def __init__(
         tree_max_nodes: int = 8,
         cache_slots: int = 0,
     ):
+        super().__init__()
```
dpalmqvist, line 466: **Redundant stats re-init shadowing the base.** `super().__init__()` already initializes `_stats_steps/_stats_proposed/_stats_accepted_draft`; these re-declarations (also `PromptLookupDecoder.__init__` and `self_speculative/decoder.py` under `# Stats (reset on prefill)`) were removed from dflash/eagle/mtp in this same PR but left in the other three — drop them so counter ownership has one home.

olmlx/engine/eagle/decoder.py
```diff
@@ -40,19 +40,14 @@
...
 from olmlx.engine.gdn_rollback import (
     _HAS_GDN,
     find_gdn_class as _find_gdn_class,
-    GDNBuffer as _GDNBuffer,
-    GDNStateCapture as _GDNStateCapture,
     get_model_layers as _get_layers,
 )
 from olmlx.engine.eagle.draft_model import EagleDraftModel
-from olmlx.engine.speculative import _eval_cache, verify_draft_greedy
+from olmlx.engine.spec_decoder_base import SpecDecoderBase
```
dpalmqvist, line 49: **Stale module docstring (also mtp/decoder.py):** still says the non-trim path installs "DFlash's `_GDNStateCapture`" and that "Layer hooking is shared with DFlash via `_patch_model` / `_get_layers` / `_LayerHook`" — both now come from `SpecDecoderBase` (`_install_gdn_capture` / `_install_layer_hooks`) and `gdn_rollback.GDNStateCapture`. Same staleness in CLAUDE.md's cross-request-KV-reuse bullet ("passes `segmented` only when non-None, so dflash/eagle/self keep their narrower `prefill` signature") — the conditional pass-through was removed in this PR.

dpalmqvist (commented):
Claude Code review (high effort, 7 finder angles + adversarial verify). 6 findings below — all are in new-base-class code or made newly visible by it; none block correctness of the refactor itself. Fixes incoming on this branch.

olmlx/engine/spec_decoder_base.py
```diff
@@ -0,0 +1,429 @@
...
+                )
+                return accepted, num_draft
+        except Exception:
+            self.reset()
+            raise
+
+    def reset(self) -> None:
+        """Tear down request state; safe to call repeatedly. Never raises
+        from the teardown steps themselves.
+
```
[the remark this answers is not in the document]
dpalmqvist: Not reachable: `GDNStateCapture.close()` releases `_GDN_PATCH_LOCK` in a `finally` (gdn_rollback.py — `close()` wraps the `__call__` restore in `try`, and the `finally` block sets `_closed`, clears the patch refs, and calls `_GDN_PATCH_LOCK.release()` unconditionally). A raising restore therefore still releases the lock, and `close()` is idempotent via the `_closed` flag. The swallow in `reset()` also doesn't change lock state relative to the previous behavior — propagating the same exception (the pre-#482 code) didn't release anything either; it just additionally skipped the unpatch/unbind/`_reset_state` teardown that the swallow now guarantees runs.

PR #485, opened 2026-06-10
feat(packaging): move audio/TTS deps to an [audio] extra, drop torch/torchvision (#469)

### LINKED ISSUE #469: packaging: move audio/TTS deps (torch, torchvision, mlx-audio, misaki, spaCy model) to an extra
## Problem

A default `uv sync` of olmlx pulls several GB of dependencies that only serve the audio endpoints:

- `torch>=2.10.0` + `torchvision>=0.25.0`
- `mlx-audio` + `misaki[en]` + spaCy + the `en_core_web_sm` GitHub-release wheel pinned via `[tool.uv.sources]`

Users who only want chat completions pay the full footprint. The careful comments in `pyproject.toml` around the spaCy pin (runtime auto-download not importable in-process, lockstep major.minor with spaCy) show how much maintenance this dependency already costs — better confined to the users who need it.

## Proposal

- Add an `[audio]` extra (mirroring the existing `[voice]` / `[otel]` pattern) carrying `mlx-audio`, `misaki[en]`, `en-core-web-sm`.
- `/v1/audio/speech` (and `--voice` TTS) return a clear 4xx/exit hint when the extra isn't installed — same UX as `--voice` without `sounddevice` today.
- Audit whether `torch`/`torchvision` are needed at *serving* time at all or only via `transformers` extras / dflash-eagle training paths; if prepare-time only, move them to a `[train]` extra.

dpalmqvist:
Closes #469.

A default `uv sync` no longer pulls the several-GB audio/TTS stack.

## Changes

- **New `[audio]` extra**: `mlx-audio`, `misaki[en]`, `en-core-web-sm` (the spaCy wheel pin stays in `[tool.uv.sources]`, which applies to optional deps too). The careful spaCy-pin comments moved with them.
- **`[voice]` is now self-referential**: `olmlx[audio]` + `sounddevice`, so `uv sync --extra voice` keeps pulling everything `--voice` needs. Verified via `uv sync --extra voice --dry-run`.
- **Clear error UX when the extra is missing** (same pattern as `--voice` without sounddevice):
  - `/v1/audio/speech`: `_load_model_tts` converts the ImportError into `ValueError("... uv sync --extra audio ...")`, which the router's existing primed-first-chunk path maps to HTTP 400 — instead of an opaque 500.
  - `olmlx chat --voice`: `_check_voice_deps` now gates on `mlx_audio` as well and exits with the install hint.
- **torch audit** (per the issue): olmlx never imports torch; `xgrammar` and `mlx-whisper` both require it transitively, so it is always installed regardless — the explicit pin is dropped and no `[train]` extra is needed. **torchvision is removed outright**: only olmlx declared it, and transformers 5.x auto-falls-back to its PIL image-processor backend (`_resolve_backend` → `"pil"`, every VLM ships a `*Pil` variant) — verified in a torchvision-free venv, including `Gemma3ImageProcessorPil` importability.

## Tests

- `tests/test_load_tts_model.py`: reworked to stub `mlx_audio` via `sys.modules` (the suite must pass without the extra installed, as CI's bare `uv sync` now is); new test for the ValueError install hint.
- `tests/test_voice_cli.py`: new test that missing `mlx_audio` exits with the hint.
- `tests/test_packaging_audio_extra.py`: new guard pinning the dependency layout (audio deps not in core, present in `[audio]`, `[voice]` includes `olmlx[audio]`).
- `tests/live/test_{tts_speech,voice_real}.py`: skip with an install-hint reason when `mlx_audio` is absent instead of failing confusingly.

Full suite in the slimmed env: **4525 passed, 3 skipped** (`-m "not real_model"`). ruff check + format clean.


dpalmqvist:
Re: aider-code-review findings —

1. **cli.py — voice dep check deferred**: Not the case. `_check_voice_deps()` runs at the top of the `--voice` branch in `cmd_chat` (cli.py:2132), before `ensure_loaded(tts_model)` and before any TTS call, so a missing `mlx_audio` exits with the install hint upfront. No change.

2. **model_manager.py — broaden ImportError catch**: Declined. A broken-but-installed mlx-audio (e.g. native lib failure raising OSError) would then surface as "install with `uv sync --extra audio`" — misleading advice for a user who already ran it. `ImportError`/`ModuleNotFoundError` is exactly the not-installed case the hint can fix; real import-time breakage should keep its traceback. `_require_sounddevice`'s broad catch isn't analogous: its hint text explicitly covers the PortAudio dylib-load failure mode.

3. **test only checks the self-reference string**: Addressed in d369adc — `test_lockfile_expands_voice_extra_to_audio_deps` now asserts uv.lock expanded `olmlx[voice]` to the audio deps + sounddevice.

dpalmqvist:
Re: second aider-code-review round —

1. **model_manager.py — ImportError misses ModuleNotFoundError**: `ModuleNotFoundError` is a subclass of `ImportError`, so it is caught. The init-failure case is the same as round 1's finding 2 — declined for the same reason (a broken-but-installed package should keep its real traceback, not get a misleading install hint).

2. **cli.py — gate doesn't cover TTS transitives**: Addressed in 84dddce — `_check_voice_deps` now also gates on `misaki` and `en_core_web_sm`, so a hand-installed mlx-audio without the extra fails upfront with the hint.

3. **"uv could stop expanding self-referential extras"**: Covered by `test_lockfile_expands_voice_extra_to_audio_deps` — uv.lock stores the *resolved* expansion, so if uv stopped expanding, lock regeneration would empty the voice extra's audio deps and the test fails.

4. **"lock check doesn't pin the voice resolution" (high)**: The test does pin it: `assert audio | {"sounddevice"} <= voice` asserts directly on the lockfile's resolved `voice` extra dependency set, not on `audio`. No change.

5. **"flat package stub"**: `_stub_mlx_audio` installs all three nested entries (`mlx_audio`, `mlx_audio.tts`, `mlx_audio.tts.utils`) in `sys.modules`, and the missing-case test nulls all three. The real nested import path is exercised against the actual package by `tests/live/test_tts_speech.py`. No change.

olmlx/cli.py
```diff
@@ -1950,13 +1950,25 @@ def _absent(name: str) -> bool:
...
-    # Only sounddevice lives in the [voice] extra; mlx-audio (Kokoro TTS) is a
-    # core dependency for /v1/audio/speech, so `--extra voice` wouldn't install
-    # it. Gate on sounddevice alone — the dep the hint can actually fix.
-    if _absent("sounddevice"):
+    # The [voice] extra carries sounddevice (PortAudio) and, via the
+    # self-referential olmlx[audio] requirement, the Kokoro TTS stack —
+    # all moved out of core deps in #469. Gate on the TTS transitives too:
+    # a hand-installed mlx-audio (pip install, no extra) lacks misaki /
+    # en_core_web_sm and would otherwise fail at the first TTS call.
+    missing = [
```
dpalmqvist, line 1958: The state "installed [voice] without [audio]" can't occur: [voice] is self-referential (olmlx[audio] + sounddevice), so `uv sync --extra voice` installs every package this gate checks. The hint is the one command that fixes all four. No change.

tests/test_packaging_audio_extra.py
```diff
@@ -0,0 +1,66 @@
...
+    for dep in ("mlx-audio", "misaki", "en-core-web-sm"):
+        assert dep in audio, f"{dep} missing from the [audio] extra"
+
+
+def test_voice_extra_includes_audio_extra(project):
+    # `--voice` speaks replies via Kokoro TTS, so installing [voice] alone
+    # must also pull the [audio] deps (self-referential extra).
+    voice = project["optional-dependencies"]["voice"]
+    assert any(
+        req.replace(" ", "").lower().startswith("olmlx[audio]") for req in voice
```
dpalmqvist, line 51: What users get from `uv sync --extra voice` is the resolved package set, which is exactly what the assertion pins (audio deps + sounddevice ⊆ voice's resolved deps). Whether uv encodes the self-reference as an edge or flattens it is a lock-format internal with no user-visible effect. No change.

olmlx/engine/model_manager.py
```diff
@@ -2943,7 +2943,16 @@ def _load_model_tts(self, hf_path: str, load_path: str):
         """
         from pathlib import Path
 
-        import mlx_audio.tts.utils as tts_utils
+        try:
+            import mlx_audio.tts.utils as tts_utils  # type: ignore[import-not-found]
+        except ImportError as exc:
```
dpalmqvist, line 2948: If misaki / en_core_web_sm are absent and that surfaces during import resolution, it raises ModuleNotFoundError — a subclass of ImportError — so this guard does catch it and the hint (`uv sync --extra audio`) is exactly right for that state. Non-ImportError failures from a *broken* install should keep their real traceback rather than a misleading install hint (same disposition as rounds 1–2). No change.

dpalmqvist:
Re: latest aider-code-review ("critical regression: en_core_web_sm treated as a Python package") — verified false. spaCy model wheels install an importable package (that's the documented usage, `import en_core_web_sm`, and the reason the wheel is pinned in `[tool.uv.sources]` at all). Empirical check in a venv with the [voice] extra:

```
sounddevice -> True
mlx_audio -> True
misaki -> True
en_core_web_sm -> True   # importlib.util.find_spec
_check_voice_deps: OK (no exit)
```

No change.

Issue #463, opened 2026-06-09
Make the event-loop-affinity contract explicit (registry RMW, unload check-then-pop, cache stores)


dpalmqvist:
## Problem

A lot of correctness rests on an **implicit** "all mutations happen on the asyncio event-loop thread" contract that is documented in comments but never enforced:

- `olmlx/engine/registry.py` (~1330-1413): `_save_mappings_locked` does a read-modify-write of `models.json`; the lock "serializes disk I/O but does not cover dict mutations in add_mapping/remove. This is safe today because all callers run on the single-threaded asyncio event loop." Two interleaved pulls that await between mutate and save could lose a mapping if a caller ever moves off-loop.
- `olmlx/engine/model_manager.py:1620-1628`: `unload()`'s `active_refs > 0` check-then-pop is atomic only because there's no `await` between them and all `acquire_ref()` callers run on the loop. Same for `_expire_stale()`.
- `_SpecCacheStore` and the prompt-cache store's in-memory mutation paths rely on the same affinity, while worker threads genuinely exist nearby (decode thread, `to_thread` disk I/O, prefetch pools).

Each individual site is fine today; the risk is that the invariant is invisible at the call sites, so a future refactor (moving a caller into a thread, adding an `await` inside a critical section) turns into silent corruption rather than an error.

## Proposal

Add a tiny debug helper and call it at the top of the mutating entry points:

```python
def assert_loop_thread() -> None:
    if settings.debug_checks:  # or __debug__
        asyncio.get_running_loop()  # raises off-loop
```

(or compare `threading.get_ident()` against the loop's thread). Cheap, and converts a violated invariant into an immediate, attributable failure. Alternatively for the registry specifically: hold the lock across mutation+save instead of save only.

_Found during a full-repo code review._

PR #486, opened 2026-06-10
fix(prompt_cache): skip re-spilling a disk-restored entry the RAM budget evicts (#466)

### LINKED ISSUE #466: prompt_cache: async_get can restore from disk and immediately re-spill the same entry under RAM-budget pressure
## Problem

In `PromptCacheStore.async_get` (`olmlx/engine/prompt_cache/store.py:~548-562`), a disk hit is restored into memory via `_set_in_memory`, then `_enforce_ram_budget()` runs. If the store is at/over budget, the **just-restored entry** can be the one evicted (or evicted alongside others), getting written straight back to disk in the same call.

Correctness is fine — the final unlink of the original disk file is guarded by `cache_id in self._entries`, so the only copy is never deleted — but under sustained budget pressure this is a read-then-rewrite of potentially multi-GB cache state per request: pure disk churn with no cache benefit.

## Fix

After `_enforce_ram_budget()`, check whether the restored entry survived:

- If it was evicted, skip re-saving it (it's already on disk — just don't unlink) and treat the call as a effectively-cold hit.
- Optionally, exempt the entry being restored from eviction for the duration of the call (evict strictly older entries first — `popitem(last=False)` should already do this since the restored entry is MRU; worth a test pinning that behavior so a refactor doesn't regress it).

dpalmqvist:
## Summary

Fixes #466. In `PromptCacheStore.async_get`, a disk hit is restored into memory via `_set_in_memory`, then `_enforce_ram_budget()` runs. If the just-restored entry is evicted by the budget (e.g. it alone exceeds `ram_budget_bytes`), it was written straight back to disk in the same call — a read-then-rewrite of potentially multi-GB cache state per request, with no cache benefit.

## Fix

The eviction-spill loop in `async_get` now skips the restored `cache_id`: its original disk file is guaranteed intact on this path (`_read_from_disk` returns the path only on a hit, and the final unlink is guarded on residency), so re-saving identical bytes is pure churn. Other budget evictees still spill normally.

The sync `get()` path is not touched — the disk-backed LLM path only uses `async_get` (the sync `store.get` call in `inference.py` is the in-memory-only `VlmPromptCacheStore`).

## Tests

- New: `test_async_get_restored_entry_evicted_by_budget_not_respilled` — restores an over-budget entry, asserts the displaced residents spill, the restored entry is not re-saved, its disk file survives, and the caller still gets the loaded state.
- Existing `test_async_get_disk_restore_enforces_ram_budget` already pins that the restored entry is MRU (older entries evict first), per the issue's suggestion.
- Hoisted the duplicated `_sized_state` helper to module level (now used three times).

### DIFF
--- a/olmlx/engine/prompt_cache/store.py
+++ b/olmlx/engine/prompt_cache/store.py
@@ -597,6 +597,12 @@ class PromptCacheStore:
         # it now so a chain of disk hits can't quietly blow the budget
         # until the next write.
         for over_id, over_state in self._enforce_ram_budget():
+            if over_id == cache_id:
+                # The just-restored entry bounced straight back out under
+                # budget pressure. Its original disk file is still intact
+                # (the unlink below is guarded on residency), so rewriting
+                # the same bytes would be pure disk churn — skip the save.
+                continue
             await self._to_thread_traced(self._save_to_disk, over_id, over_state)
         # Only unlink the original disk file if the restored entry is
         # still in RAM. If the budget evicted it back to disk, the file
--- a/olmlx/engine/prompt_cache/store.py
+++ b/olmlx/engine/prompt_cache/store.py
@@ -9,6 +9,7 @@ import asyncio
 import hashlib
 import json
 import logging
+import os
 import re
 import shutil
 from collections import OrderedDict
@@ -283,6 +284,25 @@ class PromptCacheStore:
                 file_path.unlink(missing_ok=True)
                 self._refresh_disk_bytes()
                 return None
+
+    def _touch_or_resave(
+        self, cache_id: str, state: CachedPromptState, file_path: Path
+    ) -> None:
+        """Refresh the surviving disk copy of a budget-bounced restore.
+
+        Called from async_get's spill loop when the RAM budget evicts the
+        entry that was just restored from ``file_path``. Rewriting the
+        same bytes would be pure churn (issue #466), but the file's mtime
+        must move: ``_cleanup_disk`` evicts oldest-mtime files under the
+        disk cap, so a stale timestamp would make the just-used entry the
+        first one deleted. If a sibling evictee's save already triggered a
+        cleanup that removed the file, fall back to a real save so the
+        entry isn't lost from both tiers.
+        """
+        try:
+            os.utime(file_path)
+        except FileNotFoundError:
+            self._save_to_disk(cache_id, state)
 
     def _unlink_and_refresh(self, file_path: Path) -> None:
         """Unlink a disk cache file and refresh ``bytes_on_disk``.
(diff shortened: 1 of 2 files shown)


olmlx/engine/prompt_cache/store.py
```diff
@@ -597,6 +597,12 @@ async def async_get(self, cache_id: str) -> CachedPromptState | None:
         # it now so a chain of disk hits can't quietly blow the budget
         # until the next write.
         for over_id, over_state in self._enforce_ram_budget():
+            if over_id == cache_id:
+                # The just-restored entry bounced straight back out under
+                # budget pressure. Its original disk file is still intact
+                # (the unlink below is guarded on residency), so rewriting
+                # the same bytes would be pure disk churn — skip the save.
+                continue
```
dpalmqvist, line 638: **[code-review] Lost-from-both-tiers regression when the disk size cap fires mid-loop.**

`_save_to_disk` runs `_cleanup_disk()` on every save, and `_cleanup_disk` deletes oldest-mtime files while the total exceeds `prompt_cache_disk_max_gb` (always set when the disk tier is enabled). The restored entry is last in the evictee list (it was inserted MRU), so the sibling evictees ("a", "b") are saved *first* — and their cleanups can delete the restored entry's original file, which has the **oldest mtime** in the directory. When the loop then reaches the restored entry, this `continue` skips the re-save, leaving the entry in **neither RAM nor disk**. Pre-PR code re-saved it at this point, re-creating the file. Fix: in the skip branch, touch-or-resave — `os.utime` the surviving file, falling back to `_save_to_disk` if it's gone.

olmlx/engine/prompt_cache/store.py
```diff
@@ -597,6 +597,12 @@ async def async_get(self, cache_id: str) -> CachedPromptState | None:
         # it now so a chain of disk hits can't quietly blow the budget
         # until the next write.
         for over_id, over_state in self._enforce_ram_budget():
+            if over_id == cache_id:
+                # The just-restored entry bounced straight back out under
+                # budget pressure. Its original disk file is still intact
```
dpalmqvist: **[code-review] Stale-mtime LRU inversion for the surviving disk file.**

The pre-PR rewrite gave the bounced entry a fresh mtime; skipping it leaves the file with the mtime from its original spill. `_cleanup_disk` evicts by oldest mtime, so under the disk cap the *just-accessed* entry becomes the first one deleted — an LRU inversion. Touching the file's mtime (`os.utime`) in the skip branch restores recency semantics without rewriting the bytes.

olmlx/engine/prompt_cache/store.py
```diff
@@ -284,6 +285,25 @@ def _load_from_disk(self, cache_id: str) -> CachedPromptState | None:
                 self._refresh_disk_bytes()
                 return None
 
+    def _touch_or_resave(
+        self, cache_id: str, state: CachedPromptState, file_path: Path
+    ) -> None:
```
dpalmqvist, line 290: Declining this one: the fallback runs `_save_to_disk` in a worker thread exactly like the adjacent pre-existing spill calls (`await self._to_thread_traced(self._save_to_disk, …)` in `async_get`/`async_set`) — `_cleanup_disk`'s metrics writes from a worker thread are the module's existing, documented pattern ("only pure disk I/O is dispatched to a worker thread"; metrics are GIL-safe attribute updates on a single-user server). Making the event-loop-affinity contract explicit is tracked in #463; this PR doesn't change the synchronization story.

olmlx/engine/prompt_cache/store.py
```diff
@@ -284,6 +285,25 @@ def _load_from_disk(self, cache_id: str) -> CachedPromptState | None:
                 self._refresh_disk_bytes()
                 return None
 
+    def _touch_or_resave(
```
dpalmqvist, line 288: Fixed in 2b16820: `_touch_or_resave` now catches the broader `OSError` and logs a warning instead of propagating (matching `_save_to_disk`'s internal best-effort handling); `FileNotFoundError` still falls back to a real save. Regression test: `test_async_get_touch_failure_does_not_fail_request`.

PR #488, opened 2026-06-10
feat(engine): make the event-loop-affinity contract explicit (#463)

### LINKED ISSUE #463: Make the event-loop-affinity contract explicit (registry RMW, unload check-then-pop, cache stores)

dpalmqvist:
Closes #463.

## Problem

Correctness of several lock-free critical sections rested on an *implicit* "all mutations happen on the asyncio event-loop thread" contract, documented only in comments:

- `ModelRegistry`: `add_mapping`/`remove` mutate `_mappings`/`_dirty_keys` and then read-modify-write `models.json`; the save lock only serializes disk I/O.
- `ModelManager.unload()` / `_expire_stale()`: the `active_refs > 0` check-then-pop is atomic only because every caller runs on the loop.
- `PromptCacheStore` / `VlmPromptCacheStore`: `_entries`/`_radix` mutations (including LRU promotion in `get`) have no internal locking, while worker threads genuinely exist nearby (decode thread, `to_thread` disk I/O, prefetch pools).

Each site is fine today; the risk is that the invariant is invisible at the call sites, so a future refactor that moves a caller off-loop turns into silent corruption rather than an error.

## Change

**New `olmlx/utils/loop_affinity.py`** — the app lifespan binds the loop thread at startup (`bind_loop_thread`) and unbinds at shutdown (try/finally, so a failed startup or repeated `create_app()` in tests leaves no stale binding). Mutating entry points call `assert_loop_thread("<Class.method>")`, which raises an attributable `RuntimeError` when called from any other thread.

- **Unbound = no-op**: CLI paths (`olmlx models`, `olmlx chat`) never bind and have no concurrent mutators, so they're unaffected.
- **Always on**: the check is a single thread-ident comparison — cheap enough not to hide behind a debug flag, so violations surface in production too rather than only in tests.

**Instrumented sites**: `ModelRegistry.{add_mapping,add_alias,remove}`; `ModelManager.{unload,_expire_stale}`; `PromptCacheStore.{get,set,remove,clear,evict_all_to_disk,fetch_nearest,takeover,async_get,async_set,async_evict_all_to_disk}` (the thread-offloaded pure-disk-I/O helpers `_save_entries_to_disk`/`_read_from_disk` are deliberately untouched); `VlmPromptCacheStore.{get,insert,clear}`.

**`_SpecCacheStore` is the exception** the issue lumped in: it legitimately runs on the *decode worker thread*, serialized by the inference lock — loop affinity is not its contract. It gets its own explicit enforcement instead: a non-blocking try-acquire guard (`_serialized()`) on `find`/`insert`/`clear` that raises on overlapping access instead of silently corrupting the LRU list. Sequential access from different threads (its normal life) stays legal.

**Lifespan refactor**: the startup/shutdown body moved verbatim to `_lifespan_inner`; the public `lifespan` is a thin bind/unbind wrapper. The shutdown-ordering inspection test in `test_distributed.py` now reads the inner function's source and asserts both markers are actually found (previously `-1 < -1` could pass silently).

## Tests

`tests/test_loop_affinity.py` (19 tests, written first per TDD): helper semantics, off-loop raises + no-mutation for every instrumented entry point, CLI-style unbound no-op, spec-store concurrent-access guard, and lifespan bind/unbind.

Full suite: 4559 passed, 3 skipped. ruff check + format clean.

### DIFF
--- a/olmlx/app.py
+++ b/olmlx/app.py
@@ -34,6 +34,7 @@ from olmlx.routers import (
     status,
 )
 from olmlx.routers import metrics as metrics_router
+from olmlx.utils import loop_affinity
 from olmlx.utils import metrics as metrics_mod
 from olmlx.utils import tracing as tracing_mod
 
@@ -42,6 +43,21 @@ logger = logging.getLogger("olmlx")
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
+    # Bind the event-loop thread so the lock-free mutators (registry RMW,
+    # unload's check-then-pop, prompt-cache stores) can enforce their
+    # loop-affinity contract (issue #463).  Unbound in the finally so a
+    # failed startup or repeated create_app() in tests leaves no stale
+    # binding behind.
+    loop_affinity.bind_loop_thread()
+    try:
+        async with _lifespan_inner(app):
+            yield
+    finally:
+        loop_affinity.unbind_loop_thread()
+
+
+@asynccontextmanager
+async def _lifespan_inner(app: FastAPI):
     # SIGUSR1 dumps Python stacks of all threads to stderr — use
     # `kill -USR1 <pid>` to diagnose hangs without needing sudo/py-spy.
     import faulthandler
--- a/olmlx/engine/prompt_cache/store.py
+++ b/olmlx/engine/prompt_cache/store.py
@@ -691,8 +691,14 @@ class PromptCacheStore:
             await asyncio.to_thread(self._save_entries_to_disk, snapshot)
 
     def clear(self) -> None:
-        """Remove all cache entries and disk cache files."""
-        assert_loop_thread("PromptCacheStore.clear")
+        """Remove all cache entries and disk cache files.
+
+        Deliberately NOT loop-affine (#463): the only production caller is
+        ``_close_loaded_model``, which runs on a worker thread via
+        ``asyncio.to_thread`` during unload/eviction/expiry — by then the
+        LoadedModel is popped from ``_loaded``, so the closer owns this store
+        exclusively and no loop-side caller can race it.
+        """
         self._entries.clear()
         self._radix = PrefixCacheIndex()
         self.metrics.bytes_in_ram = 0
--- a/olmlx/engine/prompt_cache/store.py
+++ b/olmlx/engine/prompt_cache/store.py
@@ -698,6 +698,10 @@ class PromptCacheStore:
         ``asyncio.to_thread`` during unload/eviction/expiry — by then the
         LoadedModel is popped from ``_loaded``, so the closer owns this store
         exclusively and no loop-side caller can race it.
+
+        Note this does blocking disk I/O (``shutil.rmtree``) — do not call it
+        from the event loop; loop-side cleanup goes through ``remove`` /
+        ``async_evict_all_to_disk`` instead.
         """
         self._entries.clear()
         self._radix = PrefixCacheIndex()
--- a/olmlx/engine/prompt_cache/store.py
+++ b/olmlx/engine/prompt_cache/store.py
@@ -20,6 +20,7 @@ from olmlx.engine.prompt_cache.metrics import CacheMetrics
 from olmlx.engine.prompt_cache.radix import PrefixCacheIndex
 from olmlx.engine.prompt_cache.state import CachedPromptState
 from olmlx.utils import tracing as _tracing
+from olmlx.utils.loop_affinity import assert_loop_thread
 
 try:
     from mlx_lm.models.cache import (
@@ -376,6 +377,9 @@ class PromptCacheStore:
 
         Checks memory first, then disk.  Returns None if not found in either.
         """
+        # _entries/_radix have no internal locking: every mutating entry point
+        # (including LRU promotion here) must run on the loop thread (#463).
+        assert_loop_thread("PromptCacheStore.get")
         state = self._entries.get(cache_id)
         if state is not None:
             self._entries.move_to_end(cache_id)
@@ -446,6 +450,7 @@ class PromptCacheStore:
         cleanup (different .cache object), or None when no cleanup is needed.
         Evicted entries are saved to disk if disk offload is enabled.
         """
+        assert_loop_thread("PromptCacheStore.set")
         evicted_id, evicted = self._set_in_memory(cache_id, state)
         if evicted_id is not None and evicted is not None:
             self._save_to_disk(evicted_id, evicted)
@@ -456,6 +461,7 @@ class PromptCacheStore:
 
     def remove(self, cache_id: str) -> None:
         """Remove a specific cache entry from memory and disk."""
+        assert_loop_thread("PromptCacheStore.remove")
         existing = self._entries.pop(cache_id, None)
         if existing is not None:
             self._radix.remove(existing.tokens, cache_id)
@@ -495,6 +501,7 @@ class PromptCacheStore:
         Used during memory pressure to free GPU memory while preserving
         cache state on disk for later restoration.
         """
+        assert_loop_thread("PromptCacheStore.evict_all_to_disk")
         self._evict_all_to_disk_sync()
 
     def _evict_all_to_disk_sync(self) -> None:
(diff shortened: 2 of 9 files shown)


olmlx/engine/prompt_cache/store.py
```diff
@@ -682,6 +692,7 @@ async def async_evict_all_to_disk(self) -> None:
 
     def clear(self) -> None:
         """Remove all cache entries and disk cache files."""
+        assert_loop_thread("PromptCacheStore.clear")
```
dpalmqvist: **Claude Code review (high effort): confirmed bug.** `PromptCacheStore.clear()`'s only production caller is `_close_loaded_model` (model_manager.py:801), which runs on a **worker thread** via `asyncio.to_thread` on every unload / LRU eviction / keep-alive expiry (model_manager.py:965, 1646, 3141). With the loop bound (any real server), this assert fires there; the per-step `except Exception` absorbs it, so the symptom is an error log on every unload plus the cache (incl. on-disk entries) never actually cleared and `prompt_cache_store` never nulled. Same for `VlmPromptCacheStore.clear()` (vlm_state.py). The close path is safe without the assert — the LoadedModel is already popped from `_loaded`, so the closer owns the store exclusively. Fix: drop the assert from both `clear()` methods (documenting the exclusive-ownership contract) and add a regression test that drives `_close_loaded_model` from a worker thread with the loop bound.

olmlx/engine/prompt_cache/store.py
```diff
@@ -681,7 +691,14 @@ async def async_evict_all_to_disk(self) -> None:
             await asyncio.to_thread(self._save_entries_to_disk, snapshot)
 
     def clear(self) -> None:
-        """Remove all cache entries and disk cache files."""
+        """Remove all cache entries and disk cache files.
+
+        Deliberately NOT loop-affine (#463): the only production caller is
+        ``_close_loaded_model``, which runs on a worker thread via
+        ``asyncio.to_thread`` during unload/eviction/expiry — by then the
```
dpalmqvist, line 698: Addressed in 1b75390: the docstring now states explicitly that `clear()` does blocking disk I/O (`shutil.rmtree`) and that loop-side cleanup goes through `remove` / `async_evict_all_to_disk`. Skipped the runtime "warn if on loop" check — that would invert the affinity contract used everywhere else in this PR (loop = allowed) for one method, and the assert layer is for catching corruption-class violations, not blocking-I/O hygiene.

olmlx/engine/prompt_cache/vlm_state.py
```diff
@@ -41,7 +43,12 @@ def enabled(self) -> bool:
 
     def clear(self) -> None:
         """Drop retained states. Counters are cumulative and NOT reset here, so
-        /api/ps reuse totals survive memory-pressure flushes."""
+        /api/ps reuse totals survive memory-pressure flushes.
+
+        Deliberately NOT loop-affine (#463): like PromptCacheStore.clear, the
+        production caller is the worker-thread close path in
+        ``_close_loaded_model``, which owns the store exclusively by then.
```
dpalmqvist, line 50: Declining a runtime guard here: `VlmPromptCacheStore.clear()` is a plain dict clear — no I/O and no partial state, so a hypothetical loop-side caller is safe by construction (the loop is the *more* privileged context everywhere else in this contract). The documented exclusive-ownership note covers why the worker-thread close path is also legal. Adding a thread check that allows both contexts would assert nothing.

olmlx/engine/speculative.py
```diff
@@ -115,6 +117,29 @@ def __init__(self, capacity: int) -> None:
         # LRU order: index 0 is the least-recently-used entry (evicted first),
         # the last index is most-recently-used.
         self._entries: list[_SpecCacheEntry] = []
+        # Guard backing _serialized() — never blocked on, only try-acquired.
+        self._access_guard = threading.Lock()
+
+    @contextmanager
+    def _serialized(self) -> Iterator[None]:
+        """Enforce the no-concurrent-access contract explicitly (#463).
```
dpalmqvist, line 125: Deliberately keeping the plain `Lock` (now documented in 1b75390): same-thread reentry (`find → ... → find/clear` mutating the LRU list mid-scan) is exactly as unsafe as cross-thread overlap, so it should raise too — an `RLock` would silently allow it and weaken the guard. No reentrant path exists today (find/insert/clear call nothing that can re-enter the store), so a spurious raise can't occur; the error message now names the reentry case so a future hit is diagnosed correctly.

olmlx/engine/prompt_cache/store.py
```diff
@@ -681,7 +691,18 @@ async def async_evict_all_to_disk(self) -> None:
             await asyncio.to_thread(self._save_entries_to_disk, snapshot)
 
     def clear(self) -> None:
-        """Remove all cache entries and disk cache files."""
+        """Remove all cache entries and disk cache files.
+
```
dpalmqvist, line 695: `clear()` has exactly one production caller — `_close_loaded_model` on the unload worker thread (verified by grep; the memory-pressure flush uses `async_evict_all_to_disk`). The docstring added in 1b75390 explicitly directs loop-side cleanup to `remove` / `async_evict_all_to_disk`. There is no loop-side caller to protect against today, and the prior revision's loop-affinity assert on this method is what broke the unload path in the first place.

olmlx/engine/speculative.py
```diff
@@ -115,6 +117,33 @@ def __init__(self, capacity: int) -> None:
...
+
+        Unlike the loop-bound prompt-cache stores, this store legitimately
+        runs on the decode worker thread — its safety comes from the
+        inference lock serializing all access, not from loop affinity.  A
+        non-blocking acquire turns an overlapping access (a refactor that
+        races the decode thread against another caller) into an immediate
+        RuntimeError instead of silent LRU-list corruption.
+        """
+        # Deliberately a plain Lock, not RLock: same-thread reentry (a future
+        # find → ... → find/clear chain mutating mid-scan) is just as unsafe
```
dpalmqvist, line 135: `find()` acquires `_serialized()` exactly once and calls nothing that can re-enter the store (`_common_prefix_len` and list ops only) — there is no "otherwise valid same-thread path" that hits the guard today. If a future refactor introduces one, raising is the intended behavior: reentry that mutates the LRU list mid-scan is as unsafe as cross-thread overlap, which is why the guard is deliberately a plain Lock (documented at the acquire site).

olmlx/app.py
```diff
@@ -42,6 +43,21 @@
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
+    # Bind the event-loop thread so the lock-free mutators (registry RMW,
+    # unload's check-then-pop, prompt-cache stores) can enforce their
+    # loop-affinity contract (issue #463).  Unbound in the finally so a
+    # failed startup or repeated create_app() in tests leaves no stale
+    # binding behind.
+    loop_affinity.bind_loop_thread()
```
dpalmqvist, line 51: Startup code inside lifespan runs on the event-loop thread, so it satisfies the bound contract by construction — binding first costs nothing there. If future startup work spawns a worker thread that calls a mutator, the assert firing is exactly the protection this PR adds, not an accident. Binding late would re-open the unprotected window the binding exists to close.

PR #489, opened 2026-06-10
fix(engine): enforce the event-loop-affinity contract with mutation tripwires (#463)

### LINKED ISSUE #463: Make the event-loop-affinity contract explicit (registry RMW, unload check-then-pop, cache stores)

dpalmqvist:
## Summary

Closes #463.

The "all mutations happen on the asyncio event-loop thread" invariant behind the registry's `models.json` read-modify-write, `unload()`'s `active_refs` check-then-pop, and the prompt-cache store's in-memory paths was documented in comments but never enforced — a future refactor (moving a caller into a thread, adding an `await` inside a critical section) would turn into silent state corruption rather than an error.

This adds `olmlx/utils/affinity.py` with two cheap, always-on tripwires and wires them into the mutating entry points:

- **`ThreadAffinityGuard`** — pins the owning thread on the first mutation and raises `RuntimeError` on any later cross-thread call. Wired into:
  - `ModelRegistry.add_alias` / `add_mapping` / `remove` (the models.json RMW)
  - `ModelManager.ensure_loaded` / `unload` / `_expire_stale` (the `_loaded` check-then-pop sites)
  - `PromptCacheStore.get` / `set` / `remove` / `evict_all_to_disk` / `fetch_nearest` / `takeover` / `async_get` / `async_set` / `async_evict_all_to_disk` (everything that mutates `_entries`/`_radix`, including LRU promotion on reads)
- **`SerializedMutationGuard`** — non-blocking lock that raises on *overlapping* entry, wired into `_SpecCacheStore.find` / `insert` / `clear`.

## Design notes (deviations from the issue's literal sketch)

- **Pin-on-first-mutation instead of `asyncio.get_running_loop()`**: registry mutators also run in sync CLI contexts with no loop at all (`olmlx models …`, plus the whole sync unit-test suite), so a literal off-loop assert would false-positive there. Pinning the first mutating thread enforces the invariant that actually matters (no cross-thread mutation) with zero configuration and no false positives, so it can stay on unconditionally rather than behind a `debug_checks` setting.
- **`_SpecCacheStore` gets an overlap guard, not a thread pin**: its mutations run on a different decode worker thread each request (`CancellableStream` worker), serialized by the inference lock — there is no single owning thread to pin. The contract there is "never concurrent", which the non-blocking-lock guard enforces directly (and a regression test asserts thread roaming stays allowed).
- **`PromptCacheStore.clear()` is deliberately exempt**: `_close_loaded_model` calls it from an `asyncio.to_thread` worker during model teardown, after the `LoadedModel` is popped from `_loaded` — exclusive access is guaranteed there without loop affinity. Covered by a test documenting the exemption.
- The registry's `_save_mappings_locked` comment now points at the enforced guard. The issue's alternative (widening `_save_lock` to cover mutation+save) was not taken — the tripwire makes any second mutating thread an immediate error, so the wider lock would protect a state the guard forbids.

## Testing

- New `tests/test_thread_affinity.py` (19 tests): guard unit tests plus per-site wiring tests (cross-thread mutation raises for registry/manager/store; overlap raises and thread-roaming passes for the spec store; teardown `clear()` exemption).
- Full suite: 4585 passed, 4 skipped.
- `ruff check` + `ruff format` clean.


dpalmqvist:
Evaluated the aider-code-review findings — no code changes; reasoning for the record (the bot comment gets overwritten on the next push):

1. **`_SpecCacheStore.clear()` raising on cross-thread teardown** — false premise: `SerializedMutationGuard` does not pin a thread; it only forbids *overlapping* entry. Cross-thread sequential use is explicitly allowed and covered by `test_sequential_cross_thread_use_passes`. An overlap between `close()`→`clear()` and an in-flight `find()`/`insert()` would be a genuine inference-lock violation, and raising there is the designed behavior.
2. **`find()` read-modify-write race on future refactor** — the whole method body is inside the `with self._guard:` block; the critical section is the block itself. No split exists to protect against today.
3. **`get()` unsafe from background threads** — verified all production callers: every guarded sync mutator (`get`/`set`/`remove`/`fetch_nearest`/`takeover`/`insert_checkpoint`) is invoked from coroutine code on the loop. The async disk path's worker thread runs `_read_from_disk`, which never touches `_entries`. `clear()` — the one teardown-path method actually called from `asyncio.to_thread` (`_close_loaded_model`) — is already exempt, with a test documenting why.
4. / 5. **`ensure_loaded`/`unload` checks stronger than the code relies on** — the strictness is the point of #463: convert a future off-loop caller into an immediate error. "Serializes through `_lock`" is not an available alternative — `_lock` is an `asyncio.Lock` (loop-bound), and `unload`'s `active_refs > 0` check-then-pop is not covered by `_lock` at all; it is atomic only by loop affinity, which is what the guard enforces.
6. **`add_alias` read side unguarded** — the contract enforced is single-*mutator*-thread; a second thread mutating mid-`add_alias` trips at its own entry point. Concurrent unguarded *reads* are pre-existing and out of scope for #463.
7. **`_save_mappings_locked` comment** — the comment claims exactly what is enforced (single mutator thread via the guard) and retains the original caveat that the lock only serializes disk I/O.
8. **First-pin race** — acknowledged and deliberate, documented in the `ThreadAffinityGuard` docstring: it is a tripwire, not a lock; a racy first pin still leaves one of the racing threads to trip it on its next call.

olmlx/engine/prompt_cache/store.py
```diff
@@ -446,6 +452,7 @@ def set(self, cache_id: str, state: CachedPromptState) -> CachedPromptState | No
         cleanup (different .cache object), or None when no cleanup is needed.
         Evicted entries are saved to disk if disk offload is enabled.
         """
+        self._mutation_guard.check()
```
dpalmqvist, line 455: **[claude-code-review]** `_set_in_memory` (above) is the deepest insertion chokepoint for `_entries`/`_radix` but carries no guard check itself — every *current* caller (`set`, `async_get`, `async_set`, `takeover`) checks at its entry point, so a future method calling `_set_in_memory` directly would bypass the tripwire entirely. `ThreadAffinityGuard.check()` is an idempotent same-thread no-op, so adding it inside the helper closes the gap at zero cost.

dpalmqvist:
Superseded by #488, which merged a lifespan-bound `loop_affinity` implementation covering the same sites (plus `VlmPromptCacheStore`) with the same `_SpecCacheStore` overlap guard and `clear()` teardown exemption.

PR #495, opened 2026-06-11
feat(dflash): --self-generate training data (upstream recipe) + benchmark validation

dpalmqvist:
## Summary

Root-causes and fixes the historically terrible DFlash acceptance rate, and validates the decoder end to end against upstream.

**The decoder was never the problem.** Verified two ways:
- Exact parity with upstream z-lab's reference MLX loop: τ matched within 2% on identical prompts/targets (theirs 3.06/1.98 vs ours 3.12/1.97).
- Greedy speculative output ≡ plain greedy baseline except at argmax near-ties (target top-2 logit margin ≤ 0.25 at every divergence — quantized-kernel numerics, inherent to speculative decoding).

**The actual bug was the training recipe**: `olmlx dflash prepare` trains on ground-truth dataset text, which measurably produces ~0.4%-acceptance drafts. The upstream recipe (arxiv:2602.06036) trains on responses *generated by the target model* "for better target alignment".

## Changes

- **`engine/dflash/selfgen.py`** (new): target-greedy training-data generation — batched with exact-length prompt grouping, epoch-cycling `(input_ids, pivot_lo)` batches, and `safe_rope_patch()`: an exact fold-batch-into-heads workaround for an **mlx 0.31.x bug where `mx.fast.rope` corrupts batch rows ≥ 1 at the B>1, L==1 decode shape** (every row but 0 silently degenerates into repetition loops). The patch is scoped to the generation loop only.
- **`prepare.py`**: `self_generate` wiring + `min_pivot` plumbing so training windows stay out of the unpredictable prompt region (`pivot_lo` = per-batch max prompt length). Legacy pivot ranges are preserved bit-exactly at `min_pivot=0`.
- **`cli.py`**: `--self-generate` / `--selfgen-seqs` / `--selfgen-max-new`.
- **`scripts/dflash_acceptance_probe.py`**, **`scripts/dflash_diagnose.py`**: acceptance + exactness (with automatic numerics-tie classification at divergences) + tok/s vs greedy; per-position draft accuracy profiling.
- **CLAUDE.md**: corrected the stale "~2% acceptance (feature-space drift + MASK convention mismatch)" note with the verified root cause and benchmark numbers.

## Validation

- Recipe effect on Qwen3-0.6B-4bit (same budget): ground-truth data → 0.4% acceptance, flat per-position profile; self-generated data → 45% position-1 accuracy with the expected decaying profile.
- Benchmark on Qwen3-8B with the upstream pretrained draft (`z-lab/Qwen3-8B-DFlash-b16`), 250 tokens, block 16, `enable_thinking=False`:

| Domain | τ (tok/step) | bf16 target | 4-bit target |
|---|---|---|---|
| JSON extraction | 9.6–10.4 | **4.00×** vs greedy | **1.67×** |
| Math step-by-step | ~6.0 | **2.31×** | 1.06× |
| Code generation | 4.5–5.5 | **1.75×** | 0.96× |
| Thinking-mode chat | ~2 | 0.88× | 0.36× |

4-bit targets cap out lower because the (block+1)-token verify forward is compute-bound on quantized weights (~4.3× a single decode step on M4 Pro), so break-even needs τ ≳ 6.

## Tests

16 new tests in `tests/test_dflash_selfgen.py` (rope-patch exactness/restoration, generation grouping + EOS handling, batch cycling, `min_pivot` bounds + legacy bit-exactness, prepare integration incl. end-to-end self-generate with fake loaders). Full dflash suite + CLI coverage: 168 passed; ruff clean.


olmlx/engine/dflash/selfgen.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    usable = [(s, plen) for s, plen in sequences if len(s) >= min_len]
+    if not usable:
+        raise ValueError(
+            f"No self-generated sequences of at least {min_len} tokens; "
+            "increase --selfgen-max-new or check the prompt stream."
+        )
+    usable.sort(key=lambda sp: len(sp[0]))
+    buckets: list[tuple[list[list[int]], int]] = []
+    for i in range(0, len(usable) - batch_size + 1, batch_size):
```
dpalmqvist: **[code-review]** The bucket loop `range(0, len(usable) - batch_size + 1, batch_size)` silently drops the tail partial batch — up to `batch_size - 1` usable sequences never enter training when `len(usable) % batch_size != 0`. The replication fallback below only runs when there are *zero* full buckets, so it doesn't cover this. Building the tail bucket with replication-fill (same trick as the fallback) fixes both and lets the duplicated fallback branch be deleted.

olmlx/engine/dflash/prepare.py
```diff
@@ -488,6 +488,7 @@ def _select_pivot(
     input_ids: mx.array,
     pad_token_id: int,
     block_size: int,
+    min_pivot: int = 0,
```
dpalmqvist, line 491: **[code-review]** Stale contract: with `min_pivot`, the docstring's None-condition below ("no row has at least 2*block_size + 1 real tokens") is no longer the whole story — the function now also returns None when `min_real < min_pivot + block_size + 1`. The skip-log, abort message, and final under-training warning in `prepare_dflash_draft` similarly cite `2*block_size+1` without mentioning `pivot_lo`, which would mislead an operator diagnosing skipped self-generate batches.

olmlx/cli.py
```diff
@@ -2816,6 +2816,11 @@ def cmd_dflash_prepare(args):
         print(f"  Train windows per step: {args.train_windows_per_step}")
     if args.use_precomputed:
         print(f"  Precomputed shards: {args.use_precomputed}")
+    if args.self_generate:
```
dpalmqvist, line 2829: **[code-review]** `--self-generate --use-precomputed` is only rejected by the function-level guard inside `prepare_dflash_draft`, which runs *after* `store.ensure_downloaded()` — a user with both flags pays a multi-GB target download before seeing the error. The other flag validations in `cmd_dflash_prepare` deliberately run pre-download; add the mutual-exclusion check there.

olmlx/engine/dflash/selfgen.py
```diff
@@ -0,0 +1,273 @@
...
+    dropped — otherwise up to ``batch_size - 1`` generated sequences
+    would silently never be trained on. ``pivot_lo`` is the max prompt
+    length within the batch so the trainer can restrict windows to the
+    response region. The iterator is infinite — the training loop's
+    ``steps`` budget is the termination condition, which is how
+    multi-epoch training falls out naturally.
+    """
+    usable = [(s, plen) for s, plen in sequences if len(s) >= min_len]
+    if not usable:
+        raise ValueError(
```
dpalmqvist, line 268: Fixed in 60a6ac0 — `cycle_training_batches` now takes `block_size` directly and filters each sequence on both floors: `len >= 2*block_size + 1` and the prompt-aware `len >= prompt_len + block_size + 1`. Covered by `test_filters_long_prompt_short_response_sequences`. The skip/abort diagnostics were also updated in e7e9a1e to mention the pivot_lo requirement.

olmlx/engine/dflash/prepare.py
```diff
@@ -488,15 +488,20 @@ def _select_pivot(
...
 ) -> int | None:
     """Pick a pivot inside the right-padded prefix shared by every batch row.
 
-    Returns ``None`` when no row has at least ``2 * block_size + 1`` real
-    tokens in its prefix — the caller should skip the batch in that case
-    rather than forcing a degenerate pivot. The pivot range is
-    ``[block_size, min_real_len - block_size - 1]`` (inclusive) so that
-    every row has both a real ``pending`` token at position ``p`` and
-    real targets at positions ``p+1..p+block_size``.
+    Returns ``None`` when the pivot range is empty — the caller should
```
dpalmqvist, line 495: Fixed in 60a6ac0 — added a `min_pivot` paragraph to `_select_pivots`'s docstring mirroring `_select_pivot`'s. The trailing-pad-detection duplication between the two functions predates this PR and is deliberate (the in-code comment explains the K=1 RNG-equivalence guarantee that factoring it out would complicate); leaving that as is.

olmlx/engine/dflash/prepare.py
```diff
@@ -951,6 +997,50 @@ def _step(
...
+            tokenizer,
+            prompts=prompts,
+            num_seqs=selfgen_num_seqs,
+            max_new=selfgen_max_new,
+            group_size=selfgen_group_size,
+            progress_callback=progress_callback,
+        )
+        # ``pad_for_pivot`` is the same pad id the cycling batcher uses,
+        # so trailing-pad pivot detection stays consistent. Falls back
+        # to 0 only when the tokenizer exposes neither pad nor eos (the
```
dpalmqvist: Fixed in HEAD — `pad_for_pivot` is now set to the batcher's actual pad id (`sequences_pad`) right after the selfgen block, so the pad-aware `_select_pivots` path always runs for self-generated batches. Covered by `test_self_generate_no_pad_tokenizer_uses_padded_pivot_path`.

olmlx/engine/dflash/selfgen.py
```diff
@@ -0,0 +1,291 @@
...
+      (response-region restriction) applies; worse, it drags the whole
+      bucket's ``pivot_lo`` up so ``_select_pivots`` silently skips the
+      batch for every row in it.
+
+    The survivors are sorted by length and packed into fixed same-size
+    buckets (padding waste stays minimal and every row in a batch shares
+    a similar pivot range). A partial tail bucket is topped up by
+    replicating its own rows (same convention as
+    ``stream_training_batches``) rather than dropped — otherwise up to
+    ``batch_size - 1`` generated sequences would silently never be
```
dpalmqvist, line 255: Fixed in de35130 — the bucket's pivot floor is now clamped to `shortest_row_len - block_size - 1`, so the pivot range can never be empty even when sorted-by-length bucketing pairs a long-prompt row with a shorter one across a length-distribution gap. The trade (a few windows may overlap the long-prompt row's prompt region, i.e. that row's pre-pivot_lo behavior) beats silently skipping the whole batch. Covered by `test_pivot_lo_clamped_to_keep_pivot_range_nonempty`. Sorting by plen first was considered but would significantly worsen padding waste, which the length sort exists to minimize.

olmlx/engine/dflash/selfgen.py
```diff
@@ -0,0 +1,291 @@
...
+            for b, t in enumerate(cur[:, 0].tolist()):
+                if not done[b]:
+                    gens[b].append(int(t))
+                    if int(t) in eos_ids:
+                        done[b] = True
+            if all(done):
+                break
+            logits = target(cur, cache=cache)
+            logits = getattr(logits, "logits", logits)
+            cur = mx.argmax(logits, axis=-1)
```
dpalmqvist, line 193: Acknowledged but deliberately not implementing: compacting done rows out of a live batched KV cache in mlx means per-layer cache surgery (slicing keys/values/offset across all layers and rebuilding the batch), which is substantial complexity for a one-time offline data-generation pass. Empirically the waste is small in the recipe's sweet spot: thinking-mode responses almost never emit EOS before --selfgen-max-new, so rows in a group finish together (the `all(done)` early-exit covers the remaining case). Happy to revisit if generation cost ever dominates prepare time.

PR #496, opened 2026-06-11
feat(engine): shard KV cache quantizer — Tier 1 (#377)

### LINKED ISSUE #377: KV cache compression: investigate Shard (asymmetric K/V, PCA-K + VQ-V + sink/window)
Investigate adding a `shard`-style KV cache quantizer alongside the existing `turboquant` / `spectralquant` options.

Upstream: https://github.com/krish1905/shard

## What Shard does

Drop-in KV-cache compression for Llama-style models. Asymmetric K/V treatment:

- **K (prefill)**: undo RoPE → per-head PCA (claim: K is effectively rank-192/1024 in its no-RoPE basis) → store int4 PCA coefficients. Attention computes Q·K directly on the int4 form via a "relative-Δ RoPE identity" that pushes the per-token rotation onto Q, so K is never reconstructed.
- **V (prefill)**: Hadamard rotation + k-means vector quantization across heads.
- **Decode buffer**: 8-bit per-token Lloyd-Max (TurboQuant-style, data-oblivious), bit-exact lossless over 750 tokens.
- **Sink+window**: first 4 + last 64 tokens kept in FP16; only the middle is compressed.

Headline numbers (Llama-3.1-8B-Instruct, B200): ~10× memory at 8K context, ~11.2× at 32K, ~FP16 quality, throughput 0.4–0.5× FP16 (memory-capacity win, not latency).

Stack upstream: PyTorch + Transformers + Triton int4 GEMM kernels.

## Mapping to olmlx

We already own a meaningful chunk of the moving parts:

| Shard component | olmlx counterpart |
|---|---|
| Lloyd-Max streaming decode quantizer | `engine/turboquant.py` — same algorithm, per-coordinate Gaussian Lloyd-Max codebook, bit-packed indices |
| PCA-based K basis | `engine/spectralquant_calibrate.py` — per-head `mx.linalg.eigh`, separates semantic vs. tail subspaces; non-uniform bit allocation already an axis we control |
| Drop-in `KVCache` swap | Pattern established by `turboquant_cache.py`, `spectralquant_cache.py` |
| Per-coordinate non-uniform bits | SpectralQuant already does this |

## What's new and would need building

1. **Asymmetric K vs V treatment.** Today TurboQuant/SpectralQuant treat both the same. New `ShardKVCache` class with different paths per side.
2. **Hadamard + k-means VQ for V.** Hadamard butterfly is small. K-means fitting runs offline like the existing Lloyd-Max fitting — sklearn during calibration, save centroids to safetensors. Runtime lookup is an index-gather. New code, not blocked by MLX.
3. **Sink + window hybrid.** Few lines on top of any existing cache.
4. **The hard part: Q·K on int4 PCA coefficients with RoPE pushed to Q.** This is where Shard's "no FP16 K reconstruction" benefit lives. Two sub-issues on MLX:
    - The relative-Δ RoPE identity that folds per-pair position rotation into Q is doable but invasive — it intercepts the attention forward, not just the cache. Our existing quantized caches dequantize-on-read and stay outside attention, giving memory savings but losing the int4-compute win.
    - The fast path upstream is a Triton int4 GEMM with on-the-fly RoPE. There is no Triton on Apple Silicon. Closest MLX equivalent is `mx.quantized_matmul` for grouped int4 plus a `mx.fast.metal_kernel` for the RoPE-fused variant. The Metal kernel is the dominant chunk of work.
5. **Per-target calibration.** Rank-192/1024 is a Llama-3.1-8B-specific number. For Qwen3 / Qwen3-Coder we'd need to discover the rank in the no-RoPE basis as part of the existing `olmlx spectral prepare` pipeline.

## Proposed two-tier plan

**Tier 1 (cheap, ~80% of the memory benefit)**
A `shard:K4_V4` mode that does PCA-K + VQ-V + sink/window, dequantizing on read. Reuses the spectralquant calibration pipeline almost verbatim; adds a Hadamard+k-means side path for V. Config: `OLMLX_KV_CACHE_QUANT=shard:4`. Calibration: `olmlx shard prepare <model>`.

**Tier 2 (the headline)**
int4-compressed Q·K via a Metal kernel with RoPE folded onto Q. Justifies the "10×" framing and is genuinely new infrastructure. Validate the math on a CPU/PyTorch reference before writing Metal.

## Caveats

- Shard's own throughput is 0.4–0.5× FP16 even with the Triton kernel. Without the kernel, Tier 1 in olmlx will be slower than that. On Apple Silicon with unified memory the value is "fit a bigger model in the same RAM" — a memory-capacity story, not a speed story, but aligned with how Flash is already positioned in olmlx.
- The relative-Δ RoPE identity assumes standard RoPE. Qwen3.5 / Qwen3-Next gated-delta linear-attention layers don't use RoPE there, so Shard would only apply to the full-attention layers (same caveat that already hits our quantized caches on hybrid models — see #284, #343).
- Disk-cache offload would need the same guard `TurboQuantKVCache` already has in `_save_to_disk`.

## Suggested files (Tier 1)

- `engine/shardquant.py` — Hadamard, VQ codebook fit/apply, PCA-K compress/decompress
- `engine/shardquant_cache.py` — drop-in `KVCache` with asymmetric K/V paths + sink/window
- `engine/shardquant_calibrate.py` — calibration entry point reusing the spectral calibration scaffolding
- CLI: extend `olmlx` with `shard prepare` subcommand
- Config: validate `OLMLX_KV_CACHE_QUANT=shard:{2,4,8}` in `config.py`

dpalmqvist:
Implements Tier 1 of #377: a `shard:{2,4,8}` KV-cache quantization mode based on the [Shard reference design](https://github.com/krish1905/shard), incorporating the amendments from the review comment on the issue.

## What's implemented

**`olmlx/engine/shardquant.py`** — primitives:
- `RopeSpec` / `detect_rope_spec` / `rope_transform`: invertible RoPE. Detects `mlx.nn.RoPE` and mlx-lm's `_freqs`-carrying custom ropes (Llama3RoPE-style); unknown variants fall back to identity (still exactness-preserving through the undo/redo pair — just less rank collapse). Parity with `nn.RoPE` is test-enforced.
- `make_v_rotation`: Sylvester-Hadamard when `head_dim` is a power of 2, seeded orthogonal QR otherwise — persisted in the calibration artifacts so the runtime has one uniform path.
- Product VQ for V (256-centroid per-position codebooks, numpy k-means at calibration, group `8//bits` → exactly `bits` bits/dim) and per-head PCA + rank truncation + Lloyd-Max scalar quant for K.

**`olmlx/engine/shardquant_cache.py`** — `ShardKVCache`:
- `[ sink (4, FP16) | middle (compressed) | window (64, FP16) ]`. K middle stored *de-roped* in per-head PCA bases; RoPE re-applied at fetch from the stored absolute positions.
- **Transient dequantize-on-read** (the strategy choice the review comment asked to make explicit): no persistent dequant side buffer, so resident memory is quantized + sink/window. The cost is full-middle dequant + re-rope per layer per decode step — this mode is a memory-capacity win, not a speed win.
- **Exact `trim()`** (window → middle → sink; both compressed regions are per-token), so prompt-cache prefix reuse works.
- Deepcopy/snapshot-safe via the default walk: no `mx.Dtype` attributes, and every mutable array is exposed through `state` so the snapshot path's flatten + `mx.eval` materializes in-place-write graphs before cross-thread reuse (#284 family).

**`olmlx/engine/shardquant_calibrate.py`** — `olmlx shard prepare <model>`:
- Per-head eigendecomposition of *de-roped* keys (the review comment's per-layer→per-head correction); per-layer rank = max participation ratio over heads; RoPE frequencies persisted so calibration-time de-rope and runtime re-rope match exactly.
- Reuses the spectral collection scaffolding via new shared helpers (`_load_calibration_model`, `collect_kv_vectors`) extracted from `spectralquant_calibrate.py` — pure refactor, spectral suites pass unchanged.

**Wiring:** `OLMLX_KV_CACHE_QUANT=shard:{2,4,8}` validation in config.py, dispatch in `_make_prompt_cache_for_lm`, `_find_shard_dir` + auto-calibrate in ModelManager (missing/mismatched calibration → HTTP 400 with the exact command; `ShardCalibrationMissingError` subclasses `SpectralCalibrationMissingError` so the existing handler applies), `_is_serializable_cache` blocks disk spill, CLI `shard prepare` subcommand, CLAUDE.md/README docs.

## Intentional v1 simplifications
- Single Lloyd-Max codebook for the kept K coefficients (not per-coordinate); per-layer (not per-head) V codebooks; rank uniform per layer (max over heads).
- No `mx.compile` on the hot path yet — plain MLX ops, correctness first.
- Speculative decoders never see shard caches (they own their caches); distributed/VLM follow the same composition rules as turboquant/spectral.

## Out of scope (Tier 2, per the issue)
int4 Q·K via a Metal kernel with the relative-Δ RoPE identity folded onto Q — the \"no FP16 K reconstruction\" win.

## Testing
- 56 new unit tests across primitives / cache / calibration / wiring (TDD; includes `nn.RoPE` parity, sink+window bit-exactness, trim into the compressed middle, deepcopy divergence, per-head basis independence).
- Live test (`tests/live/test_shard_quant_real.py`, real_model): calibrates Qwen3-0.6B-4bit and verifies coherent generation under `shard:4` — passing locally.
- Full suite: 4623 passed. ruff check + format clean.

Suggested follow-up: an A/B bench vs `spectral:4` (reconstruction cosine + decode tok/s + resident KV bytes), which the review comment identified as the natural acceptance test.

Closes #377 (Tier 1).


dpalmqvist:
## Triage of aider-code-review findings (run 1, captured before overwrite)

Verified each claim against the code; fixed what was real in 4f3a9ff.

**Fixed:**
- `inference.py:1348` (None calibration dir → opaque error): **valid edge** for a store-less manager. `make_shard_cache` now raises `ValueError` with the `olmlx shard prepare` remedy; covered by `test_none_calibration_dir_raises_with_remedy`.

**Verified false, hardened with tests anyway:**
- `shardquant.py:194` (odd rank vs pack width): the spectral packers pad odd tails and `unpack_indices` slices back to the requested dim — that's why they were reused. New `test_odd_rank_roundtrip` (rank=5, 4-bit) proves it.
- `shardquant_cache.py:112` + `:211` (RoPE offsets after trim / stale middle entries): the invariant `sink_len + mid_len + win_len == offset` survives trim, so the window front is always at absolute position `sink_len + mid_len`, and stale packed entries past `_mid_len` are overwritten before any read (same pattern as `TurboQuantKVCache.trim`). New post-trim reconstruction assertions + `test_trim_then_refill_overwrites_stale_middle` prove both.

**False positives (no change):**
- `shardquant.py:84` (`isinstance(x, mx.array)` "never works"): `mx.array` *is* the `mlx.core.array` type in MLX; `test_detects_freqs_carrying_rope` exercises exactly this path with a real `mx.array` `_freqs`.
- `shardquant.py:176` (`c_sq (P,K)` vs `(...,P,K)` broadcast): trailing-axis broadcasting is well-defined for any number of leading dims; the VQ tests run 5-D inputs.

**Acknowledged design choices (documented in the PR body, no change):**
- Pooled per-layer K codebook across heads: v1 simplification; coefficients are unit-normalized before fitting, bounding cross-head scale spread. Per-head codebooks are a natural follow-up alongside the `spectral:4` A/B bench.
- `_is_serializable_cache` excluding `ShardKVCache`: a real limitation, not policy — the packed layout has no `state`/`meta_state` restoration round trip (the `state` setter raises), same as TurboQuant/Spectral.
- Live test model size: `Qwen3-0.6B-4bit` is ~350 MB and the test is `real_model`-marked (skipped in CI).

PR #497, opened 2026-06-11
feat(shardquant): #377 follow-ups — A/B bench, K-quality fixes, compiled hot path

dpalmqvist:
The three follow-ups from #496: the A/B bench vs `spectral:4`, the per-head K codebooks, and `mx.compile` on the hot path. The bench drove the other two — it exposed a K-quality bug worth far more than the planned improvements.

## 1. A/B bench (`scripts/shard_ab_bench.py`)

Measures reconstruction cosine vs the fp16 cache, prefill/decode throughput, and **actual resident cache bytes** (every `mx.array` attribute per layer cache — charging TurboQuant for its persistent dequant side buffers). Calibrations land in the standard model-dir locations and are reused.

Baseline run immediately confirmed two predictions from the #377 review thread:
- TurboQuant's resident cache (186 MB) **exceeds fp16's** (147 MB) at this context length — the persistent-side-buffer trap.
- Shard's K reconstruction was badly broken: **0.638** cosine vs spectral's 0.992.

## 2. K-quality fixes (the real win)

Root causes, found by decomposing the error on real Qwen3-0.6B layers:
- **Missing mean-centering.** The PCA basis comes from a *centered* covariance, but compression projected un-centered unit vectors and truncated — dropping the mean direction, the dominant component of every key. The per-head mean is now persisted (`k_mean`) and subtracted/re-added around projection.
- **Participation-ratio rank collapse.** PR picked ranks 2–15/128 under a dominant eigenvalue. A rank sweep showed truncation (not 4-bit quantization) dominated the error, so rank is now the smallest capturing a cumulative-energy threshold (default 0.999, `olmlx shard prepare --k-energy` to trade K quality for bytes). Per-coordinate whitening was prototyped, measured at +0.007 cosine, and dropped — pooled Lloyd-Max is already MSE-optimal.
- **Per-head K codebooks** `(H, 2^bits)` — the cross-head scale-domination issue from the #496 review.

Tier-1 artifacts still load (no `k_mean` → un-centered path; 1-D codebook → shared); recalibration recommended.

## 3. Compiled hot path

`mx.compile`d cores (the spectralquant lru-cached pattern) for K/V compress/decompress and `rope_transform`. Two trace-stability designs matter:
- RoPE **positions are an array input** — the de-rope offset advances every decode step, and a closed-over int would force a re-trace per step.
- Middle dequant runs over the **full capacity-aligned buffers** and slices to `mid_len` — per-step `+1` shapes would also re-trace; padding norms are zero, ≤255 tokens of overcompute, the slice drops the garbage.

Plus an algorithmic prefill fix: scalar assignment against a *sorted* Lloyd-Max codebook is a midpoint-threshold count, not an `(..., R, K)` float32 abs-argmin tensor — 4× less memory traffic on the dominant prefill compress cost.

## Numbers (Qwen3-0.6B-4bit, 1k prompt, 128 decode, bits=4)

| | K cosine | V cosine | prefill t/s | decode t/s | resident (vs fp16) |
|---|---|---|---|---|---|
| shard @ main | 0.638 | 0.993 | 637 | 50 | 2.9× |
| **shard @ this PR** | **0.980** | 0.993 | **1009** | **52** | 2.4× |
| spectral:4 | 0.992 | 0.995 | 3282 | 46 | 3.7× |
| turboquant:4 | 0.996 | 0.995 | 3789 | 110 | **0.8×** (!) |

Positioning after this PR: shard trades some compression ratio (2.4× vs spectral's 3.7×, mostly the FP16 sink/window + higher kept rank) for better decode throughput and exact recent-token attention; `--k-energy` moves along that curve. Reconstruction is bit-identical between compiled and uncompiled paths.

## Testing
- 12 new unit tests (mean-centering exactness + legacy parity, per-head codebook routing, energy-rank selection, `--k-energy` threading, Tier-1 artifact backward-compat); all 69 shard tests + full suite (4655) pass; live real-model test passes.
- CLAUDE.md updated with the revised algorithm + bench summary.


dpalmqvist:
## Triage of aider-code-review findings (run 1, captured before overwrite)

Verified each against the code; none required a change.

**False positives:**
- `shardquant.py:362` (per-head gather "can select wrong centroid"): after `unpack_indices`, `idx` values are centroid indices in `[0, K)`; the flat index of `(h, idx)` in an `(H, K)` codebook is exactly `h*K + idx`, which is what `offs = arange(H) * n_levels` computes (`n_levels == K == codebook.shape[-1]`). Packed byte width is irrelevant post-unpack. Proven by `test_compress_keys_routes_codebook_per_head` (head 1 given an all-zeros codebook reconstructs to zero while head 0 is unaffected) across bits ∈ {2,4,8}.
- `shardquant.py:459` ("mean added in the wrong order"): the encoder computes `y = (x/‖x‖ − μ) @ Bᵀ`; the decoder computes `x̂ = ((y @ B) + μ)·‖x‖` — mean re-applied in normalized space, *then* rescaled, which is the exact inverse and precisely what the comment says it should be. `test_mean_offset_data_survives_truncation` requires cosine > 0.98 under truncation, which only holds with the correct inverse.
- `shardquant_calibrate.py:282` (codebook "never trained on trailing coefficients"): `fit_codebook(c[:, :k_rank])` slices each head's coefficients to the **shared layer rank**, not the head's own lower rank — the codebook is fit on exactly the coefficient range the runtime quantizes.
- `shardquant.py:341` + `:495` ("recompiles per chunk"): the cores are `lru_cache`d by shape; all full-size chunks share one trace and only the final partial chunk adds a second, so a prefill costs ≤2 traces and decode reuses one stable trace — the same documented trade-off as spectralquant's compiled kernels (`maxsize=128` absorbs varied prefill lengths). The measured effect is in the PR table: prefill 637→1009 t/s.

**Acknowledged design choices (no change):**
- `shardquant.py:346` (`mean=None` vs explicit zeros indistinguishable): intentional — passing zeros keeps the compiled-trace count low, and `x − 0` is bit-exact, so the legacy un-centered contract is preserved behaviorally (`test_mean_none_matches_legacy_behavior`). The docstring documents it.
- `shardquant_calibrate.py:286` (`k_rank = max(ranks)` over-allocates for low-rank heads): deliberate — a uniform per-layer rank keeps the packed storage rectangular; per-head ranks mean ragged tensors. The extra coefficients are valid data quantized against that head's own codebook (fit on the same range, per the point above), so the cost is bytes, not correctness.

dpalmqvist:
Re the deepseek run's dead-code nit: `_codebook_gather` was introduced in 2ab7be5 and already removed in 92be592 when its logic moved inside the compiled decompress core — it does not exist in the final tree (`grep -rn _codebook_gather olmlx/ tests/` is empty). The reviewer appears to have read the intermediate commit rather than the cumulative diff.

PR #498, opened 2026-06-12
feat(shardquant): fused compressed-middle decode — Tier 2 (#377)

dpalmqvist:
## What

Tier 2 of #377: decode-path attention for \`ShardKVCache\` computed **directly from the packed middle region** — the FP16 middle is never materialized to device memory. Tier 1's dominant decode cost was the transient full-middle dequant (unpack → codebook gather → rank back-projection → re-rope, written out and re-read by sdpa every step); Tier 2 replaces it with two slim Metal kernels plus a joint fp32 softmax in MLX.

Design doc: \`docs/superpowers/specs/2026-06-11-shard-kv-quant-tier2-design.md\` · Plan: \`docs/superpowers/plans/2026-06-11-shard-kv-quant-tier2.md\`

## How

- **Handle handoff**: at decode (q_len 1, B 1, mid_len > 0) \`update_and_fetch\` returns a \`ShardFusedKV\` handle carrying the exact sink+window K/V and a cache reference. The handle is deliberately not an \`mx.array\`: an unpatched consumer fails loud instead of silently attending over a partial history.
- **sdpa patch**: \`install_fused_sdpa(model)\` swaps the module-global \`scaled_dot_product_attention\` in each layer-defining mlx-lm module (models bind the symbol at import, so patching \`models.base\` alone does nothing). The wrapper delegates every non-handle call to the original; \`sinks\`/mask/multi-token cases materialize via the Tier-1 path and delegate.
- **K-score kernel**: each simdgroup reconstructs 4 consecutive tokens in registers (unpack {2,4,8}-bit codes → per-head codebook → Σρ y·basis + mean → ×norm → re-rope at sink+j) and dots them with each grouped query. A basis element is loaded once per simdgroup and applied to all 4 staged coefficient vectors — without that amortization the kernel loses to Tier-1's dequant GEMM.
- **V-accumulate kernel**: weight·norm·centroid accumulation in *rotated* space, chunked over tokens for parallelism; the orthonormal un-rotation folds outside the weighted sum (\`Σ wⱼnⱼ(ṽⱼ@R) = (Σ wⱼnⱼṽⱼ)@R\`).
- **Honest divergence from upstream**: Shard's "RoPE folded onto Q" identity exists because a Triton int4 GEMM can't rotate K per key. A custom Metal kernel can — reconstruct-and-re-rope in registers gives identical scores and the same no-FP16-K memory contract, with bit-comparable parity against Tier 1.
- **Config**: \`OLMLX_SHARD_FUSED\` (default **on** — shard is already opt-in via env + calibration, and Tier 2 is strictly a faster evaluation of the same math). Eligibility gate (\`kernels_supported\`: D % 32 == 0, D ≤ 128, rank ≤ 128, non-traditional rope with (dims/2) % 32 == 0) plus a plain-MLX reference path back every ineligible case — correctness never depends on the kernels.
- **No new cache state**: trim/snapshot/deepcopy/radix takeover and the disk-spill gate are untouched; the handle is per-call, never stored.

## Numbers (Qwen3-0.6B-4bit, M-series, \`scripts/shard_ab_bench.py\`)

| mode | decode tok/s @2k prompt | decode tok/s @8k | cache MB @8k |
|---|---|---|---|
| fp16 | 280.9 | 155.7 | 968.9 |
| shard (Tier 1) | 31.1 | 8.1 | 239.7 (4.0×) |
| **shard-fused (Tier 2)** | **34.2** | **11.2** | 239.7 (4.0×) |

Resident bytes identical to Tier 1 (the memory contract); the win grows with context (+10% at 2k, +38% at 8k). The PCA back-projection FLOPs are inherent to PCA-K and remain (mitigate via \`--k-energy\` rank truncation); what Tier 2 removes is the materialization traffic and allocator churn.

## Tests

- \`tests/test_shardquant_fused.py\` (24): ref-vs-Tier-1 parity (bits × rope × GQA), V-fold validation, handle conditions, trim/state invariance, patch idempotency + delegation + loud-failure, 24-step growing-middle decode parity (ref and kernel backends), config knob.
- \`tests/test_shardquant_kernels.py\` (18, GPU-marked): kernel-vs-ref parity over bits {2,4,8} × rope × rank, capacity padding, multi-chunk V partials, eligibility gate.
- \`tests/live/test_shard_quant_real.py\`: fused vs Tier-1 greedy generation on a real model (token-level agreement).
- Full unit suite: 4696 passed.

## Out of scope (v1)

Single-kernel online-softmax flash decode, prefill fusion (q_len > 1), fused \`sinks\` path, B > 1, VLM/distributed/speculative composition (unchanged), per-model \`models.json\` override for the knob.


dpalmqvist:
Verified both aider-code-review findings against the code and empirically — neither is a real bug:

**Basis offset (`h * D * D`)** — the stored `k_basis` is the *full* per-head eigenbasis with shape `(H, D, D)`; rank truncation slices the *coefficients* (`[..., :rank]` at compress, rows `r < rank` in the kernel), never the stored matrix (see `shard_compress_keys`: "basis: (H, D, D), rows = eigenvectors", and `_compiled_k_decompress_core`, which zero-pads `y` to `D` and multiplies by the full matrix). So `h * D * D` is the correct per-head stride. Empirical check: `H=4`, `rank=24 ≪ D=64` — kernel scores match the reference path per-head to ≤ 4.8e-6. A wrong stride would corrupt every head > 0; the parity suite (`tests/test_shardquant_kernels.py`, rank=48 cases) also covers this.

**`window_size=0` assertion crash** — after a spill, `update_and_fetch` sets `self._k_win = self._k_win[..., overflow:, :]`, which is a *zero-length array*, not `None` (only `trim()` nulls it, and the next decode token re-populates it). Moreover `mid_len > 0` implies the sink is full, so every decode token lands in the window before the handle is built. Empirical check: `window=0` fused decode runs fine (handle carries the 4 sink tokens, zero window tokens).

No code change needed for either.

dpalmqvist:
The second-run review re-raised the basis-stride finding — the premise is still incorrect: calibration stores the **full** `(H, D, D)` eigenbasis (rank truncation slices coefficients at compress time, never basis rows — see `shard_compress_keys`/`_compiled_k_decompress_core`, which zero-pads `y` to `D` and multiplies by the full matrix). Re-verified empirically on Metal: `H=4`, `rank=24 ≪ D=64`, per-head kernel scores match the reference to ≤ 4.8e-6; a wrong stride would corrupt every head > 0.

The fair part of the finding is that CI runners skip the Metal parity tests, so a *future* layout change would only surface locally. Hardened structurally in the latest push: `kernels_supported` now declines any cache whose stored basis isn't `(H, head_dim, head_dim)` — a hypothetical truncated-basis artifact falls back to the (CI-covered) reference path instead of striding wrong. Gate test added.

dpalmqvist:
Third-run review findings, verified on merit:

**Score concatenation order (`[exact, middle]` vs sequence order) — not a bug.** Softmax attention is permutation-invariant over the key axis as long as score columns and value rows stay paired, which they do (`w[..., :se]` pairs with `v_exact`, `w[..., se:]` with the middle V; no positional mask exists on this path — `mask is None` is a fusion precondition, and position information is carried by RoPE, applied per absolute token position on both q and reconstructed K). The 24-step decode parity test (`TestEndToEndDecodeParity`) and the live greedy test compare the fused output against Tier-1's `[sink, mid, win]` assembly directly — they agree to 3e-3 fp16 tolerance across a growing middle.

**Missing exact buffers with zero sizes — real but unreachable edge, now hardened.** `window_size=0` was already safe (a spilled window is a zero-length array, not `None`), but `sink_size=0` (possible only via direct construction; `make_shard_cache` keeps the defaults) left `_k_sink` permanently `None`, and the fused branch asserted. Latest push demotes to the Tier-1 `materialize()` path instead of asserting, with a regression test.

Issue #499, opened 2026-06-12
Adapt to mlx >= 0.31.2 thread-local streams (mlx currently pinned <0.31.2)


dpalmqvist:
## Context

`uv lock --upgrade` on 2026-06-12 pulled in mlx 0.31.2, which failed 16 tests with `RuntimeError: There is no Stream(gpu, N) in current thread.` mlx is now pinned `>=0.31.1,<0.31.2` in `pyproject.toml` (with mlx-lm held at 0.31.2, since mlx-lm 0.31.3 requires mlx >= 0.31.2 on Darwin). This issue tracks lifting that pin.

The upstream change is intentional and permanent ([v0.31.2 release notes](https://github.com/ml-explore/mlx/releases/tag/v0.31.2)): thread-local default streams, thread-safe scheduler, "MLX can be used by multiple threads for independent computations." It broke the ecosystem broadly (mlx-lm's own server: ml-explore/mlx-lm#1179, ml-explore/mlx-lm#1181, ml-explore/mlx-lm#1256).

## New semantics (verified empirically on 0.31.2)

| Probe | Result |
|---|---|
| Eval a lazy graph built on another thread (default or explicit stream) | fails |
| Worker "adopts" a foreign stream via `with mx.stream(s)` | fails — no adoption API |
| Use an already-materialized array as input to worker-built ops | works |
| Worker creates and uses its own stream | works |

Rule: **ops must be created and evaluated on the same thread; only materialized arrays may cross threads.** Subtlety: `KVCache.state` returns lazy slices built at property-access time, so touching `.state` on thread A and eval'ing on thread B fails even when the underlying buffers were eagerly eval'd (this is exactly how `test_snapshot_cache_eager_eval_materializes_state` fails).

## Upstream gate (hard blocker)

mlx-lm's `generation_stream` is a module-level stream created at import time on the main thread; every decode in a worker thread dies immediately under 0.31.2. The fix — ml-explore/mlx-lm#1275, converting it to a `threading.local()`-backed factory — is open/unmerged and in no release (0.31.3 still crashes, per ml-explore/mlx-lm#1256). **Do not start the olmlx migration until an mlx-lm release ships that fix** (or decide to vendor a monkey-patch and own it).

## olmlx migration checklist (after upstream lands)

[...]

PR #507, opened 2026-06-12
feat(engine): continuous batching via mlx-lm BatchGenerator — Phases 0-1

dpalmqvist:
## Summary

Eligible streaming chat requests join a per-model `BatchScheduler` (`engine/batching.py`) wrapping mlx-lm's `BatchGenerator` instead of serializing FIFO on the global inference lock. Concurrent decodes share each weight read (bandwidth-bound on Apple Silicon), and a queued request's chunked prefill interleaves with running decodes instead of blocking behind them.

**Measured** (Qwen3-4B-4bit, M-series): 80 tok/s single stream → **120 / 135 / 142 tok/s aggregate at 2 / 4 / 8 concurrent**, with greedy output byte-identical to the exclusive path.

Implements Phases 0–1 of `docs/batching-plan.md` (included in this PR; Phase 2+ = prompt-cache store round-trip, grammar, non-streaming).

## Architecture

- Requests submit into a thread-safe inbox; a manager task acquires the inference lock **once per busy period** (untimed FIFO — per-request queue timeouts 503 consumer-side) and runs one worker thread owning a fresh `BatchGenerator` (lazily created on first admission; `close()` syncs `generation_stream` before lock release, preserving the Metal-exclusivity invariant in the one existing mechanism).
- Fairness latch: when an exclusive request starts waiting on the lock, admission shuts for the rest of the period; running sequences finish, the lock releases FIFO to the waiter, the manager re-queues behind it.
- `_stream_completion_batched` yields the exact chunk dicts of the exclusive path — routers untouched. Per-sequence streaming detokenizers, consumer-side stop sequences/timeouts, `_inference_ref` held until the worker confirms removal (eviction can't race a forward pass).
- **v1 eligibility** (`_batch_eligible`, opt-in via `OLMLX_BATCHING`): text-only dense models on plain `KVCache`. VLM/whisper/TTS/reranker, speculative, distributed, flash(-MoE), KV-quant, grammar, per-request seed, gpt-oss channel format stay exclusive. Batched requests fresh-prefill in v1.

## ropefix

The mlx 0.31.x `mx.fast.rope` B>1/L==1 corruption is **scalar-offset only** — `BatchKVCache`'s per-row vector-offset decode path is correct (verified empirically). `safe_rope_patch` moves from `dflash/selfgen` to `engine/ropefix.py` (re-exported), folds only the buggy shape, and is held around the worker loop as defense in depth. Removal gates for #499: a Metal-gated bug-presence unit test + the live parity canary.

## Bugs found by the full suite (fixed, with regression tests)

1. **Mock vacuity**: a MagicMock model's `make_cache()` is truthy but iterates empty → the plain-KV probe passed vacuously, and tests patching `inference.settings` (truthy `.batching`) routed mocked requests onto the batched path. Probes now materialize the list; the predicate gates on `settings.batching is True`.
2. **Loop rebinding**: the scheduler lives on `LoadedModel` and could outlive the event loop owning its manager task (pytest-asyncio: one loop per test) — a submit from a new loop queued into a manager that could never run, hanging indefinitely. Now rebinds when the bound loop is closed; rejects cross-loop use while it is alive.

## Test plan

- [x] Full unit suite: 4748 passed, 0 hangs (`-o faulthandler_timeout=240`)
- [x] `tests/test_batching.py` (51): scheduler behavior on a fake generator (busy-period sharing, cancel, worker error fan-out, fairness latch, loop rebind, close), cache probes, eligibility predicate, settings, ropefix incl. removal-gate test
- [x] `tests/live/test_batching_real.py` (4, real_model): batched-vs-single greedy parity, mixed-length coherence, 3-way concurrent engine-level parity, wall-clock throughput gate
- [x] pyright: no new errors/warnings vs baseline


olmlx/engine/inference.py
```diff
@@ -2548,6 +2548,343 @@ async def _store_prompt_cache_after_generation(
...
+    elif isinstance(prompt, str):
+        tokens = tokenize_for_cache(lm.text_tokenizer, prompt)
+    else:
+        tokens = list(prompt)
+
+    scheduler = _get_batch_scheduler(lm)
+    seq = await scheduler.submit(
+        BatchRequest(
+            tokens=tokens,
+            max_tokens=max_tokens,
```
dpalmqvist: **[review:correctness] Ollama `num_predict: -1/-2` (unlimited) breaks on the batched path.** `max_tokens` is forwarded raw to `BatchGenerator`, whose `GenerationBatch` checks `self._num_tokens[i] >= self.max_tokens[i]` — `1 >= -1` is true after the first token, so an unlimited request returns exactly one token with finish_reason "length". The exclusive path's `stream_generate` uses `n == max_tokens`, treating -1/-2 as infinite. Fix: route `max_tokens <= 0` to the exclusive path (add it to `_batch_eligible`) or normalize before submit.

olmlx/engine/batching.py
```diff
@@ -0,0 +1,372 @@
...
+            try:
+                await self._acquire_gpu()
+            except BaseException as exc:  # noqa: BLE001 — fail queued requests
+                self._fail_inbox(exc)
+                if isinstance(exc, asyncio.CancelledError):
+                    raise
+                logger.exception("batch[%s]: GPU acquisition failed", self._name)
+                continue
+            try:
+                await asyncio.to_thread(self._worker)
```
dpalmqvist: **[review:correctness] Manager-task cancellation releases the inference lock while the worker thread is still running GPU work.** `await asyncio.to_thread(self._worker)` raises CancelledError on task cancellation (app shutdown), but the thread cannot be cancelled and keeps ticking `gen.next()`; the `finally: self._release_gpu()` then syncs and releases the lock concurrently with live Metal work — the #284-class hazard the 'Metal work finished before release' invariant exists to prevent. Fix: shield the to_thread future; on CancelledError set `_closing` (worker drains at next tick) and await the future before releasing, then re-raise.

olmlx/engine/inference.py
```diff
@@ -2567,6 +2904,24 @@ async def _stream_completion(
     tokenizer: Any = None,
     template_kwargs: dict | None = None,
 ) -> AsyncGenerator[dict, None]:
+    # Continuous batching (docs/batching-plan.md): eligible requests join
+    # the per-model batch instead of serializing on the inference lock.
+    # Phase 1 batched scope skips prompt-cache reuse (fresh prefill).
+    if _batch_eligible(
```
dpalmqvist, line 2971: **[review:correctness] Batched requests skip `_kv_cache_preflight_check` entirely — no memory admission, so oversized prompts produce an uncatchable Metal OOM (process death) instead of the exclusive path's 503.** With `batch_completion_size=8`, aggregate KV is unbounded. Minimum v1 fix: per-request KV estimate + 503 before submit; aggregate admission can stay Phase 3 (plan §10) but per-request protection shouldn't.

olmlx/engine/inference.py
```diff
@@ -2548,6 +2548,343 @@ async def _store_prompt_cache_after_generation(
...
+            completion_batch_size=completion_size,
+            prefill_batch_size=prefill_size,
+            prefill_step_size=prefill_step,
+        )
+
+    async def acquire_gpu() -> None:
+        await _await_deferred_cleanup()
+        # 0 disables the acquire timeout: the manager waits FIFO behind
+        # exclusive requests; batched requests time out individually.
+        await _acquire_inference_lock(0)
```
dpalmqvist: **[review:correctness] Cross-model starvation: the fairness latch only sees exclusive waiters.** `exclusive_pending = lambda: _queue_depth > 0`, but another model's BatchScheduler manager waits in `_acquire_inference_lock(0)` without bumping `_queue_depth`, so a busy model's worker never pauses admission and a steady stream to model A starves every batched request to model B into queue-timeout 503s. Fix: bump `_queue_depth` around the manager's acquire so waiting managers count as exclusive-pending for other schedulers.

olmlx/engine/inference.py
```diff
@@ -2548,6 +2548,343 @@ async def _store_prompt_cache_after_generation(
...
+    stop_sequences: list[str] | None = gen_kwargs.pop("stop", None)
+    if prompt_tokens is not None:
+        tokens = list(prompt_tokens)
+    elif isinstance(prompt, str):
+        tokens = tokenize_for_cache(lm.text_tokenizer, prompt)
+    else:
+        tokens = list(prompt)
+
+    scheduler = _get_batch_scheduler(lm)
+    seq = await scheduler.submit(
```
dpalmqvist: **[review:correctness] Orphaned batch slot: statements between `scheduler.submit()` and the try/finally that guarantees `scheduler.cancel(seq)`.** An exception or task cancellation landing in that window (detokenizer access, settings reads) leaves the worker decoding a sequence nobody consumes, to max_tokens, holding a completion slot. Fix: hoist the setup above submit and enter the guarded region immediately (seq=None; assign inside try; finally guards `seq is not None`).

olmlx/engine/inference.py
```diff
@@ -2548,6 +2548,343 @@ async def _store_prompt_cache_after_generation(
...
+                    else:
+                        event = await seq.out.get()
+                    awaiting_first = False
+                    etype = event["type"]
+                    if etype == "progress":
+                        continue
+                    if etype == "error":
+                        terminal_seen = True
+                        exc = event["exc"]
+                        raise RuntimeError(f"Batched generation failed: {exc}") from exc
```
dpalmqvist: **[review:correctness] Worker errors are type-erased: `raise RuntimeError(...) from exc` loses the original exception type**, so MemoryError (→503 via app handler) and ValueError (→422) from a batched forward all surface as generic 500s. Fix: re-raise the original `exc` (matching exclusive-path propagation); each consumer raising the same instance is acceptable.

olmlx/engine/inference.py
```diff
@@ -2548,6 +2548,343 @@ async def _store_prompt_cache_after_generation(
...
+    def _scan_stop(piece: str) -> tuple[str, bool]:
+        """Append *piece* to the accumulated text and truncate at the
+        earliest stop-sequence match (same semantics as the exclusive
+        path's inline scan)."""
+        nonlocal accumulated_text
+        if not stop_sequences:
+            return piece, False
+        prev_len = len(accumulated_text)
+        accumulated_text += piece
+        stop_idx = -1
```
dpalmqvist, line 2799: **[review:efficiency] `_scan_stop` is O(n²): `accumulated_text.find(stop_seq)` searches from index 0 on every token** over the whole generation, on the event-loop thread shared by all concurrent batched streams. A match ending in the new piece must start within `len(stop_seq)-1` of the boundary, so search from `max(0, prev_len - len(stop_seq) + 1)` — exact and O(piece). (The exclusive path has the same pre-existing pattern; fixing the new copy at least avoids importing it.)

olmlx/engine/batching.py
```diff
@@ -0,0 +1,372 @@
...
+    """
+
+    def __init__(self, request: BatchRequest, loop: asyncio.AbstractEventLoop):
+        self.request = request
+        self.loop = loop
+        self.out: asyncio.Queue[dict[str, Any]] = asyncio.Queue()
+        self.cancelled = threading.Event()
+        self.uid: int | None = None
+
+    def emit(self, event: dict[str, Any]) -> None:
```
dpalmqvist, line 139: **[review:hardening] The per-sequence event bridge is an unbounded queue with no backpressure** (vs CancellableStream's bounded Queue(32) with blocking puts): a stalled-but-connected SSE client lets the worker decode its sequence to max_tokens, buffering every event dict and occupying a batch slot. Suggest Phase-3: consumer-lag threshold → cancel, mirroring the exclusive path's bounded-buffer semantics. At minimum record in the plan's risks.

olmlx/engine/inference.py
dpalmqvist: **[review:reuse] `acquire_gpu`/`release_gpu` re-implement `_inference_locked`'s entry/exit protocol** (deferred-cleanup handshake, boundary syncs, release-on-error). Any future change to the #284 lock-boundary discipline applied only to one copy leaves the other unsafe. Worth extracting shared `_lock_entry/_lock_exit` helpers — can be the Phase-2 PR if touching the exclusive lock path here is too risky. Related: `_scan_stop` is the third copy of the stop-sequence scan (exclusive streaming + `_full_completion`), and `_fail_inbox`/`_cancel_inbox` are the same drain loop differing only in the emitted event.

olmlx/engine/inference.py
```diff
@@ -2548,6 +2548,343 @@ async def _store_prompt_cache_after_generation(
...
+                    stats.eval_count += 1
+                    detokenizer.add_token(event["token"])
+                    piece = detokenizer.last_segment
+                    if piece:
+                        piece, stop_hit = _scan_stop(piece)
+                        if piece:
+                            yield {"text": piece, "done": False}
+                    if stop_hit:
+                        scheduler.cancel(seq)
+                        cancelling = True
```
dpalmqvist: **[review:simplification] Derivable/dead state in `_stream_completion_batched` and `BatchSequence`:** `cancelling` is exactly `stop_hit or timed_out` at every read site (and the `not stop_hit` guard in the done-branch is then dead); `generation_complete` only gates the error metric, which the `exc_type` check already covers; `BatchSequence.uid` is written but never read (the worker tracks uid→seq via `active`). Dropping these shrinks the five-flag state machine to its three real states.

olmlx/engine/batching.py
dpalmqvist: Deferred to Phase 3 with a plan-doc risk entry (docs/batching-plan.md §11): the right shape is a consumer-lag threshold → cancel (mirroring CancellableStream's bounded-buffer semantics), which needs tuning against real SSE stall patterns. The slot is already freed on disconnect via the cancel path; the residual exposure is a stalled-but-connected client.

olmlx/engine/inference.py
dpalmqvist: Deferred to the Phase 2 PR (plan §10 updated): extracting shared lock entry/exit helpers means touching `_inference_locked`'s exclusive hot path, which is riskier in this PR than the duplication. The `_fail_inbox`/`_cancel_inbox` merge from this comment IS done here (b25e13a); the stop-scan consolidation rides along with Phase 2's grammar work.

Issue #123, opened 2026-03-28
bug: prompt cache state corruption on mid-stream disconnect


dpalmqvist:
## Summary

When streaming with prompt cache enabled, the cache object is mutated in-place during generation. If the client disconnects mid-stream, the partially mutated cache is returned to the store. The next request reusing that cache ID gets a corrupted cache, leading to wrong output or crashes.

## Details

Flow:
1. Cache obtained from store: `cached = store.get(cache_id)`
2. Cache dict mutated in-place during token generation
3. Client disconnects → `generation_complete` stays `False` → cache removed from store (correct)

However, the `finally` block at `inference.py:1231-1234` only removes the cache if `generation_complete` is False. There's a window where:
- Generation produces some tokens
- Cache is updated
- An exception occurs after `generation_complete` would have been set

Additionally, the cache trimming logic (`inference.py:1150-1171`) handles None-ID tokens with complex logic that increases corruption risk if the token counts diverge.

## Impact

- Corrupted KV cache → wrong model output on subsequent requests
- Severity: **High**

## Suggested Fix

- Deep-copy the cache before passing to the generator, only commit back on success
- Add validation that `len(stored_tokens) == actual_total` after trimming
- Consider not caching when None-ID tokens are encountered

PR #508, opened 2026-06-12
feat(engine): continuous batching Phase 2 — prompt-cache round trip, non-streaming, grammar, observability

dpalmqvist:
Implements **Phase 2** of `docs/batching-plan.md` (§10), following Phases 0–1 in #507.

## Prompt-cache round trip (plan §4 item 2)

- New `PromptCacheStore.take` / `async_take`: **move semantics** — the entry leaves the store atomically on the loop thread. Batched consumers are not serialized by the inference lock, so the exclusive path's get-then-remove sequence (Bug #123) isn't race-free here; two concurrent requests must never share the mutable cache object. Radix sibling matches are taken under their old id and re-stored under the new `cache_id` at finish (same effect as `takeover`).
- `_setup_batched_prompt_cache` trims the taken cache to the common prefix (plain-`KVCache` trim is offset bookkeeping, no Metal work — safe on the event loop while the worker decodes) and seeds the sequence via `insert(caches=…, all_tokens=…)`.
- On finish, the worker ships the per-sequence cache back in the `done` event. `BatchKVCache.extract` provably strips left padding (slices `left_padding[idx]:_idx` through `mx.contiguous`, resets offset) — the §11 open question, now resolved in the plan. The worker **`mx.eval`s the extracted state before it crosses threads** (#284 hazard family); the consumer re-stores prompt+generated under the `cache_id`, so radix reuse keeps working on entries that passed through a batch.
- Stop-hit cancellations still store (they're successful completions); timeout/disconnect clear `seq.want_cache` and invalidate, matching the exclusive path's invalidate-on-incomplete; a batch **queue timeout restores the taken entry** so a 503'd retry keeps its prefix (parity with the exclusive path, whose queue timeout fires before cache setup).

## Non-streaming path (plan §4 item 6)

`_full_completion` routes eligible requests through `_full_completion_batched`, which drains the batched stream (the `collect_stream` pattern) and aggregates the same result-dict shape. Side benefit: `inference_timeout` is now enforced on this path — the exclusive non-streaming path can't cancel its GPU thread, but the batch worker can drop a sequence at any tick.

## Grammar per-sequence (plan §11 → resolved)

Eligibility exclusion lifted after source-verifying mlx-lm: `GenerationBatch._step` calls each sequence's processors with `logits[e:e+1]` (`[1, vocab]`) and a per-sequence `TokenBuffer` history seeded with the full prompt (`PromptProcessingBatch.prompt()` appends prompt tokens before the generation transition) — exactly `GrammarLogitsProcessor`'s first-call-is-prompt contract. Gated by a live JSON-mode test at 3-way concurrency.

## Observability (plan §8)

- `BatchScheduler.stats()` — active/queued sequences + cumulative inserts/tokens (single-writer worker-thread counters, read from the loop).
- `/api/ps`: new `batch_metrics` field per running model.
- Prometheus: `olmlx_batch_active_sequences`, `olmlx_batch_inserts_total`, `olmlx_batch_aggregate_tokens_total` (folded through `CounterAccumulator` so scheduler recreation can't make them go backwards).

## PR #507 review follow-ups

- **Shared lock helpers**: `_enter_inference_lock` / `_exit_inference_lock` replace the three duplicated copies of the #284 lock-boundary protocol (`_inference_locked`, `_stream_completion`'s inline copy, the batch scheduler's `acquire_gpu`/`release_gpu`). Behavior-preserving; the queued-log message stays per-site.
- **One stop-sequence scan**: `engine/stop_sequences.py` (`StopScanner` for both streaming paths — bounded O(piece) incremental scan — plus `truncate_at_stop` for non-streaming) replaces the three inline copies, including the exclusive path's previously O(generation)-per-step rescan.

## Testing

- 35 new unit tests (scheduler cache round trip, consumer round trip incl. move-semantics/disconnect/queue-timeout-restore, `take`/`async_take`, eligibility, `_full_completion_batched`, collector + `/api/ps`); full suite 4799 passed.
- 3 new live tests (Qwen3-4B-4bit, all passing on this machine): cache round trip **through a real mixed-length batch** (left-padding → extract → turn-2 reuse at exact greedy parity with the exclusive path, `cache_read_tokens > 0`), batched grammar JSON mode ×3 concurrent, non-streaming parity.

Phases 3+ (aggregate admission control, fairness tuning, USER_MANUAL docs, hybrid `ArraysCache` models) remain per the plan.


olmlx/engine/batching.py
```diff
@@ -337,13 +374,22 @@ def _worker(self) -> None:
                         seq = active.get(r.uid)
                         if seq is None:
                             continue
+                        self._tokens_total += 1
```
dpalmqvist: **[low/observability]** (Claude Code review) `_tokens_total` increments for every generation response, including the EOS step whose token event is never emitted (`finish_reason == "stop"` skips the emit two lines down). The consumer-side `stats.eval_count` counts only emitted token events, so `olmlx_batch_aggregate_tokens_total` runs +1 per stop-finished sequence ahead of `olmlx_inference_tokens_total{direction="completion"}` — two "tokens generated" series that disagree on a dashboard. Suggest moving the increment under `if r.finish_reason != "stop":` (length-finished final tokens are emitted, so they stay counted) and updating `test_counters_after_drain`.

olmlx/engine/inference.py
```diff
@@ -2830,6 +2947,22 @@ def _scan_stop(piece: str) -> tuple[str, bool]:
...
+                            # Give the taken cache entry back so a retry
+                            # still gets its prefix (the exclusive path's
+                            # queue timeout fires before cache setup, so
+                            # it never loses the entry). The trimmed
+                            # object covers exactly history_tokens and is
+                            # only ever *read* by the batch (merge
+                            # copies), so re-storing is safe even if the
+                            # sequence was admitted before the sweep.
+                            if cache_setup is not None and cache_setup.cache:
+                                await lm.prompt_cache_store.async_set(
```
dpalmqvist: The restored entry can't be stale-inconsistent: it's written as `tokens=history_tokens` — exactly the prefix the trimmed cache covers, so token metadata ↔ KV depth match by construction. The batch only ever *reads* a seeded cache (`_merge_caches` copies into batched buffers; the consumer's finally cancels the sequence with `want_cache=False`, so the worker never re-stores a second copy). And the exclusive path's queue timeout doesn't treat this as a miss — it fires *before* cache setup, leaving the entry in the store and available for retry. The restore reproduces precisely that outcome; dropping the entry (pre-fix behavior) was the asymmetry.

olmlx/engine/batching.py
```diff
@@ -361,6 +410,23 @@ def _worker(self) -> None:
...
+    def _attach_cache(done: dict[str, Any], cache: list[Any], tokens: Any) -> None:
+        """Materialize an extracted per-sequence cache and attach it to a
+        done event.
+
+        ``BatchKVCache.extract`` builds the B=1 cache as a lazy slice
+        graph on the worker's stream; eager-evaluate it *here* (on the
+        worker thread, which owns the GPU) so the consumer never
+        materializes a cross-thread lazy graph when it re-stores or
+        reuses the cache (#284 hazard family).
+        """
```
dpalmqvist, line 423: Both halves are already handled: (a) the None case is guarded on this exact line — `if seq.want_cache and cache is not None` (and the sweep path additionally checks `if cache is not None` after unpacking); (b) storing a partially generated sequence is valid by construction — the stored token list is mlx-lm's `all_tokens`, i.e. exactly the tokens whose KV is in the cache, and a later request reuses only the common token prefix and trims. There is no 'complete prompt boundary' requirement for prompt-cache entries; the exclusive path likewise stores prompt+generated on stop-sequence truncation (tokens beyond what the client saw). Only stop-hit cancellations keep `want_cache` — a successful completion — and the consumer's `len(tokens) >= len(prompt)` guard excludes mid-prefill extractions.

olmlx/engine/batching.py
```diff
@@ -361,6 +410,23 @@ def _worker(self) -> None:
...
+        import mlx.core as mx
+
+        mx.eval([c.state for c in cache])
```
dpalmqvist, line 426: `state` is the canonical `_BaseCache` interface in mlx-lm — every cache class including `CacheList` exposes it (olmlx's snapshot/persistence code relies on the same contract). More to the point, v1 eligibility (`caches_plain_kv`) only admits models whose layers are exactly `KVCache`, so `extract` returns plain `KVCache` objects here; any future batchable layout has to land by editing that probe, which puts this line squarely in that change's review scope. A loud failure is the intended behavior over silently skipping persistence.

olmlx/engine/inference.py
```diff
@@ -2762,7 +2879,18 @@ async def _stream_completion_batched(
...
-    _batched_kv_preflight(lm, len(tokens), max_tokens)
+    cache_setup: _BatchedCacheSetup | None = None
+    if use_prompt_cache and make_prompt_cache is not None:
+        cache_setup = await _setup_batched_prompt_cache(lm, tokens, cache_id)
+        submit_tokens = cache_setup.suffix_tokens
+    else:
+        submit_tokens = tokens
+
+    # New-KV estimate covers only what this sequence will prefill; a
+    # reused prefix is already-resident memory moved out of the store.
```
dpalmqvist, line 2915: This mis-reads the flow: when `prompt_tokens` is None and the prompt is a string, `_stream_completion_batched` tokenizes it first (`tokens = tokenize_for_cache(...)`) and *then* runs `_setup_batched_prompt_cache(lm, tokens, cache_id)` — the lookup/trim logic is never skipped for string prompts. The preflight using the suffix length is intentional, not an underestimate: the reused prefix KV is memory that was already resident (it moved out of the store, not newly allocated), and `get_metal_memory()` in the estimate already counts it; only the suffix + generation window is new KV.

olmlx/engine/prompt_cache/store.py
```diff
@@ -659,6 +681,43 @@ async def async_get(self, cache_id: str) -> CachedPromptState | None:
...
+                await asyncio.to_thread(
+                    self._unlink_and_refresh, self._disk_file_path(cache_id)
+                )
+            self.metrics.cache_id_hits += 1
+            return state
+        if not self._disk_enabled:
+            self.metrics.cache_id_misses += 1
+            return None
+        gen_before = self._evict_generation
+        loaded, disk_path = await self._to_thread_traced(self._read_from_disk, cache_id)
```
dpalmqvist, line 706: No shared-mutation hazard exists in that interleaving: the disk fallback deserializes a *fresh* state object (`_read_from_disk` → `load_prompt_cache`), so if another coroutine repopulates the id during the await, the store holds a different object than the one returned — the move-semantics guarantee is 'no two consumers share one mutable cache object', and that holds. Popping the repopulated entry instead would be worse: it would steal a cache another in-flight request just stored. This mirrors `async_get`'s documented stance on the same window (it prefers the fresh in-memory entry; we leave it untouched).

Issue #501, opened 2026-06-12
Speculative decoding: tree-based draft verification


dpalmqvist:
From a performance brainstorm. olmlx already has 6 speculative strategies (classic/dflash/eagle/mtp/pld/self), but every one verifies a **single linear draft chain** — λ candidates in a row for classic, one block for the others.

**Idea:** Medusa/EAGLE-2-style **tree** drafting + tree attention. Instead of one candidate continuation, draft a tree of branching candidates and verify them all in a single target forward via a block-diagonal tree attention mask. Accepted-tokens-per-verify rises because you cover multiple plausible continuations at once — biggest single decode win on structured text where the draft is often *almost* right.

Fits the existing architecture:
- `SpecDecoderBase` (`engine/spec_decoder_base.py`, #467) already owns prefill/step/verify mechanics; tree verify is a new `_step_impl` shape.
- `verify_draft_greedy` would need a tree-aware variant (mask + per-branch accept walk).
- Natural fit for MTP and EAGLE drafts (they already produce per-position logits to sample a tree from).
- GDN rollback (`_GDNStateCapture`) already handles partial-accept rollback on hybrid targets — tree accept is the same rollback, deeper.

Highest-leverage of the brainstorm items. Worth it if decode throughput on code/JSON workloads is the priority.

dpalmqvist:
Closing as **not planned** — negative result from the Phase-1 spike.

The tree-drafting feature was investigated and rejected: per PR #519, tree verification benchmarked **~2× slower than the linear chain** on a 4-bit target, and the GDN/hybrid path would need from-scratch design work. Only the regression fix (reviving the bit-rotted `OLMLX_TREE_SPECULATIVE` path) + a real-model canary test were kept on `main`. The full negative-result write-up lives on branch `spec/tree-draft-verification-501`.

If the GDN-path redesign is ever worth revisiting, it should come back as a fresh, narrower issue.

Issue #358, opened 2026-05-25
Tree-based speculative verify


dpalmqvist:
## Why

Accept multiple drafted branches per target forward via a sparse-tree attention mask. Multiplies acceptance over linear verification with only a minor verify-time cost.

## Scope

- Extend `verify_draft_greedy` in `engine/speculative.py` to accept a tree-shaped draft (mask + per-branch scoring)
- Wire `classic` and `pld` (the prompt-lookup strategy) to produce tree drafts
- `dflash`/`eagle` tree drafts are out of scope (follow-up)

## Acceptance criteria

- Output identical to greedy on property tests
- ≥10% throughput gain on `classic` over linear verify on the standard bench
- Tree depth/width tunable via env + per-request param

## Files

- `engine/speculative.py`
- `engine/speculative_stream.py`

## Related

Pairs naturally with the prompt-lookup speculative strategy issue.

PR #519, opened 2026-06-14
fix(speculative): revive bit-rotted tree-forward + add real-model canary (#501)

dpalmqvist:
## What

Repairs the `OLMLX_TREE_SPECULATIVE` tree-verification path (#358), which had silently bit-rotted and was crashing on the **first tree step** against the current mlx/mlx-lm stack, and adds the real-model test coverage whose absence let it rot unnoticed.

This is the salvage from a Phase-1 investigation of #501 (tree-based draft verification). The tree *feature* itself did not pan out — benchmarked ~2× slower than the linear chain on a 4-bit target, and the GDN path needs from-scratch design work — so only the regression fix + test are kept here. (Full negative-result write-up lives on the `spec/tree-draft-verification-501` branch.)

## The two regressions fixed

Both uncaught because **no real-model test exercised the tree path**:

1. **Forward signature** — `_tree_call` accepted `(inputs, cache)`, but mlx-lm's Qwen3 wrapper now calls the inner model with a third positional `input_embeddings`, raising `TypeError` on the first tree step. Now threaded through to `embed_fn`.
2. **Mask dtype** — the additive tree mask was `float32`; current mlx `scaled_dot_product_attention` requires the mask to promote to the model's `bf16` output dtype. Now cast to the hidden-state dtype, mirroring mlx-lm's own `create_attention_mask`.

`tree_width >= 2` runs end-to-end again.

## Test

New live canary `tests/live/test_tree_speculative_real.py` (real_model; skipped in CI and when the model isn't downloaded). It asserts both the tree (`tree_width>=2`) and linear paths match the target's **pure-greedy decode**, tolerating divergences that land only on exact-tie positions (top1−top2 logit margin < `TIE_EPS`) and failing loudly on any clear-margin divergence — the real exactness contract under argmax tie-breaking. Verified: the tree path matches pure greedy exactly while the linear path tie-breaks one 0.0-margin position differently (so a naive `tree == linear` assertion was the wrong invariant).

## Scope

- Only `olmlx/engine/tree_speculative.py` (+13/−3) and the new test file.
- No behavior change to any non-tree path; `OLMLX_TREE_SPECULATIVE` remains off by default.
- GDN/hybrid targets still not supported by the tree path (separate, documented design gap — `_step_tree` assumes a trimmable `KVCache` at layer 0).

### DIFF
--- a/olmlx/engine/tree_speculative.py
+++ b/olmlx/engine/tree_speculative.py
@@ -391,12 +391,22 @@ def _patch_target_for_tree_forward(
 
     orig_call = type(inner).__call__
 
-    def _tree_call(self_: Any, inputs: mx.array, cache: Any = None) -> Any:
-        h = embed_fn(inputs)
+    def _tree_call(
+        self_: Any,
+        inputs: mx.array,
+        cache: Any = None,
+        input_embeddings: Any = None,
+    ) -> Any:
+        h = input_embeddings if input_embeddings is not None else embed_fn(inputs)
+        # Cast the additive mask to the hidden-state dtype: current mlx's
+        # scaled_dot_product_attention requires the mask to promote to the
+        # output dtype, and a float32 mask cannot downcast to a bf16 model.
+        # Mirrors mlx-lm's create_attention_mask, which builds in h.dtype.
+        mask = tree_mask.astype(h.dtype)
         if cache is None:
             cache = [None] * len(layers)
         for layer, c in zip(layers, cache):
-            h = layer(h, tree_mask, c)
+            h = layer(h, mask, c)
         return norm_fn(h)
 
     type(inner).__call__ = _tree_call  # type: ignore[method-assign]
--- a/tests/live/test_tree_speculative_real.py
+++ b/tests/live/test_tree_speculative_real.py
@@ -1,17 +1,18 @@
 """Live greedy-parity canary for tree speculative verification (#501).
 
-Greedy speculative decoding is exactness-preserving: its output must equal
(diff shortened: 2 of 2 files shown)


dpalmqvist:
Thanks — triaged the aider findings:

**Addressed** (pushed in 04e55a9):
- **#2 / #3 (canary tie-tolerance + stops at first divergence):** valid. Rewrote the canary to *teacher-force* each path's full output back through the target in one forward and assert every position is the target's greedy argmax, with per-position tie tolerance (`margin < TIE_EPS`). It now checks the **whole** sequence, so a real corruption landing after a benign tie-break is caught; any clear-margin mismatch fails loudly.
- **#1 (greedy reference reused one cache):** that was correct KV-cache usage, but the rewrite removes that helper entirely, so it's moot.

**Declined, with reasons:**
- **#4 (`zip(layers, cache)` truncation):** this mirrors mlx-lm's own `Model.__call__` (`for layer, c in zip(self.layers, cache)`) exactly — not a regression, and the cache always has one entry per layer here.
- **#5 (mask quadratic / patch leak):** the mask is ≤ `tree_max_nodes`² (≤16×16, trivial), and the patch is installed/removed under a `try/finally` in `_step_tree`, so it can't leak to the linear path.
- **#6 (monkey-patch fragility):** fair, but it's pre-existing #358 design, out of scope for a regression fix — and the new canary is precisely the guard that catches this class of break (it already caught this one). The tree path stays off by default (`OLMLX_TREE_SPECULATIVE`).

PR #526, opened 2026-06-15
feat(proxy-tuning): Stage 1 — olmlx-domain SFT data pipeline

dpalmqvist:
## Summary

**Stage 1 of 3** of the olmlx-domain proxy-tuning effort (design spec: `docs/superpowers/specs/2026-06-15-olmlx-proxy-tuning-pair-design.md`). This adds an **offline data pipeline** that mines the olmlx repo into a curated, chat-format SFT dataset for training the proxy-tuning expert M⁺.

New importable package `olmlx/proxy_tuning_pipeline/` (**never imported by the running server** — pure offline tooling; `openai` is lazy-imported):

- **Extraction** (`extract.py`) — mines grounded `(source, seed)` units from five sources: functions+docstrings, CLAUDE.md invariants, `docs/` sections, test functions, and git commits (message+diff). Secret-stripped, provenance-tagged, deduped.
- **Expansion** (`expand.py`) — a `Generator` protocol + `OpenAIGenerator` (GPT-5.4-mini, dependency-injected) turns each unit into diverse instruction→response pairs, grounded strictly in the source. Balanced-brace JSON extraction tolerates prose-wrapped replies; systemic-failure logging surfaces an all-failed run.
- **Curation** (`curate.py`) — quality filter, normalized-exact dedup, deterministic seeded train/valid split.
- **CLI** (`cli.py`) — `run_pipeline` (testable core) + `main` wires extract → expand → curate → write `train.jsonl`/`valid.jsonl` in **mlx-lm chat format**, with generated-vs-kept stats (no silent truncation).

Built via subagent-driven TDD (per-task implement + two-stage spec/quality review); a final whole-pipeline review returned **ready to merge**.

## Design choices
- `Generator` is injected → all tests run with no network (a fake client/generator); the real OpenAI call is isolated to one method.
- Dedup is normalized-exact (no embedding/ML deps — YAGNI).
- Intermediate-artifact persistence is a documented deferral; an in-code `TODO` flags adding 429 retry before the full ~1.7hr run.

## Test Plan
- [x] `uv run pytest tests/test_proxy_tuning_pipeline.py` — 30 unit/integration tests (extractors on fixtures incl. a temp git repo; OpenAI mocked via injected client; curation; end-to-end with a fake generator; malformed-file robustness)
- [x] `ruff check` + `ruff format` clean
- [ ] **Manual (yours):** real GPT-5.4-mini dry run on the repo (`OPENAI_API_KEY=... uv run python -m olmlx.proxy_tuning_pipeline.cli --repo . --out /tmp/ptdata_smoke --n-per-unit 1`) to eyeball grounding quality before the full ~10k pass (Task 15 in the plan)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ htmlcov/
 .claude/scheduled_tasks.lock
 .worktrees/
 .bench-runs/
+
+# Proxy-tuning generated datasets (large, regenerable)
+data/proxy_tuning/
--- a/docs/superpowers/plans/2026-06-15-proxy-tuning-data-pipeline.md
+++ b/docs/superpowers/plans/2026-06-15-proxy-tuning-data-pipeline.md
@@ -1,0 +1,1625 @@
+# Proxy-Tuning Data Pipeline (Stage 1) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build the offline data pipeline that turns the olmlx repo into a curated ~10k-example chat-format SFT dataset (`train.jsonl` / `valid.jsonl`) for training the proxy-tuning expert M⁺.
+
+**Architecture:** A three-phase offline pipeline in a new importable package `olmlx/proxy_tuning_pipeline/`: (1) **mechanical extraction** mines grounded `(source, seed)` units from functions, CLAUDE.md invariants, docs, tests, and git commits — pure stdlib, deterministic; (2) **expansion** sends each unit to a `Generator` (OpenAI GPT-5.4-mini by default, dependency-injected so tests use a fake) to produce diverse instruction→response pairs; (3) **curation** dedupes, filters, and splits into mlx-lm chat JSONL. A thin CLI wires the phases with intermediate artifacts under `data/proxy_tuning/` (gitignored).
+
+**Tech Stack:** Python 3.11 (stdlib `ast`, `subprocess`, `json`, `re`, `hashlib`, `random`), `openai>=1.66` (already a dependency; lazy/injected), pytest.
+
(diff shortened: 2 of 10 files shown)


dpalmqvist:
Thanks — addressed the actionable findings in 7e24ba1:

- **#1 (expand.py:47, JSON string escaping):** `_extract_json_object` now tracks string/escape state so a literal `{`/`}` inside a quoted value (responses contain olmlx code snippets) no longer miscounts depth. Added a regression test.
- **#2 (extract.py decoding):** `extract_functions`/`extract_tests` now read with `encoding="utf-8", errors="replace"`, consistent with the markdown readers.
- **#3 (cli.py:56, checkpoint reuse — high):** `run_pipeline` now records `{repo_root, n_per_unit}` in a `raw.meta.json` sidecar and **refuses** to reuse an `--out` built for a different repo/config. Backward-compatible: a legacy `raw.jsonl` with no sidecar is adopted with a warning so an in-flight run stays resumable. Tests cover both.

Intentional, not changed:
- **#4 (sort vs raw.jsonl order):** by design — `raw.jsonl` is the resume/debug checkpoint; `train.jsonl`/`valid.jsonl` is the curated, deterministically-sorted dataset. They're distinct artifacts.
- **#5 (client-init error surfacing):** a construction/auth failure is already surfaced — it fails every unit's future (per-unit warning) and triggers the `0 examples` error log; not silent.
- **#6 (git show I/O):** negligible at this scale — extracting all 7,862 units runs in seconds.

dpalmqvist:
Re: the security advisory — assessed as **not blocking** for this use:

- **Code transmission:** `motsognirr/olmlx` is a **public** repo. The source and full git history are already public, so sending them to the generation API exposes nothing that isn't already public — the "proprietary code to third party" concern doesn't apply.
- **Secrets in history:** in a public repo, any committed secret is already compromised; `strip_secrets` remains a sensible belt-and-suspenders, not the primary control. (For reference: it covers `sk-*`, `hf_*`, and `NAME=secret` shapes.)
- **Consent:** the pipeline is an explicit operator-run CLI with the user's own `OPENAI_API_KEY` — consent is the act of running it.

Worth keeping in mind only if the pipeline is ever pointed at a **private** repo: then stronger secret scanning (e.g. a dedicated scanner over history) and an explicit consent gate would be warranted before generation. Out of scope for this Stage-1 PR.

Issue #535, opened 2026-06-20
Factory smoke test: confirm Plan stage triggers on new issues


dpalmqvist:
This is a throwaway smoke-test issue to confirm the agentic factory's Plan stage wakes on a newly opened issue and runs on the self-hosted `agent` runner.

Expected: Factory · Plan picks this up, explores the repo, and either triage-closes it or posts a plan + `plan:proposed`. Since this is just a connectivity test (not a real feature request), a triage-close is the correct outcome.

No code change is being requested. Will be closed once the Plan stage has demonstrably run.

dpalmqvist:
Closing to re-open and re-trigger the now-fixed Factory · Plan stage (PR #536 merged).

PR #537, opened 2026-06-20
fix(factory): give agents gh CLI access in agent mode

dpalmqvist:
## Follow-up to #536 — the re-test still failed

Reopening #535 re-triggered Factory · Plan, which **again no-opped**: `permission_denials_count: 18`, no comment / label / close.

## Refined root cause

The SDK log shows the run is **claude-code-action agent mode** (`Auto-detected mode: agent for event: issues`), launched with:
```
SDK options: { "disallowedTools": ["Edit","Write","Bash"], systemPrompt: preset }
```
**Agent mode wires up no GitHub MCP server** (no `mcpServers` in the SDK options; no tracking comment is posted). The only way for the agent to touch GitHub is the **`gh` CLI via Bash**. #536's `--disallowedTools "Edit,Write,Bash"` on the planner therefore removed the one tool it needed, and none of the four workflows exported a token for `gh` to authenticate with.

## Fix (all four workflows)

- **Planner:** drop the tool restriction so it has `Bash`. It stays read-only on the repo via `permissions: contents: read` (it physically cannot push code or open PRs regardless of tools).
- **All four:** add `GH_TOKEN` to the Claude step `env:` so `gh` is authenticated —
  - planner → `GITHUB_TOKEN` (preserves the approval-gate invariant: its label writes can't trigger `factory-implement`)
  - explore / implement / babysit → `FACTORY_PAT` (their output must trigger downstream)
- Make each prompt use the `gh` CLI explicitly.
- Correct the stale comments that claimed the action defaults provide GitHub MCP tools.
- **`factory-plan` only:** `show_full_output: true` **temporarily**, to verify exact tool access on the next run — to be reverted in a follow-up once confirmed.

## Verification plan

Merge → reopen #535 → confirm the planner now posts a plan / triage-closes, using the full output to see the actual `gh` calls. Then a follow-up reverts `show_full_output`.

### DIFF
--- a/.github/workflows/factory-plan.yml
+++ b/.github/workflows/factory-plan.yml
@@ -53,10 +53,6 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
-          # TEMP (remove after verifying agent-mode tool access): stream the full
-          # JSON so we can see exactly which tools run. WARNING: may surface
-          # secrets in logs — drop this once the factory is confirmed working.
-          show_full_output: true
           # SECURITY — APPROVAL-GATE INVARIANT: the planner MUST stay on the
           # default GITHUB_TOKEN. Do NOT add a `github_token: FACTORY_PAT`
           # override, and keep GH_TOKEN (above) on GITHUB_TOKEN too. The single
--- a/.github/workflows/factory-plan.yml
+++ b/.github/workflows/factory-plan.yml
@@ -64,12 +64,15 @@ jobs:
           #
           # Tools: agent mode (a non-@claude `issues` trigger) provides NO GitHub
           # MCP server — the only way to touch GitHub is the `gh` CLI via Bash.
-          # So we must NOT restrict tools (an earlier --allowedTools/
-          # --disallowedTools that dropped Bash left the planner unable to act at
-          # all). The planner stays read-only on the REPO via
-          # `permissions: contents: read` — it physically cannot push code or
-          # open PRs no matter which tools it holds.
-          claude_args: --max-turns 40
+          # The planner processes UNTRUSTED issue text on a self-hosted runner,
+          # so we scope Bash to `gh` only (`Bash(gh:*)`) rather than granting a
+          # full shell: it can read/comment/label/close via gh but cannot run
+          # arbitrary commands under prompt injection. Read/Grep/Glob cover code
+          # exploration. (Dropping Bash entirely — as earlier revisions did —
+          # left the planner unable to act at all.) It is additionally read-only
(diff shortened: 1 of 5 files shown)


.github/workflows/factory-plan.yml
```diff
@@ -44,37 +44,50 @@ jobs:
...
+          # gh CLI in the agent's Bash shell needs an explicit token. Use the
+          # default GITHUB_TOKEN (NOT FACTORY_PAT): any label the planner applies
+          # is then written with GITHUB_TOKEN, which CANNOT trigger
+          # factory-implement (issues: labeled) — this is what mechanically
+          # preserves the human approval gate. See APPROVAL-GATE INVARIANT below.
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          # TEMP (remove after verifying agent-mode tool access): stream the full
+          # JSON so we can see exactly which tools run. WARNING: may surface
```
dpalmqvist: Fixed in 7ff9bcd — removed `show_full_output: true` entirely rather than committing it to the default branch. Agreed it's a real leak risk on scheduled runs if the revert were forgotten. Verification of the agent-mode `gh` fix doesn't actually need it: success is observable from the issue state (comment/label/close) plus the normal result block's `permission_denials_count`, both available without the verbose stream.

.github/workflows/factory-plan.yml
```diff
@@ -44,37 +44,46 @@ jobs:
...
-          # body drive code to merge. The plan comment doesn't need to trigger
-          # anything downstream, so GITHUB_TOKEN is sufficient.
+          # default GITHUB_TOKEN. Do NOT add a `github_token: FACTORY_PAT`
+          # override, and keep GH_TOKEN (above) on GITHUB_TOKEN too. The single
+          # human approval gate is enforced mechanically, not just by the prompt:
+          # even a prompt-injected planner that applied `plan:approved` cannot
+          # start the implementer, because labels applied with GITHUB_TOKEN do
+          # NOT trigger other workflows (factory-implement fires on
+          # `issues: labeled`). A PAT here would silently remove that protection.
           #
```
dpalmqvist, line 64: Good catch — fixed in 62fa94f. The planner only needs `gh` + file reads, so its Bash is now scoped to `Bash(gh:*)` (`--allowedTools "Read,Grep,Glob,Bash(gh:*)"`): it can read/comment/label/close via gh but can't run arbitrary shell under prompt injection. Its GH_TOKEN is also GITHUB_TOKEN scoped to contents:read + issues:write, so the token blast radius is minimal.

Explore/implement/babysit genuinely need a full shell (pytest, git, uv sync), so they can't be gh-scoped — they keep unrestricted Bash. That residual self-hosted-runner RCE exposure (injection via issue/PR/CI content they read) is inherent to what they do; I've documented it in the factory README threat-model notes (run the factory on an isolated host, keep FACTORY_PAT minimal).

PR #539, opened 2026-06-21
feat(spec): track draft–target quant compat at train and load time

### LINKED ISSUE #516: DFlash/EAGLE draft ↔ target compatibility matrix + versioning
## Story

A draft is bound to a specific target (vocab, head dims, layer ids) and its acceptance is sensitive to the target's *quantization*. A mismatched pairing does **not** error — it silently degrades to the ~0.4%-acceptance failure mode (correct output, no speedup). We need a way to record and (optionally) enforce which drafts are validated against which target builds/quants.\n\n## Why this matters\n\nThis is the real cost of hosting drafts, not the bytes. Without it the hosting + auto-resolution stories ship a footgun.\n\n## Scope\n\n- [ ] Encode the target build in the repo id or a `revision`/tag.\n- [ ] Record validated quant(s) per draft (repo card + curated map entry).\n- [ ] Decide enforcement: warn vs. refuse on a quant/target mismatch at load.\n- [ ] Surface measured acceptance per (draft, target, quant) so a working pairing is distinguishable from a degenerate one.\n\n## Cross-cuts\n\n- #512 (hosting), #513 (curated map), #514 (co-download), #515 (subdir probe)

dpalmqvist:
Closes #516

## Summary

- Adds `target_quant: str | None` to `DraftConfig` and `EagleConfig` (optional, backwards-compatible — `None` on old checkpoints skips checks silently).
- Three new helpers in `engine/speculative_loaders.py`:
  - `_quant_descriptor_from_path(path)` — reads `config.json`/`quantization_config.json`/`quantize_config.json` → `"q4_g64"` or `"bf16"`.
  - `_detect_live_quant(model)` — walks loaded MLX model for `nn.QuantizedLinear` → descriptor.
  - `_check_quant_compat(draft_quant, live_quant, *, draft_path)` — warns on mismatch; raises `RuntimeError` when `OLMLX_SPEC_STRICT_COMPAT=1`.
- `_load_dflash_decoder` and `_load_eagle_decoder` now read `target_quant` from the draft config and call `_check_quant_compat` after existing vocab/hidden-size checks.
- `prepare_dflash_draft` and `prepare_eagle_draft` compute `target_quant` at training time and persist it in `dflash_config.target_quant` / `eagle_config.target_quant` inside `config.json`.
- `SpecDecoderBase` gains `_target_quant: str = ""` and exposes it in `stats_summary()`.
- `config.py`: new `spec_strict_compat: bool = False` (`OLMLX_SPEC_STRICT_COMPAT` env var).

## Tests added

`tests/test_draft_quant_compat.py` — 16 tests:
- `_quant_descriptor_from_path`: bf16, q4 via `quantization`, q4 via `quantization_config`, GPTQ `quantize_config.json`
- DraftConfig / EagleConfig roundtrip with and without `target_quant`
- `_check_quant_compat`: match (no warn), mismatch (warn), None (skip), strict-mode (raise)
- `_detect_live_quant`: bf16 model, quantized model
- `stats_summary` includes `target_quant`

- **Metal stream / load ordering**: all new logic runs inside the existing `_load_dflash_decoder` / `_load_eagle_decoder` call sites after weight loading and vocab checks, before the decoder is returned. No inference path is touched.
- **No vocab/hidden-size check ordering change**: quant check is additive — runs after, not instead of, structural checks.
- **Backwards compatibility**: `target_quant=None` (absent from old checkpoints) skips silently.
- **Strict mode is opt-in**: default warns; `OLMLX_SPEC_STRICT_COMPAT=1` promotes to error.
- **Grammar / proxy-tuning / other strategies**: not touched.

### DIFF
--- a/olmlx/engine/speculative_loaders.py
+++ b/olmlx/engine/speculative_loaders.py
@@ -51,7 +51,8 @@ def _quant_descriptor_from_path(model_path: Path) -> str:
 
     if quant_block and isinstance(quant_block, dict):
         bits = quant_block.get("bits")
-        group_size = quant_block.get("group_size") or 64
+        group_size_raw = quant_block.get("group_size")
+        group_size = group_size_raw if group_size_raw is not None else 64
         if bits is not None:
             return f"q{bits}_g{group_size}"
 
--- a/olmlx/engine/speculative_loaders.py
+++ b/olmlx/engine/speculative_loaders.py
@@ -640,6 +640,7 @@ class SpeculativeLoaderMixin:
             draft_model=draft_model,
             block_size=block_size,
             target_layer_id=target_layer_id,
+            target_quant=draft_config.target_quant,
         )
 
     def _load_mtp_decoder(
--- a/olmlx/engine/speculative_loaders.py
+++ b/olmlx/engine/speculative_loaders.py
@@ -38,6 +38,11 @@ def _quant_descriptor_from_path(model_path: Path) -> str:
     try:
         cfg = json.loads(cfg_path.read_text()) if cfg_path.exists() else {}
     except Exception:
+        logger.warning(
+            "_quant_descriptor_from_path: failed to read/parse %s — defaulting to bf16",
+            cfg_path,
+            exc_info=True,
+        )
         cfg = {}
 
     quant_block = cfg.get("quantization") or cfg.get("quantization_config")
@@ -47,6 +52,11 @@ def _quant_descriptor_from_path(model_path: Path) -> str:
             try:
                 quant_block = json.loads(gptq_path.read_text())
             except Exception:
+                logger.warning(
+                    "_quant_descriptor_from_path: failed to read/parse %s — defaulting to bf16",
+                    gptq_path,
+                    exc_info=True,
+                )
                 quant_block = None
 
     if quant_block and isinstance(quant_block, dict):
@@ -75,7 +85,10 @@ def _detect_live_quant(model: Any) -> str:
                 group_size = module.group_size
                 return f"q{bits}_g{group_size}"
     except Exception:
-        pass
+        logger.debug(
+            "_detect_live_quant: named_modules() inspection failed — defaulting to bf16",
+            exc_info=True,
+        )
     return "bf16"
 
 
--- a/olmlx/engine/speculative_loaders.py
+++ b/olmlx/engine/speculative_loaders.py
@@ -51,7 +51,7 @@ def _quant_descriptor_from_path(model_path: Path) -> str:
 
     if quant_block and isinstance(quant_block, dict):
         bits = quant_block.get("bits")
-        group_size = quant_block.get("group_size", 64)
+        group_size = quant_block.get("group_size") or 64
         if bits is not None:
             return f"q{bits}_g{group_size}"
 
--- a/olmlx/engine/speculative_loaders.py
+++ b/olmlx/engine/speculative_loaders.py
@@ -95,15 +95,16 @@ def _check_quant_compat(
         return
     if draft_quant == live_quant:
         return
-    msg = (
+    base_msg = (
         f"DFlash/EAGLE draft at {draft_path} was trained on a '{draft_quant}' "
         f"target but the loaded target is '{live_quant}' — acceptance rate may "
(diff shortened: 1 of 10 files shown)


olmlx/engine/speculative_loaders.py
```diff
@@ -25,6 +25,101 @@
...
+            try:
+                quant_block = json.loads(gptq_path.read_text())
+            except Exception:
+                logger.warning(
+                    "_quant_descriptor_from_path: failed to read/parse %s — defaulting to bf16",
+                    gptq_path,
+                    exc_info=True,
+                )
+                quant_block = None
+
```
dpalmqvist, line 61: **Dismissed (not a bug).** The default of 64 is MLX's own canonical default for group_size when absent from quantization config — any properly-produced MLX quantized model that omits the key used 64. A model quantized with a non-default group_size (e.g. 128) would have that value in its config; if it doesn't, it either used the default or has a malformed config. In either case, `_detect_live_quant` will discover the actual value at load time. The worst outcome of a mismatch here is a spurious warning, which is safe; strict mode (`OLMLX_SPEC_STRICT_COMPAT=1`) is opt-in for situations where precision matters.

olmlx/engine/speculative_loaders.py
```diff
@@ -25,6 +25,101 @@
...
+def _detect_live_quant(model: Any) -> str:
+    """Return a quant descriptor by inspecting a loaded MLX model.
+
+    Walks the model looking for the first ``nn.QuantizedLinear`` via
+    ``module.modules()``. Returns ``f"q{bits}_g{group_size}"`` or ``"bf16"``
+    if none are found.
+    """
+    import mlx.nn as nn
+
+    try:
```
dpalmqvist, line 81: **Dismissed (deliberate debug level).** The last fix commit (2976fde) specifically addressed silent swallowing by adding `logger.debug` with `exc_info=True`. `named_modules()` is a standard `mlx.nn.Module` method; it failing on a real loaded model is essentially impossible under normal usage. `logger.debug` is intentional over `warning` to avoid noise in production logs — this path can only fire for exotic custom model objects that aren't proper `mlx.nn.Module` instances. If it ever fires, the debug log with traceback is fully diagnostic. The fallback to `"bf16"` is safe: no crash, no data corruption — a compat check is skipped, same as `target_quant=None`.

dpalmqvist:
## Babysit triage — 2026-06-21T16:53Z

**CI:** The `Tests` job on commit `2976fde` failed due to a transient GitHub Actions artifact upload stall — all 5183 tests passed per run log. Pushed an empty retrigger commit (`3603b13`) to get a clean CI run.

**Reviewer findings — all DISMISSED:**

| Finding | Source | Decision | Reason |
|---------|--------|----------|--------|
| `group_size=64` default in `_quant_descriptor_from_path` | gpt-review / deepseek-review | Dismissed | MLX's own canonical default group_size IS 64. Any model whose config omits the key used 64. If mismatched, `_detect_live_quant` catches the actual value at load time; worst outcome is a spurious warning. |
| `except Exception` → `logger.debug` in `_detect_live_quant` | gpt-review / deepseek-review | Dismissed | `named_modules()` is a base-class method that cannot fail on a real `mlx.nn.Module`. `logger.debug + exc_info=True` is the right noise level; fallback to `"bf16"` is safe (same as `target_quant=None`). |
| `_detect_live_quant` may pick up VLM vision encoder quant | gpt-review | Dismissed | DFlash/EAGLE are language-only decoders. VLMs route through separate code paths (see CLAUDE.md hybrid VLM routing) and never reach this check. |

Waiting for CI on `3603b13` before merging.

dpalmqvist:
## Babysit triage — final (2026-06-21)

CI is now green on `3603b13` (all 8 checks pass, including `gpt-review` and `deepseek-review`).

**FIX bucket:** empty — no actionable findings remain.

**Dismissed (carried forward from prior triage):**
- `group_size=64` default assumption — MLX canonical default; `_detect_live_quant` catches any real mismatch at load time.
- `except Exception → logger.debug` — `named_modules()` cannot fail on a real `mlx.nn.Module`; debug level is correct.
- VLM vision encoder quant detection concern — DFlash/EAGLE are language-only; VLMs route through separate paths per CLAUDE.md.

Merging.

PR #555, opened 2026-06-23
fix(tool_parser): classify truncated <think>/Gemma4 blocks as thinking content

### LINKED ISSUE #544: Raw <think> tags leak into content when max_tokens truncates a thinking model mid-reasoning
## Summary

When `max_tokens` (or `max_completion_tokens`) is set small enough to stop generation while a thinking model (e.g. Qwen3) is still in its `<think>` block, the raw, incomplete `<think>` XML leaks into the response's text content field instead of being handled as thinking content or discarded.

## Reproduction

```bash
MODEL="mlx-community/Qwen3-4B-4bit:latest"

# OpenAI non-stream
  -d '{"model": "'$MODEL'", "messages": [{"role": "user", "content": "Say hi."}], "stream": false, "max_tokens": 30}'
```

**Observed:**
```json
{
  "choices": [{
    "message": {
      "role": "assistant",
      "content": "<think>\nOkay, the user said \"Say hi.\" I need to respond appropriately. Let me start by greeting them warmly. Maybe use a friendly tone"
    },
    "finish_reason": "stop"
  }]
}
```

**Expected:** Either empty `content` (thinking not complete, no actual response generated) or a properly structured response with thinking in a separate field (like the Anthropic API's `thinking` content block).

### Same bug affects:

**Anthropic `/v1/messages` non-stream:**
```bash
  -H "Content-Type: application/json" -H "x-api-key: test" -H "anthropic-version: 2023-06-01" \
  -d '{"model": "'$MODEL'", "max_tokens": 50, "messages": [{"role": "user", "content": "Hi."}]}'
```
Returns a single `{"type": "text", "text": "<think>...raw incomplete thinking..."}` block instead of a `thinking` content block. The `stop_reason` also shows `"end_turn"` instead of `"max_tokens"`.

**Responses API `/v1/responses`:**
```bash
curl -s -X POST http://localhost:11434/v1/responses \
  -d '{"model": "'$MODEL'", "input": "What is 2+2?", "max_output_tokens": 200}'
```
Returns `{"type": "output_text", "text": "<think>...raw incomplete thinking..."}` instead of a `reasoning` block.

### What works correctly

All three endpoints handle thinking **correctly when `max_tokens` is large enough** to complete the think block — the Anthropic streaming endpoint, for example, correctly emits separate `thinking` and `text` content blocks in all cases.

## Root Cause

The `thinking_split.py` state machine waits for a `</think>` tag to close the thinking block before switching to text mode. When `max_tokens` terminates generation before `</think>` arrives:

1. The channel filter's `get_fallback_texts()` (or equivalent) returns the accumulated partial `<think>` content as plain text.
2. The non-stream aggregation path yields this as a text chunk to the response formatter.
3. The response formatter serializes it into the text/content field without recognizing it as thinking content.

The streaming path is partially correct (it streams `thinking_delta` events), but the non-stream path and the "max_tokens reached during thinking" case both fail.

## Environment

- Model: `mlx-community/Qwen3-4B-4bit:latest` (Qwen3 with native thinking)
- APIs affected: OpenAI `/v1/chat/completions`, Anthropic `/v1/messages`, Responses `/v1/responses` (all non-stream; also the case when streaming but max_tokens is hit during the think block)
- olmlx version: 0.1.0

dpalmqvist:
Closes #544

## Summary

When `max_tokens` halts generation while a thinking model is inside its `<think>` block, the unclosed opener and partial content leaked into the visible `content` field of all non-streaming endpoints (OpenAI, Anthropic, Responses). A parallel issue existed for Gemma4 channel blocks — a truncated block (no `<channel|>` closer) silently dropped the partial reasoning.

**Changes in `olmlx/engine/tool_parser.py`:**

- **`parse_model_output`**: After the existing `_THINK_RE`-based extraction and the orphan-`</think>` heuristic, detect an unclosed `<think>` opener (opener present, no matching `</think>` in remaining text). Reclassify the partial content as thinking; text before the opener is preserved as visible content.

- **`_extract_gemma4_blocks`**: When the `<channel|>` closer is absent, append the tail from the opener to end-of-string into `blocks` (thinking) instead of `break`-ing and silently discarding the content.

## Tests added (`tests/test_tool_parser.py` — `TestTruncatedThinking`)

| Test | What it checks |
|------|----------------|
| `test_unclosed_think_opener_classifies_as_thinking` | Raw `<think>` with no closer → partial reasoning goes to thinking, visible is empty |
| `test_unclosed_think_with_prior_content` | Preamble before opener preserved as visible; partial thinking classified correctly |
| `test_complete_pair_still_works` | Regression guard: complete `<think>…</think>` pairs still work |
| `test_truncated_gemma4_block_classifies_as_thinking` | Gemma4 opener without `<channel|>` closer → content routed to thinking |

- **No Metal-stream or inference-lock involvement** — this is pure text-parsing, executed after the generator is fully drained on the async event loop.
- The fix mirrors the contract already enforced on the streaming Anthropic path (`flush_split_thinking` maps `phase="in_think"` → `(buf, "")`, routing orphaned content to thinking and leaving visible empty).
- No router changes required; all three non-streaming surfaces already call `parse_model_output`.

### DIFF
--- a/olmlx/engine/tool_parser.py
+++ b/olmlx/engine/tool_parser.py
@@ -527,6 +527,10 @@ def _extract_gemma4_blocks(text: str) -> tuple[str, list[str]]:
 
         close = text.find("<channel|>", marker_end)
         if close < 0:
+            # Truncated block: stream ended before the closer was emitted.
+            tail = text[marker_end:].strip()
+            if tail:
+                blocks.append(tail)
             break
 
         content = text[marker_end:close].strip()
@@ -591,6 +595,16 @@ def parse_model_output(
                     else orphan_thinking
                 )
             text = text[orphan_idx + len("</think>") :].lstrip("\n")
+
+    # Detect truncated thinking block: opener present but stream ended before
+    # the matching </think> could be emitted (e.g. max_tokens hit mid-reasoning).
+    open_pos = text.find("<think>")
+    if open_pos != -1 and text.find("</think>", open_pos) == -1:
+        pre = text[:open_pos].strip()
+        partial = text[open_pos + len("<think>") :].strip()
+        if partial:
+            thinking = f"{thinking}\n{partial}".strip() if thinking else partial
+        text = pre
 
     tool_uses: list[dict] = []
     if has_tools:
--- a/tests/test_tool_parser.py
+++ b/tests/test_tool_parser.py
@@ -812,3 +812,35 @@ class TestGemma4Thinking:
         thinking, visible, tools = parse_model_output(text, has_tools=True)
         assert len(tools) == 1
         assert tools[0]["name"] == "func"
+
+
+class TestTruncatedThinking:
+    def test_unclosed_think_opener_classifies_as_thinking(self):
+        """max_tokens hits mid-reasoning: raw <think> must not leak into content."""
+        text = '<think>\nOkay, the user said "Say hi." I need to respond'
+        thinking, visible, tools = parse_model_output(text, has_tools=False)
+        assert "Okay, the user" in thinking
+        assert "<think>" not in visible
+        assert visible == ""
+
+    def test_unclosed_think_with_prior_content(self):
+        """Non-thinking preamble before the opener survives; partial thinking is
+        classified as thinking."""
+        text = "Sure! <think>Let me reason about"
+        thinking, visible, tools = parse_model_output(text, has_tools=False)
+        assert "Let me reason" in thinking
+        assert visible.strip() == "Sure!"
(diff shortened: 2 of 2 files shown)


dpalmqvist:
## Triage summary (factory/babysit)

**Fixed:** nothing — no CI failures or reviewer findings.

**Dismissed:** n/a — aider-code-review (gpt-review + deepseek-review) both returned clean; no inline comments to dismiss.

All checks green, FIX bucket empty — merging.

PR #556, opened 2026-06-23
fix(grammar): constrain json_object mode to JSON objects, not any JSON value

### LINKED ISSUE #550: response_format:{type:json_object} allows arrays and other non-object JSON — violates OpenAI spec
## Summary

`response_format: {"type": "json_object"}` in `/v1/chat/completions` and `/api/chat` with `format: "json"` does not constrain output to JSON objects. The grammar currently uses xgrammar's `compile_builtin_json_grammar()` which allows any well-formed JSON value (arrays, strings, numbers). The OpenAI spec requires `json_object` mode to guarantee the response is a JSON **object** (`{}`), not an array or other type.

## Reproduction

```bash
# Ask model to return a JSON array while json_object mode is active
  -d '{
    "messages": [{"role": "user", "content": "Return a JSON array of 3 numbers."}],
    "stream": false,
    "max_tokens": 100,
    "response_format": {"type": "json_object"}
  }'
```

**Response (actual):**
```json
{
  "choices": [{
    "message": {
      "content": "[\n  1,\n  2,\n  3\n]"
    }
  }]
}
```

Even without asking for an array, the model frequently produces them:
```bash
  -d '{"model": "mlx-community/Qwen3-4B-4bit", "messages": [{"role": "user", "content": "Return JSON with name and age."}], "response_format": {"type": "json_object"}, "max_tokens": 100, "stream": false}' \
  -H "Content-Type: application/json" | python3 -c "import sys,json; d=json.load(sys.stdin); c=json.loads(d['choices'][0]['message']['content']); print(type(c).__name__)"
# prints: list
```

## Root cause

`engine/grammar.py` line 221–222:
```python
if spec.kind == "json_object":
    compiled = compiler.compile_builtin_json_grammar()  # allows ANY JSON value
```

The docstring even acknowledges this: `"kind='json_object' allows any well-formed JSON value (xgrammar's builtin JSON grammar)"`.

## Expected behavior

Per the [OpenAI spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format): *"When using json_object mode … the model will always return a JSON object."*

`json_object` mode should use a grammar equivalent to `{"type": "object"}` JSON Schema (constraining the root to `{}`), not xgrammar's unrestricted JSON grammar. The fix would be:

```python
if spec.kind == "json_object":
    compiled = compiler.compile_json_schema({"type": "object"})
```

Note: `json_schema` mode already works correctly — `response_format: {"type": "json_schema", "json_schema": {"schema": {"type": "object", ...}}}` produces proper objects. Only the bare `json_object` mode is affected.

## Model

`mlx-community/Qwen3-4B-4bit` (reproduced consistently across multiple requests)

## Server version

`0.1.0`

dpalmqvist:
Closes #550

## Summary

- `response_format: {"type": "json_object"}` now calls `compiler.compile_json_schema({"type": "object"})` instead of `compiler.compile_builtin_json_grammar()`, which previously permitted arrays, strings, numbers, booleans, and null at the root — violating the OpenAI spec guarantee that `json_object` mode produces only a JSON object (`{}`).
- Updated the `GrammarSpec.kind` docstring to reflect the corrected semantics.
- Updated a misleading comment in `test_first_call_treats_tokens_as_prompt` that listed `[` as an allowed first token.

## Tests added

- **`TestLogitsProcessor::test_json_object_masks_array_start`** — after the prompt call, all `[`-containing tokens must be masked to `-inf`. Fails on `compile_builtin_json_grammar()`, passes after the fix.
- **`TestCompileCache::test_json_object_uses_object_schema_not_builtin_grammar`** — uses `xgr.GrammarMatcher` directly to assert that `[` is rejected as the first token by the compiled grammar. Same failure/pass pattern.

Both tests fail on current `main` and pass after the one-line change.

- **Grammar tokenizer identity** — `drop_for_tokenizer` and weakref identity validation are untouched; `cache_key()` for `json_object` remains the stable string `"json_object"`. An in-process server that already cached the old permissive grammar continues serving it until the next model unload, which is the documented bounded exposure window.
- No Metal stream, speculative decoding, prefill chunking, or distributed paths are involved in the grammar CPU-side token-mask path.

### DIFF
--- a/olmlx/engine/grammar.py
+++ b/olmlx/engine/grammar.py
@@ -53,9 +53,9 @@ _JSON_SCHEMA_TYPES = frozenset(
 class GrammarSpec:
     """Describes a grammar constraint for a single request.
 
-    ``kind="json_object"`` allows any well-formed JSON value (xgrammar's
-    builtin JSON grammar). ``kind="json_schema"`` requires ``schema`` to be
-    a JSON-Schema dict.
+    ``kind="json_object"`` constrains output to a JSON object (``{}``),
+    matching the OpenAI spec guarantee. ``kind="json_schema"`` requires
+    ``schema`` to be a JSON-Schema dict.
 
     Note: ``frozen=True`` would normally generate ``__hash__``, but the
     ``schema`` field is a dict and not hashable, so attempting to hash a
@@ -219,7 +219,7 @@ def compile_for_tokenizer(tokenizer: Any, vocab_size: int, spec: GrammarSpec) ->
     # Compile outside the lock — see docstring.
     compiler = _get_compiler(tokenizer, vocab_size)
     if spec.kind == "json_object":
-        compiled = compiler.compile_builtin_json_grammar()
+        compiled = compiler.compile_json_schema({"type": "object"})
     else:
         assert spec.schema is not None  # post_init enforces
         compiled = compiler.compile_json_schema(spec.schema)
--- a/tests/test_grammar.py
+++ b/tests/test_grammar.py
@@ -193,6 +193,29 @@ class TestCompileCache:
         c2 = compile_for_tokenizer(gpt2_tokenizer, gpt2_tokenizer.vocab_size, s2)
         assert c1 is not c2
 
+    def test_json_object_uses_object_schema_not_builtin_grammar(self, gpt2_tokenizer):
+        """Regression for #550: json_object mode must call compile_json_schema
+        with {"type": "object"}, NOT compile_builtin_json_grammar (which allows arrays)."""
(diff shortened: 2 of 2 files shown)


dpalmqvist:
## Factory triage summary

**Fixed:** nothing — no actionable findings from any reviewer.

**Dismissed:** none — the aider-code-review (GPT / DeepSeek) bots returned zero findings. No inline review threads exist.

All CI checks pass (Tests, Ruff, Pyright, CodeQL, gpt-review, deepseek-review). Merging.

Issue #513, opened 2026-06-13
Curated "known drafts" map for automatic speculative-draft resolution


dpalmqvist:
## Story

Ship a static table in the registry mapping `target hf_path → {strategy, draft_repo, block_size}`. When a user enables speculative on a known target *without* specifying a draft, olmlx fills the draft from the table. This is the difference between "the files exist somewhere" and "it just works."

## Why this is the biggest UX win

Today `models.json` is the only place to wire a draft, and it's hand-edited. With the curated map, `OLMLX_SPECULATIVE=true olmlx serve` on a supported model works with no repo ids to memorize.\n\n## Scope\n\n- [ ] Add the curated map (likely feeding `ModelConfig.resolved_speculative()` in `engine/registry.py`).\n- [ ] Resolution precedence: explicit per-model `speculative_draft_model` > global env > curated map > none.\n- [ ] Map entries reference the HF Hub repos from #512.\n- [ ] Record the validated quant per entry (ties into the compatibility matrix).\n- [ ] Tests for the resolution precedence.\n\n## Depends on\n\n- #512 (storage layer)

dpalmqvist:
Implemented in `bdb22bd` (#569) — curated known-drafts map for speculative resolution; `olmlx models pull --with-draft` co-download added in `f334e45` (#570).

Issue #512, opened 2026-06-13
Host olmlx-trained DFlash/EAGLE drafts on HF Hub under a project org


dpalmqvist:
## Story

Now that olmlx produces its own DFlash/EAGLE drafts, publish them to a project-owned HuggingFace org (e.g. `olmlx/` or `motsognirr/`), mirroring the z-lab naming convention the loader already understands: `olmlx/Qwen3-8B-DFlash`, `olmlx/Qwen3-8B-EAGLE`, etc.

## Why this is the baseline

`SpeculativeLoaderMixin._resolve_draft_path()` accepts exactly two things for `speculative_draft_model`: an HF repo id (→ `ModelStore.ensure_downloaded` → `snapshot_download`) or an absolute local path. Publishing to HF Hub requires **zero new code** — users set `speculative_draft_model: "olmlx/Qwen3-8B-DFlash"` and the existing download/cache/`.downloading`-marker machinery handles the rest. Free CDN + LFS, discoverable.

## Scope

- [ ] Create/decide the org namespace.
- [ ] Define the repo naming convention (`<target>-<STRATEGY>`), aligned with z-lab so both resolve identically.
- [ ] Publish the first batch of validated drafts.
- [ ] Each repo card records target build + validated quant (see compatibility issue).
- [ ] Document the manual `models.json` / env-var wiring in the user manual.

## Limitation

Pure hosting still leaves users hand-editing `models.json` with the repo id. Auto-resolution is the follow-up (curated map / co-download issues).

PR #570, opened 2026-06-25
feat(cli): `olmlx models pull --with-draft` co-downloads curated draft (#514)

### LINKED ISSUE #514: Co-download matching speculative draft at model pull time
## Story

When `olmlx models pull <target>` runs and a known draft exists for that target, optionally pull the matching draft too and write the `speculative_*` fields into that model's `models.json` entry automatically (via flag, e.g. `--with-draft`, or prompt).\n\n## Why\n\nProvisions a ready-to-spec-decode model in one command. The auto-registration path already exists in `ModelStore.pull()`, so this hooks into existing machinery.\n\n## Scope\n\n- [ ] Look up the target in the curated map (#513); if a draft exists, offer to pull it.\n- [ ] `--with-draft` flag (and an opt-out so users don't pull drafts they won't use).\n- [ ] After pull, write `speculative`, `speculative_strategy`, `speculative_draft_model` into the target's `models.json` entry.\n- [ ] Tests.\n\n## Depends on\n\n- #512 (storage), #513 (curated map for the lookup)

dpalmqvist:
Closes #514. Builds on the #513 curated known-drafts map.

## What

A `--with-draft` flag on `olmlx models pull <target>`. When set and a known speculative draft exists for the target (looked up in the #513 `KNOWN_DRAFTS` map), the CLI **also pulls the draft repo** and **writes the `speculative_*` fields into the target's `models.json` entry** — provisioning a ready-to-spec-decode model in one command.

```
olmlx models pull qwen3:32b --with-draft
```

## How

- `ModelStore.register_speculative_draft(name, hf_path, draft)` persists `speculative=True` + the draft's `strategy` / `draft_repo` / `block_size` onto the target's entry through the existing `add_mapping` path. It starts from the resolved entry (`dataclasses.replace`) so **existing per-model config is preserved**.
- `cmd_models_pull` reuses `store.pull()` for the draft download (same progress-printing style) after the target completes; when no draft is known for the target it prints a skip notice. The flag is **opt-in**, so opting out is simply omitting it (matches the issue's opt-out requirement).
- The trigger guard is `with_draft is True` — argparse `store_true` always yields a real `bool`, so this is correct in production while ignoring the truthy auto-attributes that bare-`MagicMock` args produce in unrelated pull tests.

## Interaction with #513's (currently empty) map

The curated map ships empty until the project-org draft repos land (#512), so `--with-draft` prints "no known speculative draft …; skipping" for real targets today. The mechanism is complete and fully tested via injected map entries; it activates automatically as `KNOWN_DRAFTS` is populated.

## Tests

- `tests/test_store.py::TestRegisterSpeculativeDraft` — writes the speculative fields and persists to disk; preserves existing per-model config (`keep_alive`); works for a full-`hf_path` target.
- `tests/test_cli.py::TestPullWithDraft` — `--with-draft` pulls the draft and calls register with the right args; no-known-draft skip; flag-off no-op.

All pass; ruff + pyright clean. Full `test_cli.py` / `test_store.py` / `test_known_drafts.py` green (246 tests).

> Note: issue #514 is labelled `plan:proposed`. Opening this for review; happy to hold for plan approval if preferred.

### DIFF
--- a/olmlx/cli.py
+++ b/olmlx/cli.py
@@ -1886,6 +1886,36 @@ def cmd_models_pull(args):
         async for status in store.pull(args.model_name):
             if msg := status.get("status", ""):
                 print(msg, flush=True)
+        # ``is True`` (not just truthiness): argparse store_true yields a real
+        # bool, so this stays correct while ignoring the truthy MagicMock attrs
+        # bare-mock CLI tests pass for unrelated pull cases.
+        if getattr(args, "with_draft", False) is True:
+            await _pull_draft()
+
+    async def _pull_draft():
+        """Co-download the curated speculative draft for the target (#514)."""
+        from olmlx.engine.registry import lookup_known_draft
+
+        resolved = store.registry.resolve(args.model_name)
+        hf_path = resolved.hf_path if resolved is not None else None
+        draft = lookup_known_draft(hf_path) if hf_path else None
+        if draft is None:
+            print(
+                f"--with-draft: no known speculative draft for "
+                f"{hf_path or args.model_name}; skipping.",
+                flush=True,
+            )
+            return
+        print(f"pulling speculative draft {draft.draft_repo}", flush=True)
+        async for status in store.pull(draft.draft_repo):
+            if msg := status.get("status", ""):
+                print(msg, flush=True)
+        store.register_speculative_draft(args.model_name, hf_path, draft)
+        print(
+            f"wired speculative draft {draft.draft_repo} "
+            f"(strategy={draft.strategy}) into config",
+            flush=True,
+        )
 
     try:
         asyncio.run(_pull())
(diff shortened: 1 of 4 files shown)


dpalmqvist:
Thanks — verified and addressed. To be precise: `ModelRegistry.resolve()` already passes full HF paths through verbatim (`org/model` → `org/model`), so untagged full-path pulls did resolve their `hf_path` correctly even without auto-registration. The real gap was **tagged** full paths (`org/model:q4`), which resolve with the tag attached and therefore miss the curated map's canonical (untagged) key. Pushed ece4cab to strip the Ollama tag before the lookup (mirroring `store.pull()`), with a regression test (`test_with_draft_strips_tag_on_full_path_target`).

PR #572, opened 2026-06-25
feat(adapters): LoRA / adapter hot-swap sharing base weights (#362)

### LINKED ISSUE #362: LoRA / adapter hot-swap
## Why

mlx-lm already supports adapters. Multiple adapters per base, sharing base weights in RAM and switched per request, is a common power-user pattern (different system prompts, fine-tunes, personalities) and currently requires loading the whole model multiple times.

## Scope

- Model addressing: `base:adapter` (e.g. `qwen3-8b:my-coder-lora`)
- `ModelStore` learns to fetch adapter weights and metadata
- `ModelManager` shares base weights across adapters loaded for the same base
- `models.json` gains an adapter section
- `olmlx models pull base:adapter` works
- All routers accept the `base:adapter` form

## Out of scope

- Stacked / merged adapters at runtime
- On-the-fly LoRA training (separate feature)

## Acceptance criteria

- Two adapters on the same base swap per request and share base weights in RAM (verify via memory snapshot)
- Both serve correctly from all routers
- `/api/ps` reports the adapter alongside the base

## Files

- `engine/model_manager.py`
- `models/store.py`, `models/manifest.py`
- `engine/registry.py`
- `routers/openai.py`, `routers/anthropic.py`, `routers/chat.py`, `routers/generate.py`, `routers/manage.py`

dpalmqvist:
Closes #362

Adds `base:adapter` model addressing so a base model's weights load once in RAM and multiple LoRA adapters layer on top, switched per request — implementing the approved plan, with one **deliberate correction** to its core mechanic (see below).

## What changed

- **`engine/registry.py`** — `AdapterConfig` dataclass; reserved `"adapters"` section in `models.json`; `is_adapter` / `resolve_adapter` / `list_adapters` / `add_adapter_mapping` (persisted, atomic, refuses to clobber an unreadable config); adapter-aware `remove`. `resolve()` is unchanged (adapters use dedicated accessors, mirroring the panel precedent).
- **`models/store.py`** — `adapter_local_path` / `is_adapter_downloaded` / `ensure_adapter_downloaded` / `pull_adapter` under a dedicated `adapters/` subtree; `pull` and `delete` dispatch on adapter names (so `olmlx models pull base:adapter` and `/api/pull` work with no router change).
- **`engine/model_manager.py`** — `structural_copy` weight-sharing helper; `_load_adapter_model`; base-pinning via `_adapter_child_refs` (LRU + expiry + unload guards, decremented at every `_loaded` removal); adapter entries reuse the base tokenizer, so `_close_loaded_model` skips the shared-tokenizer grammar drop; `_reject_adapter_base` rejects unsupported bases.
- **`/api/ps`** reports `adapter_base`; `ModelManifest` gains `adapter_base`.
- **Docs** — CLAUDE.md invariant, `models.json.example` adapters section, `max_loaded_models` help note.

## Plan correction (core mechanic)

The plan's `copy.copy(base.model)` is unsafe, and so is the obvious fallback — both verified empirically:
- `copy.copy` is **shallow**: it shares the `layers` list and every submodule, so `load_adapters` would mutate the **base** model.
- `copy.deepcopy` **duplicates** the `mx.array` weights *and* crashes on quantized models (`copy.deepcopy(mx.float16)` → "cannot pickle 'Dtype'").

`structural_copy` instead rebuilds the `Module`/list/dict tree with `mx.array` leaves and scalar attrs (`bits`, dtypes) shared by reference, copying the per-module `_no_grad` set so freezing one instance can't affect another. Confirmed on a real 4-bit model: shared weights (array identity + ~0 MB delta), base isolated through a real `load_adapters` call.

- **Grammar tokenizer identity** — adapters share the base's tokenizer object; the adapter close skips `drop_for_tokenizer` so it can't drop grammar entries the base/siblings still use. The base drops it once its last adapter is gone (kept pinned until then by `_adapter_child_refs`).
- **Metal stream hazard** — no new inference dispatch path; adapter inference rides `generate_chat` → `generation_stream` unchanged. The adapter build runs in a worker thread (like the base load) and eagerly `mx.eval`s the new LoRA params so no stream-bound lazy graph crosses back to the loop.
- **Loop affinity (#463)** — all `_adapter_child_refs` / `_loaded` mutations happen on the loop under `self._lock`.

## Tests

- `tests/test_adapter_loading.py` (**Metal-gated, `real_model`**) — end-to-end on a real model: two adapters share one base (weight-array identity + memory delta), serve correct logits, base-pinning + unload lifecycle. Passes locally.
- `tests/test_adapter_manager.py` — `structural_copy` correctness; base-pinning / child-ref accounting; LRU skips a pinned base; unload guards; grammar-drop skip; `_reject_adapter_base`.
- `tests/test_adapter_registry.py`, `tests/test_adapter_store.py`, `tests/test_adapter_ps.py` — parsing/persistence, download/pull dispatch, `/api/ps` surfacing.

`ruff check` + `ruff format --check` clean. Full `pytest -m "not real_model"` shows only failures that pre-exist on `main` (shardquant Metal kernels, one speech-router 500/400) — unrelated to this change.

## Note

LoRA adapters pin their base, so serving a base + N adapters needs `max_loaded_models ≥ N+1` (documented in config help).


dpalmqvist:
Thanks @github-actions (aider) — went through all 7. Fixed 3, pushing back on 4 with reasoning. (`31c0bc2`)

**Fixed**
- **#1 structural_copy recursion on non-tree graphs** — added a by-id memo: a submodule instance shared across layers is copied once and re-shared, and any back-reference/cycle resolves to the in-progress copy instead of recursing forever. Test: `test_shared_submodule_copied_once`.
- **#3 `size_bytes=0` undercounts** — adapters now report their LoRA-delta footprint (sum of `lora`-param bytes); base weights stay shared (already counted on the base). Asserted in the Metal-gated test (`0 < adapter.size_bytes < base.size_bytes`).
- **#4 batching flags copied from base** — agreed the structurally modified (LoRALinear) model isn't validated for continuous batching. Adapters now set `batching=False` and serve via the per-request path (out of scope for #362).

**Pushing back (already handled)**
- **#2 pin leak if `_reject_adapter_base` raises** — `_reject_adapter_base(base_lm)` is *inside* the `try`, and `finally: base_lm.release_ref()` runs on any raise (including from reject). No leak.
- **#5 child-ref drifts negative / wrong base** — `_detach_adapter_locked` guards `base._adapter_child_refs > 0` before decrementing (no underflow), and `adapter_base` is set once at creation and immutable, so it can't target the wrong base. An entry is popped from `_loaded` exactly once, so no double-decrement.
- **#6 `.downloading` marker not cleaned on failure** — this mirrors `ensure_downloaded` deliberately: the marker is removed on success (renamed to `.downloading.failed` only if `unlink` itself fails), and left in place on a download exception so `snapshot_download` resumes on the next attempt. Re-entering the download path after a failure is the intended self-healing behavior.
- **#7 `delete()` leaves the registry mapping** — the `/api/delete` router calls `registry.remove()` (now adapter-aware: pops `_adapters` + persists) right after `store.delete()`, exactly as it does for base models. `store.delete()` is filesystem-only by design; config/FS stay in sync via the router.

dpalmqvist:
Addressed the `structural_copy` tuple nit in `f245177`: the tuple branch now registers its built result in the memo, so a tuple shared from multiple places is deduped (test `test_shared_tuple_deduped`), and the misleading inline comment is corrected.

On the crash claim specifically: infinite recursion through a tuple can't actually occur — tuples are immutable, so any reference cycle that reaches a tuple must pass through a mutable container (list/dict/Module), and those are pre-registered in the memo before recursing, which terminates the walk. The memo on the tuple is for shared-tuple dedup, not recursion safety. Either way it's now uniform.

olmlx/models/manifest.py
```diff
@@ -17,6 +17,9 @@ class ModelManifest:
     family: str = ""
     parameter_size: str = ""
     quantization_level: str = ""
+    #: Base model name when this manifest describes a LoRA adapter (issue #362);
+    #: empty for ordinary base models. Default preserves existing manifests.
+    adapter_base: str = ""
```
dpalmqvist: Removed in 2822861 (YAGNI) — adapters have no manifest.json and /api/ps reads adapter_base from the LoadedModel, so the field was dead.

PR #573, opened 2026-06-25
feat(prompt_cache): disk spill for the VLM prompt cache (#491)

### LINKED ISSUE #491: Disk spill for VLM prompt cache
Extracted from the #487 triage.

The VLM prompt cache (#429, `VlmPromptCacheStore`) is in-memory only: per-`cache_id` LRU with `OLMLX_VLM_PROMPT_CACHE_SLOTS` (default 2). Known v1 limits: no radix takeover, no disk spill, no KV-quant, not counted against `OLMLX_MEMORY_LIMIT_FRACTION`.

This issue covers adding disk spill for cold VLM entries, mirroring the text-path design (`prompt_cache/` disk spill with `asyncio.to_thread` offload, plus the #466 re-spill guard).

Open questions:
- mlx_vlm's `PromptCacheState` serializability — does it have a safetensors-compatible form, or does it carry non-array state (image hashes, prefix metadata) needing a sidecar?
- Whether spill is worth it at 2 slots — the common single-conversation case is fully covered in memory.

Low urgency. Long-term.

dpalmqvist:
Closes #491.

## What

The VLM prompt cache (`VlmPromptCacheStore`) was **in-memory only**: an entry evicted past `OLMLX_VLM_PROMPT_CACHE_SLOTS` (default 2) was lost, forcing a full re-prefill on the next request for that `cache_id`. This adds a **cold disk tier**, mirroring the text-path design (`prompt_cache/store.py`).

## Why it's a clean mirror

The key open question from the issue was `PromptCacheState` serializability. It turns out to carry **only** `cache` (per-layer mlx-lm KV) and `token_ids` — no image hashes / rope deltas as separate fields. So it serializes exactly like the text path:
- `cache` → `mlx_lm.save_prompt_cache` / `load_prompt_cache` (safetensors)
- `token_ids` → safetensors metadata sidecar

## How

- New `async_get` / `async_insert` offload disk I/O with `asyncio.to_thread`. The KV state is **eager-`mx.eval`'d on the loop thread** before crossing into the worker thread, so no lazy Metal-stream graph escapes (#284).
- A **byte-capped, oldest-mtime** cleanup bounds disk use.
- Disk save is **best-effort** — a non-serializable cache or full disk never fails the in-flight request.
- **Never-filled states** (`cache=None`, inserted empty and evicted before mlx_vlm fills them) are skipped.
- `_setup_vlm_prompt_cache` is now `async` to await the disk tier; both call sites (`_stream_completion`, `_full_completion`) are already async.
- The in-memory `get`/`insert` API is unchanged (existing callers/tests unaffected); `insert` just routes through an evictee-capturing helper.

## Config (opt-in, mirrors the text knobs)

`OLMLX_VLM_PROMPT_CACHE_DISK` (default **off** — the common single-conversation case is fully covered by the in-memory slots), `OLMLX_VLM_PROMPT_CACHE_DISK_PATH`, `OLMLX_VLM_PROMPT_CACHE_DISK_MAX_GB`.

## Tests (`tests/test_vlm_prompt_cache_store.py`)

No real VLM required — a real `KVCache` filled with small arrays exercises the actual `save_prompt_cache`/`load_prompt_cache` round-trip:
- KV round-trips (restored `cache[0].offset` survives), spill-on-eviction, restore-on-memory-miss, memory-hit-skips-disk, empty-state skip, byte-cap cleanup, and disk-disabled no-op.

Verified locally on Apple Silicon: VLM store + inference cache + config + model_manager + prompt_cache suites green.

## Scope note

This is the v1 disk tier (cache_id-keyed, no radix-takeover, no KV-quant), matching the issue. Cross-request integration is exercised by the async unit tests; the end-to-end win is realised when a real VLM evicts and later restores a conversation across the configured slot count.

### DIFF
--- a/olmlx/config.py
+++ b/olmlx/config.py
@@ -122,8 +122,15 @@ class Settings(BaseSettings):
     prompt_cache: bool = True
     # Number of per-cache_id VLM KV-cache slots retained for cross-turn image-prefix
     # reuse (mlx_vlm PromptCacheState). 0 disables VLM prompt caching entirely.
-    # In-memory only (no disk spill / radix / KV-quant); 2 slots bound the memory.
+    # 2 slots bound the in-memory tier; cold entries spill to disk when
+    # vlm_prompt_cache_disk is on (#491).
     vlm_prompt_cache_slots: int = 2
+    # Disk spill for the VLM prompt cache (#491): mirrors the text-path
+    # prompt_cache_disk_* knobs. Off by default — the common single-conversation
+    # case is fully covered by the in-memory slots.
+    vlm_prompt_cache_disk: bool = False
+    vlm_prompt_cache_disk_path: Path = Path.home() / ".olmlx" / "cache" / "vlm"
+    vlm_prompt_cache_disk_max_gb: Annotated[float, Field(gt=0)] = 10.0
     prompt_cache_max_tokens: Annotated[int, Field(gt=0)] | None = 32768
     prompt_cache_max_slots: Annotated[int, Field(gt=0)] = 4
     prompt_cache_disk: bool = False
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -1961,7 +1961,7 @@ async def _setup_via_checkpoint_path(
     )
 
 
-def _setup_vlm_prompt_cache(
+async def _setup_vlm_prompt_cache(
     lm: LoadedModel,
     prompt_tokens: list[int] | None,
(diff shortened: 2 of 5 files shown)


dpalmqvist:
Good call — added `assert_loop_thread("VlmPromptCacheStore._spill")` so the eager-eval-on-loop-thread invariant is enforced consistently with get/insert/_insert_capture (commit pushed).

PR #574, opened 2026-06-25
feat(engine): StreamingLLM sink+window KV eviction for full-attention models (#505)

### LINKED ISSUE #505: StreamingLLM-style sink+window KV eviction for non-shard models
From a performance brainstorm.

ShardKVCache already keeps the first 4 (sink) + last 64 (window) tokens FP16-exact and compresses the middle — but that is shard-specific and quantization-based (bounds *bytes per token*).

**Idea:** a general StreamingLLM/attention-sink **eviction** policy as a standalone KV layout for any model: keep a few sink tokens + a sliding recent window, drop the middle entirely. This bounds KV **count** (and thus per-step attention compute + memory), a different lever than the three KV-quant schemes which bound bytes-per-token but keep all tokens. Useful for very long contexts where exactness on the middle isn't needed.

Notes:
- Distinct from pure-RotatingKVCache SWA models (gpt-oss/Step-3.5) which are *architecturally* windowed — this would apply a window to full-attention models as a runtime policy.
- Trim/prompt-cache-reuse semantics need care (evicted middle breaks exact prefix reuse).
- Opt-in per request/model; lossy, so behind a flag like the KV-quant knobs.

Worth it for long-context chat where decode slows as KV grows. Medium priority.

dpalmqvist:
Closes #505.

## What

An opt-in KV layout that bounds KV **count** (per-step attention compute + memory) for long contexts — a different lever than the three KV-quant schemes, which bound *bytes-per-token* but keep every token. Useful for long-context chat where decode slows as the cache grows.

## The key realization

mlx-lm's `RotatingKVCache(max_size=sink+window, keep=sink)` **is** exactly a StreamingLLM sink+window cache: it retains the first `sink` (attention-sink) tokens plus a rotating recent `window` and drops the middle, and `create_attention_mask` already delegates to its `make_mask`. So applying it to a full-attention model as a runtime policy needs **no custom rope/mask work** — it reuses mlx-lm's tested implementation. (Verified: `qwen3.py` etc. build the per-forward mask via `create_attention_mask(h, cache[0])`, which dispatches to `RotatingKVCache.make_mask`.)

## How

- `OLMLX_KV_EVICTION="<sink>:<window>"` (e.g. `"4:512"`) — global + per-model override (`resolved_kv_eviction`), mirroring `kv_cache_quant` end-to-end (validator, `Settings`, `ModelConfig`, `from_entry`, `LoadedModel`).
- Applied in `_make_prompt_cache_for_lm` **only to pure full-attention models** (default cache is all plain `KVCache`). Hybrid / SWA / GDN models are left untouched and logged: one per-forward mask is built from `cache[0]`, so a mixed list would mis-mask, and recurrent (`ArraysCache`) layers can't be windowed.
- **Mutually exclusive with `kv_cache_quant`** (quant bounds bytes, eviction bounds count) — quant wins, enforced both at load and in the cache builder.
- Lossy and opt-in: exact prefix reuse degrades once eviction starts (`RotatingKVCache.is_trimmable` → `False`), which the existing prompt-cache reuse path already handles.

## Tests — `tests/test_kv_eviction.py`

- Config validation + resolution precedence + `from_entry`.
- Cache selection: pure full-attention converted to `RotatingKVCache(max_size, keep)`; hybrid left untouched; quant takes precedence.
- Eviction policy at the cache level: **no-eviction exactness** — `RotatingKVCache(max_size ≥ total)` returns byte-identical K/V to a plain `KVCache` (so model output is unchanged until eviction kicks in) — and KV-count bounding under sustained decode.

Verified locally on Apple Silicon: 21/21 new + config/registry/model_manager/prompt_cache suites green (350 passed). Ruff + pyright clean.

## Scope / verification note

Correctness of the masking/rope is mlx-lm's `RotatingKVCache` (battle-tested for SWA models); this PR is the opt-in plumbing to apply it as a runtime policy on full-attention models. The no-eviction path is exactness-tested. The lossy long-context **quality** (the point of the feature) is inherent to dropping the middle and is best evaluated with a long-context eval on a real model — not asserted here, since it's an approximation, not a correctness property.

### DIFF
--- a/olmlx/config.py
+++ b/olmlx/config.py
@@ -80,6 +80,34 @@ def validate_kv_cache_quant_format(v: str | None) -> str | None:
     return v
 
 
+def validate_kv_eviction_format(v: str | None) -> str | None:
+    """Validate a ``kv_eviction`` value of the form ``'<sink>:<window>'`` (#505).
+
+    StreamingLLM-style eviction keeps the first ``sink`` tokens (attention
+    sinks) plus the last ``window`` tokens, dropping the middle. ``sink >= 0``,
+    ``window >= 1``. Validated for both the env var and ``models.json`` paths.
+    """
+    if v is None:
+        return v
+    parts = v.split(":", 1)
+    if len(parts) != 2:
+        raise ValueError(
+            f"Invalid kv_eviction={v!r}. Expected '<sink>:<window>' (e.g. '4:512')."
+        )
+    try:
+        sink, window = int(parts[0]), int(parts[1])
+    except ValueError:
+        raise ValueError(
+            f"Invalid kv_eviction={v!r}. Both sink and window must be integers "
+            f"(e.g. '4:512')."
+        ) from None
+    if sink < 0 or window < 1:
+        raise ValueError(
+            f"Invalid kv_eviction={v!r}. Require sink >= 0 and window >= 1."
+        )
+    return v
+
+
 class Settings(BaseSettings):
     # Note: ``validate_assignment=True`` applies to *all* fields, not just
     # the new speculative ones. Programmatic writes that previously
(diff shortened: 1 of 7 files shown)


dpalmqvist:
Good catch — fixed. `kv_eviction` is now serialized in `to_entry` (and included in the all-defaults bare-string short-circuit), so a per-model setting survives a models.json re-save. Added a round-trip test (`test_to_entry_round_trips`).

PR #580, opened 2026-06-26
fix(responses): stream /v1/responses in real time (#547)

### LINKED ISSUE #547: /v1/responses streaming buffers entire generation — zero TTFT benefit vs non-streaming
## Summary

`POST /v1/responses?stream=true` (OpenAI Responses API) does not actually stream. It buffers the entire model output before sending any SSE events, so the first event (`response.created`) arrives only after generation is complete. Clients see the same time-to-first-token (TTFT) whether they use streaming or not.

## Reproduction

**Model:** `mlx-community/Qwen3-4B-4bit:latest`

```python
import http.client, json, time

conn = http.client.HTTPConnection("localhost", 11434)
conn.request("POST", "/v1/responses", 
    body=json.dumps({
        "model": "mlx-community/Qwen3-4B-4bit:latest",
        "input": "/no_think Count from 1 to 20",
        "stream": True,
        "max_output_tokens": 200
    }),
    headers={"Content-Type": "application/json"})

start = time.time()
resp = conn.getresponse()
events = []
buf = b""
while True:
    chunk = resp.read(1)
    if not chunk:
        break
    buf += chunk
    if buf.endswith(b"\n\n"):
        block = buf.strip().decode()
        buf = b""
        for line in block.split('\n'):
            if line.startswith('event:'):
                events.append((time.time() - start, line.strip()))

for t, ev in events:
    print(f"{t:.3f}s  {ev}")
```

**Output (all events arrive in one burst after ~1s of generation):**
```
0.981s  event: response.created
0.981s  event: response.in_progress
0.981s  event: response.output_item.added
0.982s  event: response.content_part.added
0.982s  event: response.output_text.delta
0.982s  event: response.output_text.done
0.983s  event: response.content_part.done
0.983s  event: response.output_item.done
0.983s  event: response.completed
```

All 9 events arrive within **2ms** of each other, ~1 second after the request was made. Compare with `/v1/chat/completions` streaming which sends the first token in ~189ms and streams at ~11ms/token.

Additionally, `response.created` incorrectly includes the **final** output token count (not 0 as per spec):
```json
{"type": "response.created", "response": {"usage": {"output_tokens": 145, ...}}}
```

## Root Cause

`olmlx/routers/responses.py`, function `_stream_response` (line 340–):

```python
async def _stream_response(result, ...):
    """Buffer the engine output, parse once, replay full semantic events."""
    full_text = ""
    stats = None
    async for chunk in result:          # ← consumes ALL tokens first
        if chunk.get("done"):
            stats = chunk.get("stats")  # ← stats has final counts
            break
        full_text += chunk.get("text", "")
    
    # parse everything, then send events
    in_progress_resp = base_response("in_progress", [], stats, None)  # stats already final
    yield ev("response.created", {"response": in_progress_resp})      # ← emitted AFTER all tokens
```

The docstring `"Buffer the engine output, parse once, replay full semantic events"` confirms this is intentional, but it makes streaming semantically equivalent to non-streaming.

## Expected

- `response.created` should emit within ~50ms of the request, before generation starts, with `usage.output_tokens: 0`
- `response.output_text.delta` events should carry individual tokens as they're generated
- `response.reasoning.delta` (or equivalent) events should stream thinking tokens during the thinking phase

## Observed

- First event arrives after the entire generation completes (~1s for 20 tokens, scales linearly)
- All events arrive within 2ms of each other
- `response.created` contains the final token count instead of 0

## Impact

Any OpenAI-compatible client using the Responses API with streaming (`stream: True`) gets zero benefit from streaming — UX shows the same latency as non-streaming. For long generations (300+ tokens) this is a 3–5 second blank wait before anything displays.

dpalmqvist:
Closes #547.

## Problem

`POST /v1/responses?stream=true` did not actually stream. `_stream_response` drained the entire `generate_chat` async iterator before emitting any SSE event, so all events arrived in a burst after generation completed — identical TTFT to non-streaming. Secondary bug: `response.created` carried the final `output_tokens` count instead of `0`, since stats were only available after all tokens.

## Fix

`_stream_response` now routes by tool presence:

- **Text path** (no tools, new) — emits `response.created`/`response.in_progress` with zero usage **before** consuming any token, then routes each engine chunk through the shared `split_thinking_parts` state machine (same one the Anthropic/Ollama routers use), emitting one `response.output_text.delta` per content fragment and a reasoning item for thinking. A pure- or truncated-thinking response closes its reasoning item before `response.completed`, and never opens an empty message item.
- **Tools path** (unchanged) — the existing buffered logic, moved verbatim into a helper. `parse_model_output` needs the full brace-balanced output to detect tool calls and assign upfront tool IDs, so streaming is impossible there.

No CLAUDE.md Metal-stream / prefill / speculative invariants apply — this is pure async SSE routing with no MLX calls. The `thinking_expected` engine meta chunk is forwarded into the splitter state so a non-thinking model mentioning a literal `</think>` isn't misclassified.

## Known limitation

Thinking-detection's `detect` phase still buffers the first ≤200 bytes of a non-thinking response before the first content delta (inherent to the shared splitter — same behavior as the Anthropic and Ollama routers). Responses longer than that — including the issue's 145-token repro — stream incrementally once past the detect window.

## Tests (TDD)

Added to `TestStreaming`, written failing first:

- `test_response_created_has_zero_output_tokens` — `response.created` usage is 0; final event carries the real count.
- `test_per_chunk_content_deltas` — multi-chunk content yields one delta per chunk (old path emitted a single combined delta).
- `test_thinking_item_closes_before_message_opens` — reasoning `output_item.done` precedes message `output_item.added`.
- `test_truncated_thinking_flushes_reasoning_item` — mid-`<think>` truncation still closes the reasoning item, opens no message item, and reports `incomplete`.

Full suite: 707 passed, 12 skipped across the router/streaming tests; `ruff check` + `ruff format` clean.

### DIFF
--- a/olmlx/routers/responses.py
+++ b/olmlx/routers/responses.py
@@ -16,6 +16,7 @@ from olmlx.engine.tool_parser import (
     resolve_tool_names,
 )
 from olmlx.routers.common import build_inference_options, resolve_openai_think
+from olmlx.routers.thinking_split import flush_split_thinking, split_thinking_parts
 from olmlx.schemas.responses import ResponsesRequest, ResponsesResponse
 from olmlx.utils.images import normalize_image_block
 
@@ -345,7 +346,16 @@ async def _stream_response(
     tools: list[dict] | None,
     conversation: list[dict],
 ):
-    """Buffer the engine output, parse once, replay full semantic events."""
+    """Emit Responses SSE events from the engine stream.
+
+    Routes by whether tools are requested.  The **text path**
+    (``_stream_text_response``) emits ``response.created`` immediately — before
+    consuming any token — and one ``response.output_text.delta`` per engine
+    chunk, so clients see real time-to-first-token (issue #547).  The **tools
+    path** (``_stream_tools_response``) must buffer the whole generation:
+    ``parse_model_output`` needs the full, brace-balanced output to detect tool
+    calls and assign upfront tool IDs, so streaming is impossible there.
+    """
     seq = 0
 
     def ev(event: str, data: dict) -> str:
@@ -361,7 +371,8 @@ async def _stream_response(
         obj["status"] = status
         return obj
 
-    try:
+    async def _stream_tools_response():
+        """Buffer the engine output, parse once, replay full semantic events."""
         full_text = ""
         raw_text = ""
         done_reason = None
@@ -388,7 +399,7 @@ async def _stream_response(
         fill_missing_required_args(tool_uses, tools)
         output_items = _build_output_items(thinking, visible_text, tool_uses)
 
-        in_progress_resp = base_response("in_progress", [], stats, None)
+        in_progress_resp = base_response("in_progress", [], None, None)
         yield ev("response.created", {"response": in_progress_resp})
         yield ev("response.in_progress", {"response": in_progress_resp})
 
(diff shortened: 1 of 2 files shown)


olmlx/routers/responses.py
```diff
@@ -467,6 +478,207 @@ def base_response(status: str, output_items: list[dict], stats, done_reason):
...
+                    reasoning_open = True
+                    reasoning_index = 0
+                    events.append(
+                        ev(
+                            "response.output_item.added",
+                            {"output_index": reasoning_index, "item": reasoning_item()},
+                        )
+                    )
+                thinking_text += fragment
+                return events
```
dpalmqvist: Intentional, and leaving as-is. On a streaming surface `output_item.added` carries empty content and the text streams afterward — the message item's `added` event does the same here. The full text lands on `output_item.done` and in `response.completed`. Full content on `added` is specific to the buffered tools path; mirroring it would be less correct for the streaming convention, not more.

olmlx/routers/responses.py
```diff
@@ -467,6 +478,207 @@ def base_response(status: str, output_items: list[dict], stats, done_reason):
...
+                        "logprobs": [],
+                    },
+                )
+            )
+            return events
+
+        initial = base_response("in_progress", [], None, None)
+        yield ev("response.created", {"response": initial})
+        yield ev("response.in_progress", {"response": initial})
+
```
dpalmqvist, line 630: Fixed in ca0b32f — both streaming loops now check `done` before `thinking_expected` so a terminal chunk carrying both keys can't have its stats/done_reason dropped. Added a regression test (`test_done_honored_when_combined_with_thinking_expected`).

olmlx/routers/responses.py
```diff
@@ -467,6 +480,209 @@ def base_response(status: str, output_items: list[dict], stats, done_reason):
...
+                "summary": [],
+                "content": content,
+            }
+
+        def message_item(status: str) -> dict:
+            return {
+                "type": "message",
+                "id": message_id,
+                "status": status,
+                "role": "assistant",
```
dpalmqvist, line 523: Addressed in 2b80382 by streaming reasoning properly rather than front-loading content on `added`. Thinking now emits `response.reasoning_text.delta` per fragment plus a `response.reasoning_text.done` with the full text (before the item's `output_item.done`), mirroring how visible content streams via `output_text.delta`. This fulfills issue #547's request for token-level reasoning streaming, so clients render thinking incrementally instead of waiting for `done`.

PR #581, opened 2026-06-26
fix(engine): propagate finish_reason=length on non-streaming exclusive path (#565)

### LINKED ISSUE #565: Engine: _full_completion_inner drops GenerationResponse.finish_reason — all non-streaming exclusive-path endpoints report wrong stop reason when max_tokens hit
## Summary

`_full_completion_inner` (the non-streaming, non-batched code path used when `OLMLX_BATCHING=false`, the default) never reads `GenerationResponse.finish_reason` from `mlx_lm.stream_generate`. As a result, the result dict returned to every non-streaming router never contains `done_reason = \"length\"`, so all three API surfaces (OpenAI, Anthropic, Ollama) incorrectly report a natural-stop reason even when generation was truncated by `max_tokens`.

## Affected surfaces

| API | Wrong value | Correct value |
|-----|-------------|---------------|
| OpenAI `/v1/chat/completions` (non-stream) | `finish_reason: \"stop\"` | `finish_reason: \"length\"` |
| Anthropic `/v1/messages` (non-stream) | `stop_reason: \"end_turn\"` | `stop_reason: \"max_tokens\"` |
| Ollama `/api/chat` (non-stream) | `done_reason: \"stop\"` | `done_reason: \"length\"` |

The **streaming** path and the **batched** non-streaming path (`OLMLX_BATCHING=true`) are unaffected — both correctly report the stop reason.

## Reproduction

Model: `mlx-community/Qwen3-4B-4bit` (thinking model; 200 tokens exhausted by `<think>` block)

```bash
# OpenAI non-streaming
  -d '{
    "messages": [{"role": "user", "content": "What is 17 × 43 + 82 ÷ 2 − 15? Show detailed step-by-step reasoning."}],
    "max_tokens": 200,
    "stream": false
  }' | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['choices'][0]['finish_reason'], d['usage']['completion_tokens'])"
```

**Actual:** `stop 200`  
**Expected:** `length 200`

```bash
# Anthropic non-streaming
  -H "x-api-key: test" -H "anthropic-version: 2023-06-01" \
  -d '{"model": "mlx-community/Qwen3-4B-4bit", "max_tokens": 50,
       "messages": [{"role": "user", "content": "Tell me about Paris"}]}' \
  | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['stop_reason'], d['usage']['output_tokens'])"
```

**Actual:** `end_turn 50`  
**Expected:** `max_tokens 50`

Note: streaming correctly returns `finish_reason: \"length\"` / `stop_reason: \"max_tokens\"` for the same requests.

## Root cause

`mlx_lm.stream_generate` yields `GenerationResponse` dataclass objects. The final response has:

```
finish_reason: Optional[str] = "stop" | "length"
```

`_full_completion_inner` (in `engine/inference.py`) accumulates text from every response but reads only `prompt_tokens` and `generation_tokens` from the final one. It never reads `finish_reason`. The result dict it returns (`{"text": text, "done": True, "stats": stats}`) therefore never contains `done_reason`.

The routers have correct logic — for example in `routers/openai.py` lines 463–466:

```python
elif done_reason in ("timeout", "length"):
    finish_reason = "length"
else:
    finish_reason = "stop"   # ← always taken because done_reason is None
```

But `done_reason` is always `None` from this path, so the fallback fires unconditionally.

## Fix

In `_full_completion_inner`, after extracting the last `GenerationResponse` into `result`, add:

```python
# Propagate stop reason from mlx-lm
finish_reason = getattr(result, "finish_reason", None)
if finish_reason == "length":
    result_dict["done_reason"] = "length"
```

This requires no changes to the routers.

## Related issues

- #558 — OpenAI non-streaming omits `content` field when thinking exhausts budget.  The implementation plan in that issue claims `finish_reason` is "already handled correctly", but it is not: this engine bug is the reason `finish_reason: \"stop\"` appears instead of `\"length\"` in the observed response.  The plan's mock-based tests pass only because they stub `done_reason = \"length\"` directly on the mock return value — bypassing the missing engine propagation.

dpalmqvist:
Fixes #565.

## Problem

`_full_completion_inner` (the default non-streaming, non-batched path, `OLMLX_BATCHING=false`) accumulated text and token counts from mlx-lm's final `GenerationResponse` but never read its `finish_reason`. The returned result dict therefore lacked `done_reason`, so all three router surfaces fell back to a natural-stop reason even when generation was truncated by `max_tokens`:

| API | Reported (wrong) | Correct |
|-----|------------------|---------|
| OpenAI `/v1/chat/completions` | `finish_reason: "stop"` | `length` |
| Anthropic `/v1/messages` | `stop_reason: "end_turn"` | `max_tokens` |
| Ollama `/api/chat` | `done_reason: "stop"` | `length` |

The streaming path and the batched non-streaming path already report the stop reason correctly.

## Fix

Read `finish_reason` off the final result and forward `"length"` as `done_reason`; the routers already map it. `getattr(..., None)` keeps it safe across the `stream_generate` / speculative / mlx_vlm sub-paths. No router or schema changes.

## Tests (TDD)

Added three methods to `TestFullCompletionInner`:
- `test_length_finish_reason_propagated` — failed before the fix (`KeyError: 'done_reason'`), passes after
- `test_stop_finish_reason_not_propagated` — `"stop"` adds no key
- `test_none_finish_reason_not_propagated` — missing attribute adds no key

All 275 tests in `tests/test_inference.py` pass; ruff lint + format clean.

### DIFF
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -4213,6 +4213,11 @@ async def _full_completion_inner(
                 text = visible
 
         result_dict: dict = {"text": text, "done": True, "stats": stats}
+        # Propagate the stop reason from mlx-lm's final GenerationResponse so the
+        # non-streaming routers can report max_tokens truncation. getattr(..., None)
+        # is safe across all sub-paths (stream_generate / speculative / mlx_vlm).
+        if getattr(result, "finish_reason", None) == "length":
+            result_dict["done_reason"] = "length"
         if raw_text is not None:
             result_dict["raw_text"] = raw_text
         if tool_uses:
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -3954,6 +3954,11 @@ async def _full_completion(
         text, hit = truncate_at_stop(result_dict.get("text", ""), stop_sequences)
         if hit:
             result_dict["finish_reason"] = "stop"
+            # The stop sequence is applied post-hoc, so mlx-lm may have run on to
+            # max_tokens and set done_reason="length". A stop-sequence hit means
+            # the visible generation ended at the stop, so "stop" wins — drop the
+            # length marker that routers key on.
+            result_dict.pop("done_reason", None)
         result_dict["text"] = text
     return result_dict
 
--- a/tests/test_inference.py
+++ b/tests/test_inference.py
@@ -577,6 +577,64 @@ class TestStopSequenceHandling:
     @patch("olmlx.engine.inference._inference_locked")
     @patch("olmlx.engine.inference._inference_ref")
     @patch("olmlx.engine.inference._full_completion_inner")
+    async def test_stop_sequence_hit_overrides_length_done_reason(
+        self, mock_inner, mock_ref, mock_locked
+    ):
+        """A post-hoc stop-sequence hit means the user-visible generation ended
+        at the stop, so it must win over a mlx-lm done_reason='length' (max_tokens
+        reached after the stop sequence was already emitted). Otherwise routers
+        keyed on done_reason would report 'length' for a stop-truncated response."""
+        gen_kwargs = {"stop": ["D"]}
+        stats = MagicMock()
+        lm = MagicMock()
+        lm.inference_queue_timeout = 30.0
+        lm.sync_mode = None
+
+        mock_locked.return_value.__aenter__.return_value = None
+        mock_ref.return_value.__enter__.return_value = None
+        mock_inner.return_value = {
+            "text": "A B C D E F G",
+            "done": True,
+            "stats": stats,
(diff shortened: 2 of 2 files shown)


dpalmqvist:
Good catch — addressed in cd5608e. Since stop sequences are applied post-hoc in `_full_completion` (popped from gen_kwargs, never passed to mlx-lm), a stop hit can co-occur with mlx-lm's `done_reason="length"`. On a stop hit we now drop the length marker so "stop" wins; genuine max_tokens truncation (no stop match) keeps `done_reason="length"`. Added two tests covering both directions of the interaction.

PR #584, opened 2026-06-26
fix(api): require max_tokens on /v1/messages; provider-format validation errors (#578, #579)

### LINKED ISSUE #578: Anthropic `/v1/messages` accepts requests without required `max_tokens` field (returns 200 instead of 400)
## Summary

The Anthropic `/v1/messages` endpoint silently accepts requests that omit the required `max_tokens` field, responding HTTP 200 with a default token limit. The real Anthropic API returns HTTP 400 with `{"type": "error", "error": {"type": "invalid_request_error", "message": "max_tokens: field required"}}` for such requests.

## Reproduction

**Model:** `mlx-community/LFM2-1.2B-4bit`

```bash
curl -sv -X POST http://127.0.0.1:11434/v1/messages \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "model": "mlx-community/LFM2-1.2B-4bit",
    "messages": [{"role": "user", "content": "hi"}]
  }'
```

## Observed

```
HTTP/1.1 200 OK
{"id":"msg_...","type":"message","role":"assistant","content":[{"type":"text","text":"Hello! How can I assist you today?"}],...}
```

## Expected (per Anthropic API spec)

```
HTTP/1.1 400 Bad Request
{"type":"error","error":{"type":"invalid_request_error","message":"max_tokens: field required"}}
```

## Root cause

`schemas/anthropic.py` line 74:

```python
max_tokens: int = Field(4096, ge=1)
```

The field has a default value of `4096`, making it optional in Pydantic. Per the Anthropic API specification, `max_tokens` is a **required** field. Removing the default would cause Pydantic to reject requests without it with a 422, and the exception handler could then map that to the correct Anthropic 400 error format.

## Impact

Code targeting the real Anthropic API that omits `max_tokens` by mistake will get a clear error there but silently succeed here with a potentially-surprising default limit. This difference in behaviour makes olmlx a less faithful drop-in for Anthropic API clients during development/testing.

### LINKED ISSUE #579: Validation errors on OpenAI/Anthropic endpoints return raw Pydantic 422 format instead of provider-specific error format
## Summary

When a request body fails Pydantic validation (e.g., empty `messages` array, invalid `n`, missing required field), the server returns HTTP 422 with a raw Pydantic `{"detail": [...]}` response body. OpenAI SDK clients and Anthropic SDK clients expect provider-specific error envelopes — `{"error": {"message": ..., "type": ..., "code": ...}}` for OpenAI and `{"type": "error", "error": {...}}` for Anthropic. Client SDKs that parse `error.message` will raise unexpected errors or fail silently.

## Reproduction

**OpenAI endpoint — empty messages array:**

```bash
curl -sv -X POST http://127.0.0.1:11434/v1/chat/completions \
  -d '{"model": "mlx-community/LFM2-1.2B-4bit", "messages": []}'
```

**Observed:**
```
HTTP/1.1 422 Unprocessable Entity
{"detail":[{"type":"value_error","loc":["body","messages"],"msg":"Value error, messages cannot be empty","input":[],"ctx":{"error":{}}}]}
```

**Expected (per OpenAI API spec):**
```json
{
  "error": {
    "message": "messages cannot be empty",
    "type": "invalid_request_error",
    "code": "invalid_request_error"
  }
}
```

**Other affected inputs:** `n > 1`, malformed JSON body, `max_completion_tokens < 1`, `temperature > 2`.

**Anthropic endpoint — missing max_tokens (once #578 is fixed):**

The same 422 Pydantic format would appear, whereas Anthropic clients expect:
```json
{"type": "error", "error": {"type": "invalid_request_error", "message": "..."}}
```

## Root cause

FastAPI's default 422 handler is not overridden for provider-specific routes. Application-level `ValueError` handlers (which DO format correctly as `{"error": "..."}` for Ollama, `{"error": {...}}` for OpenAI) only fire for exceptions raised inside the route handler, not for Pydantic model validation errors that prevent the route from being entered.

Fix: register a `RequestValidationError` exception handler in `app.py` that inspects the request path prefix and formats the error body accordingly (`/v1/chat`, `/v1/completions` → OpenAI format; `/v1/messages` → Anthropic format; `/api/` → Ollama format).

## Impact

Any Python, TypeScript, or other SDK that does `error.message` / `error["message"]` on a 422 response will get `KeyError` / `TypeError` instead of a descriptive error string. This breaks standard SDK error handling for invalid inputs.

dpalmqvist:
Fixes #578. Fixes #579.

These two issues share a single root fix — a `RequestValidationError` handler — so they're resolved together rather than artificially split.

## #578 — Anthropic `/v1/messages` must require `max_tokens`

`max_tokens` carried a default of `4096`, so omitting it silently succeeded (200) instead of returning a 400 like the real Anthropic API. Changed to `Field(..., ge=1)` (required).

The real Anthropic `count_tokens` endpoint takes **no** `max_tokens` field, so making it required everywhere would make olmlx *less* faithful there. Instead, `/v1/messages/count_tokens` now uses a dedicated `AnthropicCountTokensRequest` subclass that keeps `max_tokens` optional while inheriting all other fields/validators.

## #579 — validation errors must use the provider error envelope

Request-body Pydantic failures returned FastAPI's raw `422 {"detail": [...]}` body, which OpenAI/Anthropic SDK clients can't parse (`error.message` → KeyError). Added a `RequestValidationError` handler in `app.py` that:
- flattens the Pydantic errors into one message (`max_tokens: field required`, `messages cannot be empty`, etc.),
- returns **400** in the provider envelope dispatched by request path via the existing `_make_error_response` helper (Anthropic `{"type":"error","error":{...}}`, OpenAI `{"error":{...}}`, Ollama `{"error": ...}`).

In-handler `HTTPException(422)` paths (e.g. an unrecognized `format` value) are deliberately unaffected — only Pydantic body-validation is reshaped.

## Tests (TDD)

- New: `test_missing_max_tokens_returns_400`, `test_count_tokens_without_max_tokens_succeeds`, schema `test_messages_request_max_tokens_required` / `test_count_tokens_request_max_tokens_optional`, app-level `test_validation_error_anthropic_format` / `test_validation_error_openai_format`.
- Updated existing 422-expecting tests to 400 + provider envelope across OpenAI, Ollama (chat/generate/embed/models), agent, rerank, and edge-case surfaces.
- Full `pytest -m "not real_model"` suite: the only remaining failures are pre-existing, environment-specific Metal-kernel (`test_shardquant_*`) and a TTS handler test — all confirmed to fail identically on a clean tree without this diff.

## Invariants

No CLAUDE.md Metal-stream / prefill / speculative / distributed invariants touched — entirely in the HTTP request-validation layer.

### DIFF
--- a/olmlx/app.py
+++ b/olmlx/app.py
@@ -7,6 +7,7 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 
 from fastapi import FastAPI, Request
+from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from fastapi.staticfiles import StaticFiles
@@ -320,6 +321,41 @@ def create_app() -> FastAPI:
     # Added last → runs outermost, so the per-request root span wraps every
     # other middleware and all downstream engine spans nest under it.
     app.add_middleware(RootSpanMiddleware)
+
+    @app.exception_handler(RequestValidationError)
+    async def request_validation_error_handler(
+        request: Request, exc: RequestValidationError
+    ):
+        # FastAPI's default handler returns a raw 422 {"detail": [...]} body.
+        # Provider SDKs expect provider-shaped error envelopes, so translate the
+        # Pydantic errors into a single 400 message dispatched by request path.
+        parts: list[str] = []
+        for err in exc.errors():
+            loc = err.get("loc") or ()
+            field = str(loc[-1]) if loc else "body"
+            msg = str(err.get("msg", "invalid value"))
+            if err.get("type") == "missing":
+                parts.append(f"{field}: field required")
+            elif msg.startswith("Value error, "):
+                # A custom field-validator already produced a descriptive
+                # message; surface it verbatim without a redundant field prefix.
+                parts.append(msg[len("Value error, ") :])
+            elif len(loc) > 1:
+                # Generic constraint/type error (e.g. "Input should be ...") —
+                # prefix the field name so the client knows what was invalid.
+                parts.append(f"{field}: {msg}")
+            else:
+                parts.append(msg)
+        message = "; ".join(parts) if parts else "invalid request"
+        logger.warning("Request validation error on %s: %s", request.url.path, message)
+        return _make_error_response(
+            request.url.path,
+            400,
+            message,
+            "invalid_request_error",
+            "invalid_request_error",
+            "invalid_value",
+        )
 
     @app.exception_handler(ValueError)
     async def value_error_handler(request: Request, exc: ValueError):
--- a/olmlx/routers/anthropic.py
+++ b/olmlx/routers/anthropic.py
@@ -30,6 +30,7 @@ from olmlx.engine.tool_parser import (
 )
 from olmlx.schemas.anthropic import (
     AnthropicContentBlock,
+    AnthropicCountTokensRequest,
     AnthropicMessagesRequest,
     AnthropicMessagesResponse,
     AnthropicTokenCountResponse,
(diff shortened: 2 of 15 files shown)


dpalmqvist:
Re: the aider-review note on `AnthropicCountTokensRequest` losing the parent's `max_tokens` global-limit validator — verified this is not the case. Pydantic v2 inherits `@field_validator` registrations by field name and still applies them when a subclass redefines the field. Empirically both classes reject `max_tokens=1e9` via `validate_token_limit`, so the docstring's "inheriting every other field and validator" claim holds.

PR #585, opened 2026-06-26
fix(anthropic): no spurious thinking block for whitespace-only fragments (#557)

### LINKED ISSUE #557: Anthropic streaming emits spurious thinking block with /no_think (streaming/non-streaming inconsistency)
## Summary

When `stream: true`, the Anthropic `/v1/messages` endpoint emits a **spurious `thinking` content block** even when the user prompt contains `/no_think` (Qwen3's thinking-suppression token). The non-streaming path correctly returns only a `text` block. This is a streaming-vs-non-streaming inconsistency.

## Reproduction

Model: `mlx-community/Qwen3-4B-4bit`

```bash
# Streaming — emits thinking block (WRONG)
  -H "x-api-key: test" \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "messages": [{"role": "user", "content": "Say: yes. /no_think"}],
    "max_tokens": 20,
    "stream": true
  }' | grep "content_block_start"
```

**Actual output (streaming):**
```
data: {"type": "content_block_start", "index": 0, "content_block": {"type": "thinking", "thinking": ""}}
data: {"type": "content_block_start", "index": 1, "content_block": {"type": "text", "text": ""}}
```

Two content blocks: a `thinking` block (with delta content `\n\n`) followed by a `text` block (`Yes.`).

```bash
# Non-streaming — one text block (CORRECT)
  -H "x-api-key: test" \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "messages": [{"role": "user", "content": "Say: yes. /no_think"}],
    "max_tokens": 20,
    "stream": false
  }'
```

**Actual output (non-streaming):**
```json
{"content": [{"type": "text", "text": "Yes."}], "stop_reason": "end_turn", ...}
```

One text block only — correct.

## Root Cause

With `/no_think`, Qwen3 outputs `</think>` as an orphan close tag at the start of its response (after leading whitespace from the chat template), followed by the actual response text. The incremental `split_thinking_parts` state machine in `_stream_thinking_state_machine` (`routers/anthropic.py`) detects this orphan close and classifies the preceding `\n\n` as thinking content (per the orphan-close heuristic for `thinking_expected=True` models). Since `"\n\n"` is truthy, the guard in `_route()`:

```python
if not fragment:
    return []
```

does **not** filter it out, so a `thinking` content block is opened and the `\n\n` is emitted.

The **non-streaming** path receives already-processed text from `generate_chat(stream=False)` — the engine strips the thinking portion internally and returns `thinking=None` — so `parse_model_output` never sees the leading `\n\n` and correctly returns no thinking content.

## Expected Behaviour

With `/no_think`, the Anthropic streaming response should emit **only a `text` content block**, matching the non-streaming path. The empty/whitespace-only `\n\n` fragment should not trigger a `thinking` block.

## Fix Hint

In `_route()` (`routers/anthropic.py`), extend the guard from `if not fragment` to `if not fragment.strip()` so whitespace-only thinking fragments are dropped rather than creating a thinking block. Mirror any corresponding fix in `_emit_content_block` for the buffered tool path.

## Server Log / Observed Values

Streaming: content block types observed = `['thinking', 'text']`  
Non-streaming: content block types observed = `[('text', "'Yes.'")]`

Reproduced consistently across multiple runs.

dpalmqvist:
Fixes #557.

## Problem

On the Anthropic `/v1/messages` **streaming** path, a request with `/no_think` emitted a spurious `thinking` content block. Qwen3 emits leading whitespace (`\n\n`) before an orphan `</think>`; the thinking splitter classifies that whitespace as a thinking fragment, and `_route` opened a thinking content block for it because `"\n\n"` is truthy. The **non-streaming** path correctly emits only a `text` block — so this was a streaming/non-streaming inconsistency.

## Fix

- `_route`: skip a whitespace-only fragment when it would **open** a new block (`current is None`). Whitespace *inside* an already-open block is preserved, so legitimate inter-token spacing in real thinking/text content is untouched. This is narrower (and safer) than the issue's suggested blanket `if not fragment.strip()`, which would have dropped inter-word whitespace in streamed text.
- Buffered tool path: mirror the guard — `if thinking` → `if thinking and thinking.strip()`.

## Tests (TDD)

- New `test_streaming_no_think_whitespace_no_spurious_thinking_block` — reproduces the `\n\n</think>Yes.` scenario; asserts no `thinking` content block, empty thinking text, and `text == "Yes."`. Failed before the fix (`['thinking', 'text']`), passes after.
- Full `tests/test_routers_anthropic.py` suite passes (128 tests).

## Invariants

No CLAUDE.md invariants touched — purely SSE block-emission logic; no inference, stream selection, or cache plumbing.

### DIFF
--- a/olmlx/routers/anthropic.py
+++ b/olmlx/routers/anthropic.py
@@ -532,13 +532,14 @@ async def _stream_thinking_state_machine(result):
         nonlocal current, text_block_emitted
         if not fragment:
             return []
-        # A whitespace-only fragment must not *open* a new block. With
-        # `/no_think`, Qwen3 emits leading whitespace before an orphan
+        # A whitespace-only *thinking* fragment must not open a thinking block.
+        # With `/no_think`, Qwen3 emits leading whitespace before an orphan
         # `</think>`, which the splitter classifies as a thinking fragment;
-        # opening a thinking block for it produces a spurious block the
-        # non-streaming path never emits (issue #557). Whitespace *inside* an
-        # already-open block is preserved.
-        if current is None and not fragment.strip():
+        # opening a block for it produces a spurious thinking block the
+        # non-streaming path never emits (issue #557). Scoped to the thinking
+        # channel so leading whitespace in plain-text streams is preserved, and
+        # to block-opening so whitespace inside an open thinking block survives.
+        if channel == "thinking" and current is None and not fragment.strip():
             return []
         block_type = "thinking" if channel == "thinking" else "text"
         events = []
--- a/olmlx/routers/anthropic.py
+++ b/olmlx/routers/anthropic.py
@@ -420,7 +420,9 @@ async def _stream_buffered_with_tools(
 
     block_idx = 0
 
-    if thinking:
+    if thinking and thinking.strip():
+        # Mirror the incremental path (issue #557): a whitespace-only thinking
+        # span must not produce a thinking content block.
         for event in _emit_content_block(
             block_idx,
             "thinking",
@@ -529,6 +531,14 @@ async def _stream_thinking_state_machine(result):
         """
         nonlocal current, text_block_emitted
         if not fragment:
+            return []
+        # A whitespace-only fragment must not *open* a new block. With
+        # `/no_think`, Qwen3 emits leading whitespace before an orphan
+        # `</think>`, which the splitter classifies as a thinking fragment;
+        # opening a thinking block for it produces a spurious block the
+        # non-streaming path never emits (issue #557). Whitespace *inside* an
+        # already-open block is preserved.
+        if current is None and not fragment.strip():
             return []
         block_type = "thinking" if channel == "thinking" else "text"
         events = []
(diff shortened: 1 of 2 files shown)


dpalmqvist:
Thanks — good catch. Narrowed the guard to `channel == "thinking"` in 2bbd00c so leading whitespace in plain-text streams is preserved; the bug is specific to the orphan `</think>` on the thinking channel.

dpalmqvist:
On the divergence concern: the guard can't drop leading whitespace from *genuine* orphan thinking. The orphan-close branch in `split_thinking_parts` (`thinking_split.py:144`) emits the **entire** pre-`</think>` prefix as a single `("thinking", buf[:close_idx])` fragment — text and whitespace together. So a pure-whitespace thinking fragment reaches `_route` with `current is None` *only* when the whole prefix is whitespace, i.e. the `/no_think` case. Genuine orphan thinking always carries non-whitespace in that first fragment, so `.strip()` is truthy, the block opens, and its internal/leading whitespace is preserved. Subsequent in-`<think>` fragments arrive with `current == "thinking"`, where the guard doesn't apply. Verified against the full suite incl. `test_streaming_qwen35_long_orphaned_thinking`.

PR #590, opened 2026-06-27
fix(engine): stop sequences ignore matches inside <think> blocks (#588)

### LINKED ISSUE #588: Stop sequences fire inside <think> block — generation halts before producing any visible content
## Summary

When a thinking-capable model (e.g. Qwen3) is used with a `stop` sequence, the stop sequence is matched against the **entire** model output including the `<think>...</think>` block. If the stop string appears inside the thinking block, generation is cut off before any visible text is produced — yielding `content: null` (OpenAI) or empty `content: ""` (Ollama) with `finish_reason: stop` even though the model never reached the reply section.

## Reproduction

**Model:** `mlx-community/Qwen3-4B-4bit`  
Stop sequence `"three"` appears naturally in the thinking block ("…one two three…") before the model has produced any visible text.

```bash
# OpenAI API (non-streaming) — returns null content
  -d '{
    "messages": [{"role": "user", "content": "Say: one two three four five"}],
    "max_tokens": 500,
    "stop": ["three"],
    "stream": false
  }' | python3 -m json.tool
```

**Actual response:**
```json
{
  "choices": [{
    "message": {"role": "assistant", "content": null},
    "finish_reason": "stop"
  }],
  "usage": {"completion_tokens": 13, ...}
}
```

Only 13 tokens were generated — the stop string was matched inside the thinking block before any `</think>` was produced.

```bash
# Ollama API with think=true — confirms thinking cut off
curl -s -X POST http://localhost:11434/api/chat \
  -d '{
    "messages": [{"role": "user", "content": "Say: one two three four five"}],
    "stream": false,
    "think": true,
    "options": {"num_predict": 500, "stop": ["three"]}
  }' | python3 -c "
import sys, json; d=json.load(sys.stdin)
print('done_reason:', d['done_reason'])
print('thinking:', repr(d['message'].get('thinking','')[:100]))
print('content:', repr(d['message']['content']))
print('eval_count:', d['eval_count'])
"
```

**Actual output:**
```
done_reason: stop
thinking: 'Okay, the user wants me to say a sequence of words: "one", "two", "'
content: ''
eval_count: 13
```

Thinking was cut off at 13 tokens, `"three"` was matched inside `<think>`.

## Expected Behaviour

Stop sequences must only be applied to **visible** output (tokens after `</think>`). If the stop string appears only inside the thinking block and not in visible text, generation should continue until:
- The stop string appears in visible text → `finish_reason: stop`
- `max_tokens` is exhausted → `finish_reason: length`
- The model reaches EOS → `finish_reason: stop`

**Contrast with thinking disabled** (same stop string, same prompt):
```bash
curl -s -X POST http://localhost:11434/api/chat \
  -d '{
    "messages": [{"role": "user", "content": "Say: one two three four five"}],
    "stream": false,
    "think": false,
    "options": {"num_predict": 200, "stop": ["three"]}
  }' | python3 -c "
import sys,json; d=json.load(sys.stdin)
print('done_reason:', d['done_reason'])
print('content:', repr(d['message']['content']))
"
```
```
done_reason: stop
content: 'One, two,'
```

With `think: false`, the stop correctly fires at "three" in the visible output after generating the preceding words.

## Impact

Any use of stop sequences with thinking models silently produces empty responses. This is a correctness regression that affects all clients passing `stop` to Qwen3 or other thinking-enabled models via any API surface (Ollama, OpenAI, Anthropic).

## Location

The stop sequence matching is applied in the mlx-lm generation loop against the raw token stream (which includes `<think>` tokens). The fix should defer stop-sequence evaluation until after the `</think>` marker has been seen (i.e., after `engine/thinking_split.py` or an equivalent state machine has determined the token is in the visible region).

## Server log excerpt (relevant entries)

```
INFO  Generation complete: 1 prompt tokens, 13 tokens generated (...)
```

dpalmqvist:
Fixes #588.

## Problem

Stop sequences were matched against the **entire** raw model output, including `<think>...</think>` content. When the stop string appeared inside the reasoning block (e.g. `stop=["three"]` with a prompt that makes the model count in its thinking), generation halted before producing any visible text — yielding `content: null` (OpenAI) / `content: ""` (Ollama) with `finish_reason: stop`, on every API surface, for any thinking model.

## Fix

Make both stop helpers in `engine/stop_sequences.py` thinking-aware (opt-in via `thinking_aware=True`):

- **`StopScanner`** tracks the `detect → in_think → passthrough` phase incrementally (bounded to the boundary window, still O(piece)) and honours a stop match only when it lies in visible text — before a `<think>` open tag or after its `</think>` close. Matches inside the thinking block are skipped.
- **`truncate_at_stop`** finds the first `<think>` span and ignores matches within it.

`engine/inference.py` threads a `thinking_expected` flag (already computed by `_resolve_thinking_active`) from `generate_completion`/`generate_chat` through the streaming, batched-streaming, batched non-streaming, and exclusive non-streaming paths to the three scanner/truncator construction sites. `thinking_aware` defaults `False`, so non-thinking callers and existing tests are unchanged.

`_THINKING_PAIRS` is duplicated from `routers/thinking_split.py` (importing it would create a circular import via `inference.py`); a guard test asserts the two constants stay in sync.

## Tests (TDD)

- New `TestStopScannerThinkingAware` / `TestTruncateAtStopThinkingAware` in `tests/test_stop_sequences.py` — cover stop-inside-think (skipped), stop-in-visible (fires), pre-think, no-think (identical to non-aware), piece-boundary spans, the Gemma `<|channel>` format, and open+close in one piece.
- New `_full_completion` integration tests in `tests/test_inference.py::TestStopSequenceHandling` — stop inside think isn't truncated; stop in visible text still fires.
- `_THINKING_PAIRS` cross-module sync guard.
- Broader sweep (stop_sequences, inference, finish_reason, openai/chat/generate/anthropic routers, batching): 679 pass.

## Invariants

No CLAUDE.md invariants touched — pure Python string scanning; no Metal/GPU, prefill, grammar, or speculative code. The new threading is keyword-only with a `False` default.

### DIFF
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -1633,6 +1633,7 @@ async def generate_completion(
                 keep_alive=keep_alive,
                 grammar_active=grammar_active,
                 adopt_pin=True,
+                thinking_expected=thinking_expected,
             )
             # Ownership of the pin transfers to the wrapper; its finally
             # releases on any generator exit. Mark the flag for the outer
@@ -1664,6 +1665,7 @@ async def generate_completion(
                         keep_alive=keep_alive,
                         grammar_active=grammar_active,
                         adopt_pin=True,
+                        thinking_expected=thinking_expected,
                     )
                 result["thinking_expected"] = thinking_expected
                 return result
@@ -2983,6 +2985,7 @@ async def _stream_completion_batched(
     prompt_tokens: list[int] | None = None,
     use_prompt_cache: bool = False,
     cache_id: str = "",
+    thinking_expected: bool = False,
 ) -> AsyncGenerator[dict, None]:
     """Batched counterpart of ``_stream_completion`` (plan §2/§4).
 
@@ -3036,7 +3039,7 @@ async def _stream_completion_batched(
         else settings.inference_timeout
     )
 
-    stop_scanner = StopScanner(stop_sequences)
+    stop_scanner = StopScanner(stop_sequences, thinking_aware=thinking_expected)
     stop_hit = False
     timed_out = False
     lagged = False
@@ -3257,6 +3260,7 @@ async def _stream_completion(
     messages: list[dict] | None = None,
     tokenizer: Any = None,
     template_kwargs: dict | None = None,
+    thinking_expected: bool = False,
 ) -> AsyncGenerator[dict, None]:
     # Continuous batching (docs/batching-plan.md): eligible requests join
     # the per-model batch instead of serializing on the inference lock.
@@ -3277,6 +3281,7 @@ async def _stream_completion(
             prompt_tokens=prompt_tokens,
             use_prompt_cache=use_prompt_cache,
             cache_id=cache_id,
+            thinking_expected=thinking_expected,
         ):
             yield chunk
         return
@@ -3512,7 +3517,9 @@ async def _stream_completion(
             with Timer() as eval_timer:
                 inf_start = time.monotonic()
                 token = None
-                stop_scanner = StopScanner(stop_sequences)
+                stop_scanner = StopScanner(
+                    stop_sequences, thinking_aware=thinking_expected
+                )
                 async for token in stream:
                     # Always accumulate for prompt cache (raw stream, not filtered)
                     stats.eval_count = token.generation_tokens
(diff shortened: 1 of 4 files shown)


dpalmqvist:
Good catch — fixed in cdff54a. The phase machine now cycles detect↔in_think (accumulating closed spans) instead of terminating in passthrough, and `truncate_at_stop` collects all think spans, so stop sequences are skipped inside *every* thinking block — matching `thinking_split.py`, which re-enters the thinking state on each open tag. Added `TestThinkingAwareMultipleBlocks` covering a stop in a second `<think>` block (scanner + whole-text).

PR #648, opened 2026-07-10
mlx-vlm 0.6.4 bump + gemma4 audio-tower sanitize workaround

dpalmqvist:
## Summary

Bumps mlx-vlm 0.5.0 → 0.6.4 and carries a workaround for a 0.6.x load regression that breaks every gemma-4 checkpoint with an audio tower.

### Bump

- `mlx-vlm>=0.6.4,<0.7`; resolves cleanly against the existing `mlx 0.32` / `mlx-lm 0.31.3` pins, and the locked transformers 5.11.0 fits mlx-vlm's new `<5.13` cap. Lock delta is mlx-vlm itself plus two small transitive deps (python-multipart, starlette).
- Audited the full `mlx_vlm` API surface olmlx uses — `load`, `stream_generate` (incl. `prompt_cache_state` / `logits_processors` kwargs), `apply_chat_template`, `generate.PromptCacheState`, `generate.generation_stream`, `utils.MODEL_REMAPPING`, the `models.<type>` package layout — all unchanged through 0.6.0's AR/diffusion-engine restructure. `stream_generate` still does not accept `prompt_progress_callback`, so streaming.py's VLM exclusion remains required.

### Regression workaround: `engine/gemma4_sanitize_fix.py`

mlx-vlm 0.6.0 removed the `is_mlx_format` guard around `sanitize_weights`, so gemma4's unconditional PyTorch→MLX conv transposes now also run on pre-converted mlx-community checkpoints — scrambling the audio-tower conv weights and failing the load:

```
Expected shape (128, 3, 3, 1) but received shape (128, 3, 1, 3) for
parameter audio_tower.subsample_conv_projection.layer0.conv.weight
```

A/B-verified: the same checkpoint loads fine under 0.5.0. Upstream main already guards the transposes by shape (`expected_in`), but that fix is unreleased as of 0.6.4.

The shim wraps `gemma4.Model.sanitize` to pre-invert already-MLX-layout conv weights (using upstream main's exact discriminant) so the following transpose restores them — correct under both the broken 0.6.4 sanitize and the fixed upstream one. Applied idempotently at every `mlx_vlm.load` site. `tests/test_gemma4_sanitize_fix.py::test_removal_gate_upstream_still_broken` fails loudly once a fixed mlx-vlm release lands, prompting removal (ropefix convention).

## Verification

- Full suite: 5501 passed, 40 skipped. (Two failures pre-exist on main and are unrelated — A/B-confirmed under 0.5.0: `test_self_attention_fused_sdpa_matches_reference` (borderline 1e-4 tolerance) and `test_speech_non_tts_model_400` (500 vs 400).)
- End-to-end on a live server with gemma-4-e4b-it-4bit: image chat answers correctly, streamed multi-turn works, VLM prompt-cache prefix reuse observed in logs.
- `tests/live/test_gemma4_unified_text.py` (real 12B model) passes — gemma4_unified keeps its pinned text-tower routing.

## Follow-ups (not in this PR)

- Report the sanitize regression upstream (Blaizzy/mlx-vlm) so the guard makes 0.6.5.
- mlx-vlm now ships a `gemma4_unified` module, so the text-only pin for those checkpoints could be relaxed to gain vision — needs separate verification.

### DIFF
--- a/olmlx/engine/gemma4_sanitize_fix.py
+++ b/olmlx/engine/gemma4_sanitize_fix.py
@@ -1,0 +1,113 @@
+"""Workaround for mlx-vlm 0.6.x scrambling gemma4 audio conv weights on
+pre-converted MLX checkpoints.
+
+mlx-vlm 0.5.0's ``load_model`` skipped ``sanitize_weights`` entirely for
+checkpoints whose safetensors carry the ``format: mlx`` metadata (the
+``is_mlx_format`` guard). mlx-vlm 0.6.0 removed that guard, so gemma4's
+``Model.sanitize`` conv transposes — PyTorch ``[out, in, kH, kW]`` ->
+MLX ``[out, kH, kW, in]`` — now also run on mlx-community checkpoints whose
+audio-tower conv weights are ALREADY in MLX layout, scrambling them:
+
+    Expected shape (128, 3, 3, 1) but received shape (128, 3, 1, 3) for
+    parameter audio_tower.subsample_conv_projection.layer0.conv.weight
+
+This breaks loading every gemma4 checkpoint that ships an audio tower
+(e2b/e4b/12B/26B/31b families). Upstream main has since added shape-based
+guards to the transposes (``expected_in``), but that fix is unreleased as
+of 0.6.4.
+
+The patch wraps ``Model.sanitize``: conv weights that are already in MLX
+layout (detected with upstream main's exact discriminant — last dim equals
+the layer's input channels) are pre-inverted to PyTorch layout so the
+wrapped sanitize's transpose restores them. This stays correct if the
+installed sanitize is the fixed (guarded) upstream version too: the
+pre-inverted PyTorch-layout weight fails the guard's already-MLX check and
+gets transposed back, so the patch never double-applies.
+
+Call ``ensure_gemma4_sanitize_patch()`` after ``import mlx_vlm`` and before
+``mlx_vlm.load(...)``.
+
+Remove once the pinned mlx-vlm release carries the upstream guard —
+``tests/test_gemma4_sanitize_fix.py::test_removal_gate_upstream_still_broken``
+fails loudly when that time comes.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Callable
+
+logger = logging.getLogger("olmlx.engine.gemma4_sanitize_fix")
+
+_PATCH_ATTR = "_olmlx_gemma4_sanitize_patch"
+_original: Callable[..., dict[str, Any]] | None = None
+
+
+def _pre_invert_mlx_layout(model_self: Any, weights: dict[str, Any]) -> dict[str, Any]:
+    """Return ``weights`` with already-MLX-layout audio conv weights
+    transposed to PyTorch layout, so the following unconditional sanitize
+    transpose lands them back in MLX layout."""
+    fixed: dict[str, Any] = {}
+    for k, v in weights.items():
+        if (
+            "subsample_conv_projection" in k
+            and "conv.weight" in k
+            and getattr(v, "ndim", 0) == 4
+        ):
+            # Upstream main's discriminant: MLX layout iff the last dim is
+            # the layer's input-channel count (1 for layer0, the previous
+            # layer's out-channels for layer1). kH/kW are small odd kernel
+            # sizes, so they can't collide with 128-ish channel counts.
+            expected_in = None
+            audio_config = getattr(model_self.config, "audio_config", None)
+            if audio_config is not None:
+                if ".layer0." in k:
+                    expected_in = 1
+                elif ".layer1." in k:
+                    expected_in = audio_config.subsampling_conv_channels[0]
+            if expected_in is not None and v.shape[-1] == expected_in:
+                # Inverse of sanitize's transpose(0, 2, 3, 1).
+                v = v.transpose(0, 3, 1, 2)
+        elif "depthwise_conv1d.weight" in k and getattr(v, "ndim", 0) == 3:
+            if v.shape[-1] == 1:
+                # Depthwise conv1d has in=1: MLX layout is [out, kW, 1].
+                # transpose(0, 2, 1) is its own inverse.
+                v = v.transpose(0, 2, 1)
+        fixed[k] = v
+    return fixed
+
+
+def ensure_gemma4_sanitize_patch() -> None:
+    """Idempotently patch ``mlx_vlm.models.gemma4.gemma4.Model.sanitize``.
+
+    Safe to call unconditionally before any ``mlx_vlm.load``; a no-op if
+    already applied or if the module layout has changed (the removal-gate
+    test catches the day this stops being needed).
+    """
+    global _original
+    try:
+        from mlx_vlm.models.gemma4.gemma4 import Model
+    except ImportError:  # pragma: no cover - future mlx-vlm restructure
+        logger.debug("mlx_vlm gemma4 module not found; sanitize patch skipped")
+        return
+
+    current = Model.sanitize
+    if getattr(current, _PATCH_ATTR, False):
+        return
+    _original = current
+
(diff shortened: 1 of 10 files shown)


olmlx/engine/gemma4_sanitize_fix.py
```diff
@@ -0,0 +1,113 @@
+"""Workaround for mlx-vlm 0.6.x scrambling gemma4 audio conv weights on
+pre-converted MLX checkpoints.
+
+mlx-vlm 0.5.0's ``load_model`` skipped ``sanitize_weights`` entirely for
+checkpoints whose safetensors carry the ``format: mlx`` metadata (the
+``is_mlx_format`` guard). mlx-vlm 0.6.0 removed that guard, so gemma4's
+``Model.sanitize`` conv transposes — PyTorch ``[out, in, kH, kW]`` ->
+MLX ``[out, kH, kW, in]`` — now also run on mlx-community checkpoints whose
```
dpalmqvist, line 8: **[review:correctness]** The root-cause claim is factually wrong: the `is_mlx_format` guard is present in the 0.6.0 and 0.6.3 wheels (verified) and was removed only in **0.6.4**. This docstring, the pyproject comment, and the test docstring all say 0.6.0/0.6.x, which (a) misleads anyone debugging 0.6.0–0.6.3 reports and (b) hides that pinning `<0.6.4` was a monkeypatch-free alternative — if 0.6.4 is chosen for its features, that tradeoff should be recorded.

olmlx/engine/gemma4_sanitize_fix.py
```diff
@@ -0,0 +1,113 @@
...
+        elif "depthwise_conv1d.weight" in k and getattr(v, "ndim", 0) == 3:
+            if v.shape[-1] == 1:
+                # Depthwise conv1d has in=1: MLX layout is [out, kW, 1].
+                # transpose(0, 2, 1) is its own inverse.
+                v = v.transpose(0, 2, 1)
+        fixed[k] = v
+    return fixed
+
+
+def ensure_gemma4_sanitize_patch() -> None:
```
dpalmqvist, line 94: **[review:coverage]** The removed guard was global, but this shim only covers gemma4. Verified in the installed 0.6.4: `qwen3_omni_moe/audio.py` transposes `audio_tower.conv2d*` weights unconditionally, `phi4mm/audio.py` transposes every 4-dim weight, `falcon_perception` re-applies a shape-preserving `.w13.` interleave (silent corruption — no shape error), and rfdetr/rt_detr_v2/sam3 have the same unguarded conv transposes. Pre-converted MLX checkpoints of those families loaded fine under 0.5.0 and now fail (or silently scramble) through the same generic mlx-vlm fallback paths. Worth documenting the scope limitation here so the next family failure is a fast diagnosis rather than a fresh investigation.

olmlx/engine/gemma4_sanitize_fix.py
```diff
@@ -0,0 +1,113 @@
...
+    logger.debug("Patched mlx_vlm gemma4 Model.sanitize (MLX-layout conv guard)")
+
+
+def _original_sanitize() -> Callable[..., dict[str, Any]]:
+    """The unpatched upstream sanitize (for the removal-gate test)."""
+    if _original is not None:
+        return _original
+    from mlx_vlm.models.gemma4.gemma4 import Model
+
+    return Model.sanitize
```
dpalmqvist: **[review:correctness]** `_original_sanitize()` returns the *patched wrapper* if this module is reloaded after the patch was applied: the early-return in `ensure_gemma4_sanitize_patch` sees `_PATCH_ATTR` and skips setting `_original` (fresh module global = None), so the fallback returns `Model.sanitize` — the wrapper. The removal-gate test then fails spuriously and instructs deleting a still-needed workaround. Reproduced with `importlib.reload`. Fix: stash the original callable on the wrapper itself (e.g. make `_PATCH_ATTR` carry it) instead of a module global.

olmlx/engine/gemma4_sanitize_fix.py
```diff
@@ -0,0 +1,113 @@
...
+    try:
+        from mlx_vlm.models.gemma4.gemma4 import Model
+    except ImportError:  # pragma: no cover - future mlx-vlm restructure
+        logger.debug("mlx_vlm gemma4 module not found; sanitize patch skipped")
+        return
+
+    current = Model.sanitize
+    if getattr(current, _PATCH_ATTR, False):
+        return
+    _original = current
```
dpalmqvist: **[review:robustness]** Only `ImportError` is caught. Importing `mlx_vlm.models.gemma4.gemma4` executes the gemma4 package `__init__` including transformers ProcessorMixin/image-processor class definitions — a class-creation-time `AttributeError`/`TypeError` from a future in-range transformers release would propagate out of `ensure_gemma4_sanitize_patch()` at all six call sites and break loads of unrelated non-gemma4 VLMs (which never imported gemma4 code before this PR). Catch broad `Exception` here — the patch is best-effort; a broken gemma4 import should degrade to gemma4-only breakage, not all-VLM breakage.

olmlx/engine/gemma4_sanitize_fix.py
```diff
@@ -0,0 +1,113 @@
...
+
```
dpalmqvist, line 119: **[review:robustness]** Silent double-degrade under the open `<0.7` pin: if a future 0.6.x moves/renames the gemma4 module without shipping the upstream guard, this logs only at DEBUG and no-ops, while the test file's `pytest.importorskip` makes the removal gate silently *skip* instead of fail. The protection machinery then degrades with no dev-time signal. Suggest logging at WARNING when the module is missing (an mlx-vlm in `<0.7` without `models.gemma4.gemma4` is surprising).

olmlx/engine/model_manager.py
```diff
@@ -2592,6 +2592,9 @@ def _try_lm_then_vlm(
             logger.warning("mlx-lm failed for %s (%s), trying mlx-vlm", label, exc)
             import mlx_vlm
 
+            from olmlx.engine.gemma4_sanitize_fix import ensure_gemma4_sanitize_patch
+
+            ensure_gemma4_sanitize_patch()
```
dpalmqvist: **[review:cleanup]** This `import mlx_vlm` + `ensure_gemma4_sanitize_patch()` pair is now copy-pasted at six call sites (3× model_manager, flash/prepare, flash_moe_model, spectralquant_calibrate). A seventh `mlx_vlm.load` site added later will forget the ritual and gemma4 audio checkpoints will fail only on that path. One chokepoint helper (`load_vlm(path, **kw)` that applies the patch and delegates to `mlx_vlm.load`) makes the pairing structural and turns the eventual workaround removal into a one-line edit.

olmlx/engine/gemma4_sanitize_fix.py
```diff
@@ -0,0 +1,136 @@
...
+        ):
+            # Upstream main's discriminant: MLX layout iff the last dim is
+            # the layer's input-channel count (1 for layer0, the previous
+            # layer's out-channels for layer1). kH/kW are small odd kernel
+            # sizes, so they can't collide with 128-ish channel counts.
+            expected_in = None
+            audio_config = getattr(model_self.config, "audio_config", None)
+            if audio_config is not None:
+                if ".layer0." in k:
+                    expected_in = 1
```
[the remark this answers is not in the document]
dpalmqvist: Verified against the installed mlx-vlm 0.6.4: `SubSampleConvProjection.__init__` hardcodes exactly `self.layer0` and `self.layer1` (`models/gemma4/audio.py:94-95`) — the subsampling layer count is structural, not derived from `subsampling_conv_channels`, so a `layer2` weight cannot exist in any checkpoint this mlx-vlm version can load. A 3-layer variant would require a new mlx-vlm release, at which point the removal-gate test forces this workaround to be revisited anyway. The hardcoded discriminant also deliberately mirrors upstream main's fixed sanitize verbatim (same `.layer0.`/`.layer1.` checks), which keeps the shim auditable against the upstream fix — generalizing here would be dead code that diverges from the reference implementation. Declining.

PR #649, opened 2026-07-10
feat(flash): trained expert lookahead for Flash-MoE — prefetch + scored eviction

dpalmqvist:
## Summary

Replaces pure-LRU expert caching on the Flash-MoE path with a trained-predictor pipeline: per-MoE-layer-pair low-rank lookahead heads predict the next MoE layer's expert set from the current layer's hidden state, driving (1) background SSD prefetch that overlaps expert loads with compute, and (2) predicted-need cache eviction so `cache_budget_experts` can shrink at equal hit rate.

Spec: `docs/superpowers/specs/2026-07-10-moe-expert-lookahead-design.md` · Plan: `docs/superpowers/plans/2026-07-10-moe-expert-lookahead.md`

- **`ScoredLayerCache`** (`_ssd_base.py`) — eviction victim = lowest predicted score among non-protected keys; plain LRU without scores; just-inserted key is victim of last resort, never protected keys.
- **`MoeLookaheadBank`** (`moe_predictor.py`) — SparsityPredictor-family heads over consecutive MoE layer pairs (non-contiguous indices supported); sidecar with trained-pair mask + training provenance; stale banks rejected at load.
- **`MoePrefetcher`** (`moe_prefetch.py`) — mirrors the dense Prefetcher's concurrency contract (single prediction thread owns the only background `mx.eval`); every failure path sets the done event, so `wait()` can never hang the forward pass.
- **`FlashMoE`** — load-bearing wait → eval(inds) → submit → load ordering; prefetch never affects outputs (verified by identical-outputs test).
- **Store** — `prefetch_experts` warms the cache off the existing I/O pool; scores are consume-once; in-flight request experts protected during the load only.
- **Config** — `OLMLX_FLASH_MOE_PREFETCH` / `_LOOKAHEAD_MARGIN` / `_PREFETCH_MAX_POSITIONS` / `_SCORED_EVICTION`; prefetch is forced off under `self_speculative` (early-exit draft loop would leave predictions unconsumed → concurrent `mx.eval`); calibration loaders default off.
- **Training** — `olmlx flash train-moe-lookahead` (or `flash prepare --train-lookahead`): records router traces via a Flash-MoE-wrapped forward pass, trains with recall-biased BCE, prints per-pair holdout recall@m.

## Test plan

- 130 new/updated tests across cache, predictor, store, prefetcher, forward integration, training pipeline, and CLI — including concurrency regression tests (score-push hang, protection bypass) found and fixed during review.
- Full suite at HEAD: 5515 passed / 40 skipped; 17 failures reproduce identically on the base commit (pre-existing shardquant/rerank numeric flakiness).
- Manual perf checklist (real SSD-bound MoE: hit-rate at reduced budget, decode tok/s, temperature-0 output identity) still to be run — see plan's "Post-plan verification".

### DIFF
--- a/olmlx/engine/flash/moe_prefetch.py
+++ b/olmlx/engine/flash/moe_prefetch.py
@@ -1,0 +1,177 @@
+"""Speculative expert prefetcher for Flash-MoE.
+
+While MoE layer L computes, a trained lookahead head predicts which experts
+the NEXT MoE layer will route to and starts background SSD reads, so the next
+layer's ``load_experts`` finds them cached. The same prediction's score
+vector steers cache eviction (``ScoredLayerCache``) when enabled.
+
+Concurrency contract (identical to the dense ``Prefetcher`` in
+``prefetch.py``): ``mx.eval`` is not safe for concurrent calls, so the single
+prediction thread is the ONLY background thread that evaluates arrays; the
+calling thread materializes the hidden state before enqueueing, and must not
+call ``mx.eval`` between ``submit()`` and the next ``wait()``. Prediction can
+never affect correctness — a misprediction or failure only costs latency.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+from concurrent.futures import ThreadPoolExecutor
+
+import mlx.core as mx
+
+from olmlx.engine.flash.moe_predictor import MoeLookaheadBank
+from olmlx.engine.flash.moe_weight_store import FlashMoeWeightStore
+from olmlx.engine.flash.prefetch import PrefetchStats, _LayerPrefetchState
+
+logger = logging.getLogger(__name__)
+
+
+class MoePrefetcher:
+    """Orchestrates background expert prefetching from SSD."""
+
+    def __init__(
+        self,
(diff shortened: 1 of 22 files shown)


olmlx/engine/flash/moe_prefetch.py
```diff
@@ -0,0 +1,185 @@
...
+        state = _LayerPrefetchState()
+        with self._lock:
+            self._pending[next_layer] = state
+        try:
+            self._predict_executor.submit(
+                self._do_predict_and_io, layer_idx, next_layer, hidden_state, state
+            )
+        except RuntimeError:  # executor shut down
+            with self._lock:
+                self._pending.pop(next_layer, None)
```
dpalmqvist, line 94: Not a race in practice: `submit()` is only ever called from the single forward-pass thread (`FlashMoE.__call__`), the same single-caller contract the dense `Prefetcher.submit` relies on — there is no concurrent submitter to interleave between the check and the registration. Documented the invariant in a comment at the check site (00dbe63) so this is explicit in the code rather than implicit.

PR #650, opened 2026-07-10
perf(flash): overhaul MoE expert prefetch; default it off

dpalmqvist:
## Summary

Benchmarked head-to-head on Qwen3.5-35B-A3B-4bit (32 GB M-series, cache budget 128 experts/layer), **every Flash-MoE expert-prefetch configuration lost 30–45% decode throughput to plain LRU caching** (best ~14.5 vs a reproducible 20.4 tok/s). This PR removes prefetch's structural overheads, fixes two real training bugs found during the investigation, adds the diagnostics that explain the result — and flips `flash_moe_prefetch` to **off by default**.

## Changes

- **Inline NumPy prediction** — `MoeLookaheadBank.predict_next_np` computes the lookahead head on the forward-pass thread in pure NumPy; the dedicated prediction thread and its background `mx.eval` are gone, eliminating a per-MoE-layer cross-thread rendezvous (measured ~2.5 tok/s).
- **Non-blocking prefetch** — `FlashMoE` no longer `wait()`s on prefetch I/O; `load_experts` races in-flight inserts (the store's `protect()` already supports this) and pending entries self-clear on completion (identity-guarded, so a racing resubmit's fresh state is never clobbered).
- **`--max-positions` on `flash train-moe-lookahead`** — the hidden 4096-position cap silently bounded every pair's training set regardless of `--samples`; raising it lifted holdout recall 0.17–0.54 → 0.30–0.53.
- **`--self-generate` decode tracing** — record traces from the model's own decode steps over chat prompts (ultrachat via dflash's `iter_dataset_prompts`) instead of teacher-forced prefill. Measured serve-time decode recall of the prefill-trained bank was **0.14 vs its 0.39 holdout** (~2.8× overstated); decode-trained banks reach 0.35 on real decode states. Same recipe and rationale as dflash `--self-generate`.
- **Per-pair recall persistence + serve-time gating** — `pair_recalls` in the sidecar, `apply_recall_gate`, `OLMLX_FLASH_MOE_PREFETCH_MIN_RECALL` (default 0.0 = off; legacy sidecars without recalls are never gated).
- **Default off** — even at 2.6× better real recall, throughput was unchanged (14.1 tok/s): predicted-and-routed experts are mostly already LRU-cached, so speculative reads cost more SSD bandwidth than their marginal hits save. Re-enable with `OLMLX_FLASH_MOE_PREFETCH=true`.

## Benchmark ledger (7-prompt throughput set, budget 128)

| Config | tok/s |
|---|---|
| LRU, no prefetch | **20.4, 20.4** (two runs) |
| Prefetch: original → inline-NumPy → non-blocking | 13.0 → 15.5 → 15.7 |
| Prefetch: retrained C4 bank (real decode recall 0.11@8) | 16.1 / 13.2 |
| Prefetch: recall-gated 0.40 / 0.45 | 13.6 / 15.4 |
| Prefetch: decode-trained bank (real recall 0.28@8) | 14.1 / 12.3 |

## Test plan

- 17 new/updated tests (NumPy/MLX prediction equivalence, inline-prediction contract, pending self-clear, recall roundtrip + gate semantics + legacy-bank safety, decode-tracer prefill-skip/cap/restore, sidecar provenance, config defaults).
- Full suite: only 2 failures, both reproduce on clean `main` (`test_rerank_model` SDPA, `test_routers_speech`) — pre-existing, unrelated.
- ruff lint + format clean.

### DIFF
--- a/olmlx/engine/flash/moe_lookahead_train.py
+++ b/olmlx/engine/flash/moe_lookahead_train.py
@@ -158,7 +158,13 @@ def record_moe_router_traces_decode(
 
     ``prompt_source`` yields chat-templated prompt token lists;
     ``_generate_fn(model, tokenizer, prompt_ids, max_new_tokens)`` is a test
-    hook replacing the default ``mlx_lm.stream_generate`` drive.
+    hook replacing the default ``mlx_lm.stream_generate`` drive. The drive
+    MUST decode with batch size 1 — the recorder identifies decode steps as
+    single-position forwards, so a batched drive records nothing. The first
+    single-position forward of each prompt is dropped: mlx_lm always feeds
+    the last prompt token as its own forward (teacher-forced, not a decode
+    state), so per prompt exactly one boundary position would otherwise
+    contaminate the trace.
     """
     import mlx.nn as nn
 
@@ -184,13 +190,13 @@ def record_moe_router_traces_decode(
                 self._inds_sink.append(np.array(flat_inds))
             return self._inner(x)
 
-    if _generate_fn is None:
+    def _default_generate_fn(model, tokenizer, prompt_ids, max_new):
+        from mlx_lm import stream_generate
 
-        def _generate_fn(model, tokenizer, prompt_ids, max_new):
-            from mlx_lm import stream_generate
+        for _ in stream_generate(model, tokenizer, prompt_ids, max_tokens=max_new):
+            pass
 
-            for _ in stream_generate(model, tokenizer, prompt_ids, max_tokens=max_new):
-                pass
+    generate_fn = _generate_fn if _generate_fn is not None else _default_generate_fn
 
     if prompt_source is None:
         from olmlx.engine.dflash.selfgen import iter_dataset_prompts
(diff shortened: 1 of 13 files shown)


olmlx/engine/flash/moe_lookahead_train.py
```diff
@@ -132,6 +133,141 @@ def __call__(self, x: mx.array) -> mx.array:
...
+    single-position forward of each prompt is dropped: mlx_lm always feeds
+    the last prompt token as its own forward (teacher-forced, not a decode
+    state), so per prompt exactly one boundary position would otherwise
+    contaminate the trace.
+    """
+    import mlx.nn as nn
+
+    from olmlx.engine.flash.flash_moe_model import _find_moe_module
+
+    class _DecodeRecorder(nn.Module):
```
dpalmqvist, line 173: **[cleanup/reuse — follow-up]** `_DecodeRecorder` + its install/restore/collect scaffold duplicates ~45 lines of `record_moe_router_traces`'s `_Recorder` machinery; the only substantive difference is the record predicate (single-position vs budget-capped all-positions). The restore-in-finally and `object.__setattr__` ownership-bypass logic are subtle and correctness-critical — a fix applied to one recorder will silently miss the other. Worth extracting one shared recorder parameterized by a position filter in a follow-up (kept out of this PR to limit review surface).

olmlx/engine/flash/moe_predictor.py
```diff
@@ -116,6 +153,58 @@ def predict_next(
...
+            return
+        self._np_heads = [
+            (
+                np.array(head.down.weight.astype(mx.float32)),
+                np.array(head.up.weight.astype(mx.float32)),
+            )
+            for head in self.heads
+        ]
+
+    def predict_next_np(
```
dpalmqvist, line 174: **[cleanup/reuse — follow-up]** `predict_next_np` is a hand-mirrored twin of `predict_next` (and of `SparsityPredictor.__call__`'s math): the top-m selection tail is byte-for-byte duplicated, and the forward math re-encodes the head's layout by convention with no compile-time link. If the head ever gains a bias/activation change or the margin rule is tuned, tests exercising the mx path will pass while the NumPy serve path silently diverges. Follow-up: extract a shared `_top_m(scores, margin)` and keep a single documented source of truth for the head forward.

olmlx/engine/flash/moe_prefetch.py
```diff
@@ -163,22 +161,48 @@ def _do_prefetch() -> None:
...
         try:
             self._io_executor.submit(_do_prefetch)
         except RuntimeError:  # executor shut down
-            state.done.set()
             with self._lock:
+                self._pending.pop(next_layer, None)
                 self.stats.submitted -= 1
+            state.done.set()
+
+    def wait(self, layer_idx: int) -> None:
```
dpalmqvist, line 186: **[cleanup/simplification — follow-up]** With the forward path no longer calling `wait()`, the `_LayerPrefetchState`/`threading.Event` machinery (and the set-done-before-clear ordering it imposes on `_do_prefetch`'s finally) is exercised only by tests — `_pending` could be a plain `set[int]` for dedup. Kept in this PR because tests use `wait()` as a deterministic I/O barrier; a follow-up could replace those with polling on the store cache and delete the Event entirely.

PR #658, opened 2026-07-12
fix(turboquant): materialize flat-path packed KV buffers on the worker thread

dpalmqvist:
## The crash

Flash-MoE + TurboQuant models (GLM-4.5-Air, Qwen3-235B) still crashed with `RuntimeError: There is no Stream(gpu, N) in current thread` at flash-MoE's `mx.eval(inds)` (`flash_moe.py:77`) on cross-request cache reuse — a **fifth** distinct member of the flat-path cross-thread-stream family, surviving all of #653/#654/#655/#657.

Reproduced from `/tmp/olmlx.log`: request 1 (fresh 529-token cache) completes and stores; request 2 (13k tokens, prefix hit 4/533 → 529-token trim) crashes on a *different* worker thread during reuse.

## Root cause

`TurboQuantKVCache.update_and_fetch` returns views over the `_key_dequant`/`_value_dequant` **side** buffer, so the packed-buffer writes (`_key_indices[..., prev:offset, :] = k_idx`, plus norms/values) never enter the returned token's graph. mlx-lm's per-token `mx.eval` therefore never materializes them — they accumulate as a lazy `slice_update` chain bound to the generating worker thread's Metal stream.

- **Checkpoint path** materializes them via `snapshot_cache_for_persistence` → `_pin_state_to_offset`.
- **Flat path** (all trimmable TurboQuant models, `uses_checkpoint_persistence=False`) stored the *live* object with no such eval, so the next request's worker thread crashed rebuilding dequant / deferred-trimming from a dead-thread-bound graph.

Only TurboQuant is affected — Spectral/Shard dequantize *from* the packed buffers on every fetch, so the token eval materializes them.

Confirmed with a cross-thread reproduction: build+shed on thread A without eval → `no Stream(gpu,0)` on thread B reuse; with `mx.eval(packed)` → works.

## Fix

- `TurboQuantKVCache.ensure_state_materialized()` — `mx.eval`s the four packed buffers (no slice, so the live cache keeps its step-aligned headroom).
- `materialize_lazy_cache_state()` duck-typed dispatcher (streaming.py), called **on the generating worker** at end-of-generation: the streaming path wraps the mlx generator in a `finally` finalizer (runs on exhaustion/`gen.close()`, before `_SENTINEL`/store); the non-streaming path calls it in `_generate_sync` before its `mx.synchronize`.
- Per-token eval was rejected — it would reintroduce the O(n²) materialization the side buffer exists to avoid.

## Tests (TDD — written first, watched fail, then green)

- `test_flat_path_reuse_survives_worker_thread_handoff` (Metal-gated) — build-on-T1 / reuse-on-T2 reproduction of the exact crash.
- Dispatcher + streaming-wiring tests in `test_streaming.py`.
- All 506 tests across turboquant/streaming/prompt_cache/inference/probe suites pass; ruff clean.


dpalmqvist:
Evaluated the 6 automated-review (gpt-5.4-mini) findings against the code; all refuted, no changes needed:

1. **"materialize skipped if `stream_generate` raises before first yield"** — `mlx_lm.stream_generate` is a generator function (`inspect.isgeneratorfunction` → True), so calling it never runs the body or raises at call time; any error surfaces inside `yield from gen`, where `_finalizing`'s `finally` runs. And on any generation error the cache is never stored (the `_store_prompt_cache_after_generation` call is gated on `generation_complete`), so it can't cross threads regardless.
2. **"nested-container cache defeats `_cache_needs_materialize`"** — the mlx-lm prompt cache is always a flat `list` of per-layer cache objects; there is no nesting.
3. **"store may persist a different instance than the one materialized"** — the flat path stores the same object; `_set_in_memory` only calls `_shed_transient_buffers` (in-place null of the dequant side buffers), no copy.
4. **"redundant eval on every large trim"** — `mx.eval` on already-materialized leaves is a no-op; the only real work happens when `trim`'s shrink branch actually created lazy slices (exactly when the eval is required), and it runs at most once per over-`max_cache_tokens` turn.
5. **"`ensure_state_materialized` derefs 3 possibly-None buffers"** — the four packed buffers are always allocated/sliced/nulled together (`update_and_fetch` asserts all four non-None; the `.state` property uses the identical single `_key_indices is None` guard), so no reachable inconsistent state exists.
6. **"`_pin_state_to_offset` refactor diverges packed/dequant views"** — the refactor is byte-identical in behavior; it only extracts the existing `mx.eval(...)` into `_eval_packed_buffers()`. `_pin_state_to_offset` runs on snapshot copies only (documented), and dequant is rebuilt from packed on resume — unchanged.

Issue #644, opened 2026-07-10
Malformed tool.function.parameters (non-dict) crashes /v1/chat/completions and /api/chat with 500


dpalmqvist:
## Summary
A tool/function definition whose `parameters` field is not a JSON object (e.g. a string) crashes the server with an unhandled `AttributeError` — 500 on `/v1/chat/completions` (both streaming and non-streaming) and `/api/chat` — instead of a clean 4xx validation error.

## Reproduction

```bash
curl -s http://localhost:11434/v1/chat/completions -d '{
  "model": "mlx-community/Qwen2.5-0.5B-Instruct-4bit:latest",
  "messages": [{"role": "user", "content": "hi"}],
  "tools": [{"type": "function", "function": {"name": "foo", "parameters": "not-an-object"}}]
}'
```

Also reproduces on `/api/chat` (Ollama):
```bash
curl -s http://localhost:11434/api/chat -d '{
  "model": "mlx-community/Qwen2.5-0.5B-Instruct-4bit:latest",
  "messages": [{"role": "user", "content": "hi"}],
  "tools": [{"type": "function", "function": {"name": "foo", "parameters": "not-an-object"}}],
  "stream": false
}'
```

And in streaming mode on `/v1/chat/completions` (with `"stream": true`), the crash surfaces as an SSE error event after a 200 has already been sent.

## Observed
```json
{"error":{"message":"AttributeError: 'str' object has no attribute 'get'","type":"server_error","code":"internal_error"}}
```
HTTP 500 (Ollama) / HTTP 500 (OpenAI non-stream).

Server log:
```
File ".../olmlx/routers/openai.py", line 448, in openai_chat
    _thinking, visible_text, tool_uses = parse_model_output_post(
  File ".../olmlx/routers/streaming_common.py", line 189, in parse_model_output_post
    fill_missing_required_args(tool_uses, declared_tools)
  File ".../olmlx/engine/tool_parser.py", line 934, in fill_missing_required_args
    required = set(params.get("required", []))
                    ^^^^^^^^^^
AttributeError: 'str' object has no attribute 'get'
```

## Expected
A clean `400`/`422` validation error (e.g. "tools[0].function.parameters must be an object"), matching the behavior already implemented for `/v1/messages` (Anthropic), which correctly rejects a non-dict `input_schema` with a 400:
```json
{"type":"error","error":{"type":"invalid_request_error","message":"input_schema: Input should be a valid dictionary or object to extract fields from"}}
```

[...]

Issue #643, opened 2026-07-08
/api/embed always returns prompt_eval_count: null despite computing the real count


dpalmqvist:
## Summary
`/api/embed` always responds with `"prompt_eval_count": null`, even though `generate_embeddings()` already computes the exact input token count and returns it — the route just discards it.

`olmlx/routers/embed.py:15-29`:
```python
@router.post("/api/embed")
async def embed(req: EmbedRequest, request: Request):
    manager = request.app.state.model_manager
    texts = req.input if isinstance(req.input, list) else [req.input]

    with Timer() as t:
        embeddings, _ = await generate_embeddings(       # <-- total_tokens discarded via `_`
            manager, req.model, texts, keep_alive=req.keep_alive
        )

    return EmbedResponse(
        model=req.model,
        embeddings=embeddings,
        total_duration=t.duration_ns,
        # prompt_eval_count / load_duration never set -> both stay None
    )
```

`generate_embeddings` (`olmlx/engine/inference.py:4689-4699`) explicitly documents and returns `(embeddings, total_tokens)`, where `total_tokens` is "the summed token count across all inputs (for OpenAI `usage.prompt_tokens`)" — it's already used for that purpose in the OpenAI `/v1/embeddings` route, just not here.

## Reproduction
Server: `uv run olmlx serve` on `http://localhost:11434`, model `mlx-community/Qwen2.5-0.5B-Instruct-4bit`.

```
$ curl -s http://localhost:11434/api/embed -d '{
  "model": "mlx-community/Qwen2.5-0.5B-Instruct-4bit:latest",
  "input": "Hello world"
}'
{"model":"mlx-community/Qwen2.5-0.5B-Instruct-4bit:latest","embeddings":[[...]],"total_duration":3779875,"load_duration":null,"prompt_eval_count":null}
```

Compare with the OpenAI-compatible `/v1/embeddings` route on the identical input, which correctly reports the token count via `usage.prompt_tokens`:
```
$ curl -s http://localhost:11434/v1/embeddings -d '{"model":"mlx-community/Qwen2.5-0.5B-Instruct-4bit:latest","input":["Hello world"]}'
{... "usage":{"prompt_tokens":2, "completion_tokens":0, "total_tokens":2}}
```

## Observed vs expected
- Observed: `prompt_eval_count` is always `null` on `/api/embed`.
[...]

PR #664, opened 2026-07-17
Fix two API-surface bugs: embed prompt_eval_count (#643) + non-dict tool params crash (#644)

dpalmqvist:
Fixes two independent, self-contained API-correctness bugs surfaced by the explorer review.

## #643 — `/api/embed` always returns `prompt_eval_count: null`
`generate_embeddings()` already computes and returns the summed input token count, but the `/api/embed` route discarded it via `_`, so `prompt_eval_count` was always `null`. This diverged from real Ollama and from olmlx's own `/v1/embeddings` surface, which reports the identical count as `usage.prompt_tokens`.

**Fix:** capture the token count and pass it through as `prompt_eval_count`.

## #644 — malformed `tool.function.parameters` (non-dict) crashes with 500
`Tool.function` (and its `parameters`) is deliberately typed as free-shape `dict[str, Any]` at the request boundary, so a client can send a tool whose `parameters` is a string. `fill_missing_required_args` then called `.get()` on it unconditionally, raising `AttributeError: 'str' object has no attribute 'get'` deep inside `parse_model_output_post` — a **500** on `/v1/chat/completions` (streaming and non-streaming) and `/api/chat`. Because it ran *after* generation, the crash also wasted the completed inference; in streaming mode it surfaced as an SSE error *after* the 200 had already been sent.

**Fix:** reject the malformed request at the router boundary with a clean **400**, before generation, mirroring the Anthropic surface (whose typed `input_schema` already rejects a non-dict `input_schema`). A shared `validate_declared_tools()` helper (in `streaming_common.py`) is called at the top of the OpenAI and Ollama chat handlers; it raises `ValueError` on a present-but-non-dict `function` or `function.parameters`, which the app's existing `ValueError` handler renders as a 400 with the correct per-surface error envelope. Failing fast also stops wasting inference and closes the streaming case (rejects before the stream opens). Anthropic needs no change — its typed schema already 400s.

## Tests
- `tests/test_routers_embed.py`: assert `prompt_eval_count` is surfaced for both string and list inputs.
- `tests/test_streaming_common.py::TestValidateDeclaredTools`: unit-level — well-formed/absent params pass; non-dict tool/`function`/`parameters` raise `ValueError` with the offending index in the message.
- `tests/test_routers_openai.py::TestMalformedToolSchemaRejected` and `tests/test_routers_chat.py::TestChatMalformedToolSchemaRejected`: end-to-end 400 on both non-streaming and streaming, with no `generate_chat` mock (the request must fail before dispatch — any model call would be a bug), asserting the per-dialect error envelope.

Written test-first (all failed before the fix). Affected suites — `test_routers_openai.py`, `test_routers_chat.py`, `test_streaming_common.py`, `test_fill_missing_args.py`, `test_routers_embed.py` — pass (164+); ruff check + format clean.

### DIFF
--- a/olmlx/routers/chat.py
+++ b/olmlx/routers/chat.py
@@ -11,7 +11,11 @@ from olmlx.engine.grammar import parse_response_format
 from olmlx.engine.inference import generate_chat
 from olmlx.engine.panel import panel_generate_chat
 from olmlx.routers.common import format_error, resolve_think_flag
-from olmlx.routers.streaming_common import collect_stream, parse_model_output_post
+from olmlx.routers.streaming_common import (
+    collect_stream,
+    parse_model_output_post,
+    validate_declared_tools,
+)
 from olmlx.routers.thinking_split import (
     flush_split_thinking,
     split_thinking_streaming,
@@ -163,6 +167,10 @@ async def chat(req: ChatRequest, request: Request):
     options = req.options.model_dump(exclude_none=True) if req.options else {}
     messages = [m.model_dump(exclude_none=True) for m in req.messages]
     tools = [t.model_dump(exclude_none=True) for t in req.tools] if req.tools else None
+    # Reject malformed tool schemas before generation (a non-dict
+    # ``parameters`` would otherwise crash post-parse — see #644). Raises
+    # ValueError → 400 via the app's ValueError handler.
+    validate_declared_tools(tools)
     max_tokens = options.pop("num_predict", settings.default_max_tokens)
     cache_id = request.headers.get("x-cache-id", "")[:256]
     enable_thinking = resolve_think_flag(req.think)
--- a/olmlx/routers/openai.py
+++ b/olmlx/routers/openai.py
@@ -28,6 +28,7 @@ from olmlx.routers.streaming_common import (
     parse_buffered_output,
     parse_model_output_post,
     sse_error_event,
+    validate_declared_tools,
 )
 from olmlx.routers.thinking_split import (
     flush_thinking_buffer,
@@ -320,6 +321,10 @@ async def openai_chat(req: OpenAIChatRequest, request: Request):
         req.max_completion_tokens,
     )
     manager = request.app.state.model_manager
+    # Reject malformed tool schemas before generation (a non-dict
+    # ``parameters`` would otherwise crash post-parse — see #644). Raises
+    # ValueError → 400 via the app's ValueError handler.
+    validate_declared_tools(req.tools)
     messages = [m.model_dump(exclude_none=True) for m in req.messages]
     # A malformed image content part (e.g. image_url with no url) is a client
     # error — surface as 422 rather than an uncaught 500.
--- a/olmlx/routers/streaming_common.py
+++ b/olmlx/routers/streaming_common.py
@@ -165,6 +165,39 @@ def parse_buffered_output(
     )
 
 
+def validate_declared_tools(tools: list[dict[str, Any]] | None) -> None:
+    """Reject malformed tool definitions at the router boundary (issue #644).
+
+    ``Tool.function`` and its ``parameters`` are typed as free-shape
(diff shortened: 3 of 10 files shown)


olmlx/routers/streaming_common.py
```diff
@@ -165,6 +165,39 @@ def parse_buffered_output(
...
+            raise ValueError(f"tools[{i}] must be an object")
+        func = tool.get("function")
+        # ``function`` is optional here (some callers pass bare-shaped tools);
+        # only a *present, non-dict* value is malformed.
+        if func is None:
+            continue
+        if not isinstance(func, dict):
+            raise ValueError(f"tools[{i}].function must be an object")
+        params = func.get("parameters")
+        if params is not None and not isinstance(params, dict):
```
dpalmqvist: **Validation depth gap — the same 500 still fires one level deeper.** `validate_declared_tools` only checks that `function.parameters` is a dict, but the actual crash site `fill_missing_required_args` (unchanged) also assumes `parameters.properties` is a dict (`properties.items()`), `parameters.required` is iterable (`set(...)`), and each property value is a dict (`v.get(...)`).

So a request like `parameters: {"required": ["a"], "properties": "x"}` (or `"required": 3`) passes this validator and still 500s **after** generation — the exact crash class #644 set out to fix, one level deeper.

Either validate those sub-fields here too, or make `fill_missing_required_args` coerce non-dict `properties`/`required` to empty (the belt-and-braces the reverted guard had).

olmlx/routers/chat.py
```diff
@@ -163,6 +167,10 @@ async def chat(req: ChatRequest, request: Request):
     options = req.options.model_dump(exclude_none=True) if req.options else {}
     messages = [m.model_dump(exclude_none=True) for m in req.messages]
     tools = [t.model_dump(exclude_none=True) for t in req.tools] if req.tools else None
+    # Reject malformed tool schemas before generation (a non-dict
```
dpalmqvist, line 170: **Third tool surface left uncovered: `/v1/responses`.** The Responses API runs the same vulnerable path — `_stream_tools_response` (`responses.py:379`) and the non-streaming handler (`responses.py:801`) both call `parse_*` with `fill_missing_args=True`, and `_convert_tools` passes `parameters` straight through (`t.get("parameters", {})`). No `validate_declared_tools` is called anywhere in `responses.py`, so a non-dict `parameters` on `/v1/responses` still crashes with a 500 after generation.

Add `validate_declared_tools(req.tools)` in `create_response` (right after `_convert_tools(req.tools)` at `responses.py:730`) to close the same hole on the third tool surface. (Filing here since `responses.py` is not in this diff.)

olmlx/routers/openai.py
```diff
@@ -320,6 +321,10 @@ async def openai_chat(req: OpenAIChatRequest, request: Request):
         req.max_completion_tokens,
     )
     manager = request.app.state.model_manager
+    # Reject malformed tool schemas before generation (a non-dict
+    # ``parameters`` would otherwise crash post-parse — see #644). Raises
+    # ValueError → 400 via the app's ValueError handler.
+    validate_declared_tools(req.tools)
```
dpalmqvist, line 327: **Minor — status-code inconsistency within the same handler.** This raises `ValueError` → **400**, but ~15 lines below, schema-shape client errors (`parse_response_format`, `_normalize_multimodal_messages`) are surfaced as `HTTPException(422)`. So `/v1/chat/completions` now returns 422 for a malformed `response_format` but 400 for a malformed tool schema.

The 400 matches the issue’s stated goal (mirror the Anthropic surface), so this is likely intentional — flagging only for consistency. If you want uniformity, the other two could move to the `ValueError`→400 path too.

dpalmqvist: Intentional — the 400 matches the issue’s goal of mirroring the Anthropic surface. Leaving `response_format`/multimodal on 422 to avoid churn; not changing in this PR.

Issue #619, opened 2026-07-07
Registry load() stores raw mapping keys while mutators normalize: untagged entries un-removable


dpalmqvist:
## Summary
`Registry.load()` stores mapping keys raw while every mutator normalizes them, so a hand-edited untagged models.json entry becomes silently un-removable: `DELETE /api/delete` removes the weights directory and returns 200, but the registry entry persists — now pointing at a deleted directory.

## Location
- `olmlx/engine/registry.py:1535` (raw store), `:1894-1909` (normalized pop)
- `olmlx/routers/manage.py:60-70` (delete path)

## Details
`load()` does `self._mappings[k] = ...` with the raw key (`normalized` is computed at `:1526` but only used for panels/adapters), whereas `remove()` pops `normalize_name(name)` (`:latest` appended). Untagged keys are a supported input — `resolve()` falls back to the raw name at `:1587`.

For a hand-edited entry keyed `"foo"`, `DELETE /api/delete` deletes the weights dir via `store.delete` and returns 200, but `registry.remove("foo")` pops `"foo:latest"`, finds nothing, and the models.json entry persists — pointing at a deleted directory, so the next request silently re-downloads or 400s. `_raw_unrecognized` has the same raw-store (`:1538`) / normalized-pop (`:1905`) mismatch, and a later `add_mapping("foo", ...)` creates a duplicate `"foo:latest"` entry alongside `"foo"`.

## Suggested fix
Normalize the key at load time (`self._mappings[normalized] = ...`).

## Severity
Medium — inconsistent registry state and un-removable/duplicated entries after delete.

---
_Filed from a whole-repo review by Claude Fable 5. Findings were verified against callers and, for parsers, reproduced against live inputs._

dpalmqvist:
Fixed in `5586e71` (#666) — registry load() normalizes untagged keys so entries are removable.

Issue #610, opened 2026-07-07
Anthropic streaming always reports input_tokens: 0


dpalmqvist:
## Summary
Anthropic streaming responses always report `input_tokens: 0`, diverging from the non-streaming path which reports the real prompt token count. Every streaming client sees zero input tokens for cost/context tracking.

## Location
- `olmlx/routers/anthropic.py:775` — `message_start` hardcodes `"input_tokens": 0`
- `olmlx/routers/anthropic.py:834-845` — `message_delta.usage` carries only `output_tokens`
- meta dicts at `:486-489` and `:634-639` extract only `stats.eval_count`

## Details
The non-streaming path (`anthropic.py:937-942`) reports real `input_tokens` from `stats.prompt_eval_count`, which is equally available on the streaming done chunk (`out.stats` / `stats` are captured — only `eval_count` is read). Streaming clients — including Claude Code, a first-class client given the billing-header handling at `anthropic.py:119-153` — see zero input tokens, partially masked only when prompt caching populates `cache_read_input_tokens`.

## Suggested fix
Put `stats.prompt_eval_count` into the streaming meta dict and emit it (corrected `message_start`, or in `message_delta.usage`). One line per path.

## Severity
High (accounting) — broken cost/usage tracking for all Anthropic streaming clients.

---
_Filed from a whole-repo review by Claude Fable 5. Findings were verified against callers and, for parsers, reproduced against live inputs._

dpalmqvist:
Fixed in `5586e71` (#666) — Anthropic streaming now reports real input_tokens (earlier partial in `ce8140a`).

PR #666, opened 2026-07-17
fix: Anthropic streaming input_tokens (#610) + registry untagged-key normalization (#619)

dpalmqvist:
Fixes two independent, whole-repo-review bugs.

## #610 — Anthropic streaming always reports `input_tokens: 0`

The non-tools streaming path (`_stream_thinking_state_machine`) captured the done-chunk `stats` but only read `eval_count`, so `message_delta.usage` always emitted `input_tokens: 0`. This diverged from both the non-streaming path and the buffered-tools streaming path, which already report the real prompt count. Every streaming client (including Claude Code) saw zero input tokens for cost/context tracking.

**Fix:** capture `stats.prompt_eval_count` in the thinking state machine and put it in the meta dict — the existing `message_delta.usage` wiring (already used by the tools path) does the rest. One-line-per-path, matching the shipped tools-path pattern.

## #619 — Untagged models.json entries are un-removable

`Registry.load()` stored mapping keys **raw**, while every mutator normalizes (`:latest` appended). A hand-edited untagged entry (`{"foo": ...}`) became silently un-removable: `DELETE /api/delete` removes the weights dir and returns 200, but `registry.remove("foo")` pops `"foo:latest"`, finds nothing, and the entry persists — now pointing at a deleted directory.

**Fix (both sides):**
- `load()` stores under the **normalized** key (also for `_raw_unrecognized` and the panel fallback), so in-memory mutators find the entry.
- `_save_mappings()` normalizes **on-disk** keys when re-reading the file (skipping the reserved `adapters` section), so removal/overlay match the normalized key and re-adding overwrites rather than creating a duplicate `foo` + `foo:latest` pair. The issue's suggested one-liner (normalize at load) alone would leave the raw key on disk to reload on restart.

## Tests

- `test_streaming_reports_input_tokens` — non-tools stream reports real `input_tokens` in `message_delta.usage` (fails before fix).
- `test_load_normalizes_untagged_mapping_key`, `test_remove_untagged_mapping_deletes_from_disk`, `test_readd_untagged_mapping_no_duplicate` — the #619 in-memory + on-disk behavior (all fail before fix).
- `test_save_preserves_adapters_section` — guard that disk-key normalization doesn't mangle `adapters` → `adapters:latest`.

All written test-first (TDD). Full `test_registry.py`, `test_adapter_registry.py`, `test_registry_panel.py`, `test_routers_anthropic.py`, and `test_routers_manage.py` suites pass; ruff clean.

### DIFF
--- a/olmlx/engine/registry.py
+++ b/olmlx/engine/registry.py
@@ -1529,13 +1529,16 @@ class ModelRegistry:
                         self._panels[normalized] = PanelConfig.from_entry(normalized, v)
                     except (ValueError, TypeError) as exc:
                         logger.warning("Skipping invalid panel entry %r: %s", k, exc)
-                        self._raw_unrecognized[k] = v
+                        self._raw_unrecognized[normalized] = v
                     continue
                 try:
-                    self._mappings[k] = ModelConfig.from_entry(v)
+                    # Store under the normalized (`:latest`-tagged) key so the
+                    # mutators — which all normalize (remove/add_mapping) —
+                    # can find hand-edited untagged entries (issue #619).
+                    self._mappings[normalized] = ModelConfig.from_entry(v)
                 except (ValueError, TypeError) as exc:
                     logger.warning("Skipping invalid models.json entry %r: %s", k, exc)
-                    self._raw_unrecognized[k] = v
+                    self._raw_unrecognized[normalized] = v
             self._validate_panels()
         if self._aliases_path.exists():
             try:
@@ -1842,7 +1845,15 @@ class ModelRegistry:
                     f"existing file is not a JSON object (got "
                     f"{type(loaded).__name__})."
                 )
-            disk_data = loaded
+            # Normalize on-disk keys so removal/overlay (which key by the
+            # normalized `:latest` form) match hand-edited untagged entries,
+            # instead of leaving a dangling entry or writing a duplicate
+            # (issue #619). "adapters" is a reserved section keyed by its
+            # colon-named children, not a model — leave it untouched.
+            disk_data = {
+                dk if dk == "adapters" else self.normalize_name(dk): dv
+                for dk, dv in loaded.items()
+            }
             disk_read_ok = True
 
         if not disk_read_ok and self._mappings:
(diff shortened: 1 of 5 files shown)


olmlx/engine/registry.py
```diff
@@ -1842,7 +1845,15 @@ def _save_mappings_locked(self):
...
-            disk_data = loaded
+            # Normalize on-disk keys so removal/overlay (which key by the
+            # normalized `:latest` form) match hand-edited untagged entries,
+            # instead of leaving a dangling entry or writing a duplicate
+            # (issue #619). "adapters" is a reserved section keyed by its
+            # colon-named children, not a model — leave it untouched.
+            disk_data = {
+                dk if dk == "adapters" else self.normalize_name(dk): dv
+                for dk, dv in loaded.items()
+            }
```
dpalmqvist: **Consistency: on-disk key normalization only covers one of the two save paths.**

This normalizes disk keys in `_save_mappings_locked`, but the parallel read-modify-write in `_save_adapters` → `_read_models_config_for_update` (line 1783) reads the file **raw** and writes it back unchanged. So an adapter save (`add_adapter_mapping` / adapter removal) round-trips a hand-edited untagged `"foo"` key verbatim, while a model save rewrites it to `"foo:latest"` — the on-disk form depends on which save ran last.

Impact is low and self-healing (removal always routes through `_save_mappings`, which normalizes, so #619's removability guarantee holds; the untagged key just lingers on disk until the next model-mutating save). But the fix would be more robust living in the shared helper: normalize keys inside `_read_models_config_for_update` (skipping `"adapters"`) so both `_save_mappings` and `_save_adapters` agree, rather than special-casing one path. Not a blocker.

dpalmqvist: Addressed in bdd8f4e — hoisted the normalization into a shared `_normalize_disk_keys` helper used by both `_save_mappings_locked` and `_read_models_config_for_update` (adapter save path), so both agree on the `:latest` form. Added a test locking the adapter-save path and updated `test_add_adapter_mapping_persists`.

Issue #613, opened 2026-07-07
Non-streaming completions with stop sequences generate to max_tokens under the inference lock


dpalmqvist:
## Summary
Non-streaming completions with a stop sequence generate all the way to `max_tokens` — the stop is only applied post-hoc by truncating the finished text — wasting GPU time under the global inference lock and storing post-stop garbage in the prompt cache.

## Location
- `olmlx/engine/inference.py:3831` (stop popped before the lock)
- `:3964-3984` (post-hoc `truncate_at_stop`)
- `:4124-4142` (drain loop), `:3922-3929` (cache stored with post-stop tokens)

## Details
`_full_completion` pops `stop` before the lock and only applies it post-hoc via `truncate_at_stop`. But `_generate_sync` already iterates `mlx_lm.stream_generate` token-by-token — it could feed a `StopScanner` (as the streaming and batched paths do) and break at the match.

Today a non-streaming request whose stop fires at token 20 with `max_tokens=4096` burns ~4076 tokens of GPU time holding the global lock, and the prompt cache is stored including all the post-stop tokens. The reported `eval_count` is also approximate (re-tokenizing truncated text, `:3981`).

## Suggested fix
Feed a `StopScanner` in `_generate_sync` and break at the match, mirroring the streaming/batched paths.

## Severity
Medium (performance/correctness) — up to max_tokens of wasted generation per stopped non-streaming request, plus polluted cache.

---
_Filed from a whole-repo review by Claude Fable 5. Findings were verified against callers and, for parsers, reproduced against live inputs._

dpalmqvist:
Fixed in #669 (squash-merged to `main` as 9409e2d).

Issue #604, opened 2026-07-07
Stop-sequence hit stores a live, still-mutating prompt cache (silent corruption on next hit)


dpalmqvist:
## Summary
On a stop-sequence match, the streaming completion path stores the prompt cache **without cancelling the generation worker thread first**, so the still-running worker keeps decoding into the very cache object being stored by reference. This misaligns the cache and produces silently wrong output on the next cache hit.

## Location
- `olmlx/engine/inference.py:3546` — stop-sequence `break` with no `stream.cancel()` (contrast the timeout path at `:3574`, which cancels)
- `olmlx/engine/inference.py:3627` — `_store_prompt_cache_after_generation` runs immediately after

## Details
When a stop sequence matches, `_stream_completion` breaks out of the token loop, sets `generation_complete = True`, and stores the prompt cache. But `CancellableStream`'s worker thread is still executing `mlx_lm.stream_generate` — its queue holds up to ~32 buffered tokens (`utils/streaming.py:122`), and each `queue.put` waits up to 10s, so the worker keeps running forward passes that mutate `prompt_cache`. `PromptCacheStore._set_in_memory` stores the state **by reference** (`prompt_cache/store.py:446`), no copy.

Consequences:
1. The stored `CachedPromptState.tokens` undercounts the cache's real offset (which grows by up to ~33 more tokens after the store). The next request's trim math (`_setup_prompt_cache:2210`) then leaves untracked tokens in the KV cache and generation resumes **positionally misaligned** — coherent-looking but wrong output.
2. `_shed_transient_buffers` / `trim_prompt_cache` run on the loop thread concurrently with the worker's `update_and_fetch` on the same cache objects.
3. Wasted GPU work after the stop.

The batched path handles this correctly: `scheduler.cancel(seq)` on stop (`:3170`), and the cache is only handed back in the worker's `done` event.

## Suggested fix
Cancel + `drain_and_join` the stream before storing (as the timeout path does), or skip the store on stop like the timeout path.

## Severity
High — silent output corruption on the next request that hits the poisoned cache.

---
_Filed from a whole-repo review by Claude Fable 5. Findings were verified against callers and, for parsers, reproduced against live inputs._

dpalmqvist:
Fixed in #669 (squash-merged to `main` as 9409e2d).

PR #669, opened 2026-07-18
fix(inference): stop sequences honored mid-generation (#613 + #604)

dpalmqvist:
Fixes two stop-sequence handling bugs on the **single-request (non-batched)** inference paths. The batched paths already handle both cases correctly.

## #613 — Non-streaming stop sequences generate to `max_tokens`

`_full_completion` popped `stop` before the lock and only applied it post-hoc via `truncate_at_stop`, so a non-streaming request whose stop fires at token 20 with `max_tokens=4096` burned ~4076 tokens of GPU time holding the global inference lock, and the prompt cache was stored including all post-stop tokens.

**Fix:** `_generate_sync`'s `mlx_lm.stream_generate` loop now feeds a `StopScanner` and breaks at the earliest match (mirroring the streaming/batched paths). The generator is closed explicitly on the early break so mlx-lm's `wired_limit.__exit__` runs its generation-stream sync on the worker thread. The full text (including the stop marker) is preserved so the existing post-hoc `truncate_at_stop` still trims it and sets `finish_reason`. Because non-streaming is single-threaded and pull-driven, the stored cache's token metadata now aligns exactly with the KV offset.

Scope: the standard mlx_lm text path. The VLM and speculative sub-paths are unchanged and continue to rely on post-hoc truncation.

## #604 — Streaming stop-sequence hit stores a live, still-mutating prompt cache

On a stop match, `_stream_completion` broke out of the token loop and stored the prompt cache **without cancelling the generation worker thread**. The worker (queue depth ~32) kept running forward passes that mutated the very `prompt_cache` object being stored by reference, so the stored `tokens` metadata undercounted the cache's real offset — the next request's trim math then resumed positionally misaligned (coherent-looking but wrong output). It also ran `trim`/`_shed_transient_buffers` on the loop thread concurrently with the worker's `update_and_fetch`.

**Fix:** the stop branch now calls `stream.cancel()` (mirroring the timeout path) so the worker stops decoding, and skips the store — dropping any stale/shared entry for the `cache_id` instead. The prefix is re-prefilled on the next turn. This matches how the timeout path and the batched scheduler already behave.

## Tests
- `test_stop_sequence_breaks_generation_early` / `test_no_stop_sequence_consumes_full_stream` — assert `_full_completion_inner` halts the generator at the match (and doesn't when no stop is set), collecting only up-to-stop tokens while returning the full text for post-hoc truncation.
- `test_stop_cancels_worker_and_skips_cache_store` — drives `_stream_completion` end-to-end with a stop-bearing token stream and asserts the worker is cancelled, `_store_prompt_cache_after_generation` is **not** awaited, and the stale entry is removed.

Ran `tests/test_inference*.py`, `tests/test_stop_sequences.py`, `tests/test_streaming*.py`, `tests/test_prompt_cache*.py`, and all router suites — green. ruff clean.

### DIFF
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -4282,10 +4282,19 @@ async def _full_completion_inner(
                 # mlx-lm's wired_limit.__exit__ runs its generation-stream sync
                 # now, on this worker thread — mirrors CancellableStream._run.
                 # Guarded: mlx_lm.stream_generate returns a real generator, but
-                # tests may substitute a plain iterator with no close().
+                # tests may substitute a plain iterator with no close(). The
+                # close itself is wrapped like CancellableStream._run — GPU
+                # teardown (wired_limit sync) can raise, and an exception here
+                # would supersede a real in-flight error from the loop body.
                 _close = getattr(mlx_gen, "close", None)
                 if callable(_close):
-                    _close()
+                    try:
+                        _close()
+                    except Exception:
+                        logger.debug(
+                            "stream_generate close() failed during teardown",
+                            exc_info=True,
+                        )
             # Store full text on the result for downstream extraction
             if result is not None:
                 result = (result, "".join(text_parts))
--- a/olmlx/engine/inference.py
+++ b/olmlx/engine/inference.py
@@ -3596,6 +3596,7 @@ async def _stream_completion(
         )
         timed_out = False
         stop_hit = False
+        stopped = False
 
         with (
             _tracing.span("decode") as _decode_span,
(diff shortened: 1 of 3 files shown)


olmlx/engine/inference.py
```diff
@@ -4218,25 +4241,51 @@ def _generate_sync():
...
+                            break
+            finally:
+                # Close the generator explicitly on an early (stop) break so
+                # mlx-lm's wired_limit.__exit__ runs its generation-stream sync
+                # now, on this worker thread — mirrors CancellableStream._run.
+                # Guarded: mlx_lm.stream_generate returns a real generator, but
+                # tests may substitute a plain iterator with no close().
+                _close = getattr(mlx_gen, "close", None)
+                if callable(_close):
+                    _close()
```
dpalmqvist: **Robustness nit (parity with `CancellableStream._run`):** this `mlx_gen.close()` is unguarded, but the reference it mirrors wraps close in `try/except` (`utils/streaming.py:186-189`) precisely because GPU teardown (`wired_limit.__exit__` → `mx.synchronize`) can raise.

Failure scenario: if `mlx_lm.stream_generate` raises mid-decode (e.g. Metal OOM), the `finally` runs `_close()`, which throws `GeneratorExit` into the generator and runs its wired_limit sync. If that sync also raises, the new exception supersedes the original in the `finally`, so the caller sees the teardown error instead of the real OOM — degraded diagnostics on the failure path.

```suggestion
                _close = getattr(mlx_gen, "close", None)
                if callable(_close):
                    try:
                        _close()
                    except Exception:
                        logger.debug(
                            "stream_generate close() failed during teardown",
                            exc_info=True,
                        )
```

PR #673, opened 2026-07-18
fix(api): /api/generate context continuation (#656) + streaming first-byte latency (#659)

### LINKED ISSUE #656: /api/generate 'context' field (legacy multi-turn continuation) is accepted but completely non-functional
## Summary

`/api/generate`'s legacy `context` field — the mechanism the real Ollama API uses for stateless multi-turn continuation (return the `context` token-id array from one response, pass it back as `context` in the next request) — is declared in both `GenerateRequest` and `GenerateResponse` schemas but is completely non-functional: the request field is silently accepted and ignored, and the response never populates it.

This breaks any Ollama-compatible client/SDK that uses `/api/generate` + `context` for multi-turn conversations instead of `/api/chat` + `messages` (this is the original/classic Ollama pattern, still used by e.g. older LangChain `Ollama` LLM wrappers and various Ollama SDKs).

## Reproduction

Server: `uv run olmlx`, model `mlx-community/Qwen2.5-0.5B-Instruct-4bit`.

```bash
curl -s http://localhost:11434/api/generate -d '{
  "model": "mlx-community/Qwen2.5-0.5B-Instruct-4bit",
  "prompt": "Continue: 1, 2, 3,",
  "context": [1,2,3,4,5],
  "stream": false
}'
```

Observed:
```json
{
    "model": "mlx-community/Qwen2.5-0.5B-Instruct-4bit",
    "created_at": "...",
    "response": "1. The first step in creating a new project is to define the purpose and goals of the project.\n\n2. Next, you will need to decide on the technology stack...",
    "done": true,
    "done_reason": "stop",
    ...
}
```

- Passing an arbitrary/garbage `context` array (`[1,2,3,4,5]`, not even valid tokens for this model) produces no validation error and has zero observable effect on the prompt.
- The response contains no `context` key at all, even though `GenerateResponse.context` (`olmlx/schemas/generate.py:42`) declares it.

## Expected

Either:
1. Implement the real Ollama semantics: the response should return the full token-id sequence (prompt + generated) as `context`, and a request-supplied `context` should be decoded back into raw prior tokens and used as the continuation prefix (bypassing the chat template, since `context` is the classic non-chat completion mode), or
2. If this is intentionally unsupported, drop the field from the schemas (or reject requests that set it) rather than silently accepting and discarding it, since keeping it in the schema advertises Ollama-API compatibility that doesn't exist.

## Code pointers

- `olmlx/schemas/generate.py:19` — `GenerateRequest.context`, never read anywhere.
- `olmlx/schemas/generate.py:42` — `GenerateResponse.context`, never set anywhere.
- `olmlx/routers/generate.py` — the `/api/generate` handler never references `req.context`, and `generate_completion()` (`olmlx/engine/inference.py:1473`) has no `context`/token-ids parameter at all.

## Server log

No errors — the request is accepted with 200 OK, this is a silent behavioral gap, not a crash.

### LINKED ISSUE #659: All chat streaming surfaces buffer ~200 chars (or the whole response) before the first byte — thinking-tag detector, not passthrough
## Summary

Every streaming chat surface (`/api/chat`, `/api/generate`, `/v1/chat/completions`, `/v1/messages`, `/v1/responses`) withholds the **first ~200 characters** of a response before emitting anything to the client — and for responses shorter than that, the entire response is buffered and sent as one chunk only once generation has fully finished. This defeats the purpose of streaming (progressive rendering / low time-to-first-token) for a large share of real chat responses, even when the model never emits `<think>` tags.

Only the non-chat `/v1/completions` (raw text completion) is unaffected.

## Root cause

`olmlx/routers/thinking_split.py`'s `split_thinking_parts` state machine, used by every chat-surface router to split `<think>...</think>` content out of the stream, has a `detect` phase that accumulates all incoming text into `state["buffer"]` and only starts emitting when either:
1. an open/close thinking tag is found, or
2. the buffer exceeds `_STREAM_DETECT_LIMIT` (200 chars, `thinking_split.py:37`) — at which point the *entire accumulated buffer* is flushed as a single chunk (`thinking_split.py:152-163`).

Unlike the `passthrough` phase (which only ever holds back the longest possible partial-tag suffix, a handful of characters — see `_longest_open_tag_suffix`), the `detect` phase holds back **everything** until the 200-char watermark, even though most of that text can never be part of a tag (e.g. as soon as a chunk has no trailing partial-tag prefix, it's provably safe to flush). So instead of ~20 bytes of worst-case latency, every streaming response pays up to a 200-char latency tax before the client sees a single byte — and responses under 200 chars never stream incrementally at all.

The comment on `_STREAM_DETECT_LIMIT` states the intent explicitly: "Sized to let non-thinking models start streaming quickly" — but 200 chars is ~40-50 tokens, which is far from "quick" for anything beyond a tiny toy model.

## Reproduction

Server: `uv run olmlx`. Any small model works; shown with `mlx-community/Qwen2.5-0.5B-Instruct-4bit` and `mlx-community/Qwen3-4B-4bit` (neither configured for thinking).

```bash
# /api/generate — same behavior on /api/chat, /v1/chat/completions, /v1/messages, /v1/responses
curl -s http://localhost:11434/api/generate -d '{
  "model": "mlx-community/Qwen2.5-0.5B-Instruct-4bit",
  "prompt": "Write a short paragraph about the ocean.",
  "stream": true
}'
```

Client-side timing harness (httpx, wall-clock per chunk):
```
  0 t= 0.207s words= 40 done=False text='The ocean is a vast and mysterious place that is home to an '
  1 t= 0.210s words=  1 done=False text=' of'
  2 t= 0.212s words=  1 done=False text=' endless'
  ...
```

Same test against `mlx-community/Qwen3-4B-4bit` (a realistically-sized model, ~11ms/token decode):
```
  0 t= 1.663s chars= 206 done=False text="The ocean covers more than 70% of Earth's surface, stretchin"
  1 t= 1.673s chars=   1 done=False text=','
  2 t= 1.684s chars=   4 done=False text=' and'
  ...
```
Chunk 0 alone contains ~50 tokens (~550ms of decode time at this model's speed) that a real client would have received incrementally starting right after prefill, but instead receives all at once.

Confirmed on all affected surfaces — first SSE/NDJSON event contains the same oversized blob:
- `/v1/chat/completions` (OpenAI SSE `delta.content`)
- `/v1/messages` (Anthropic `content_block_delta` — for a response under 200 chars, e.g. "Bonjour! C'est un plaisir de vous rencontrer...", the *entire* message arrives in a single `content_block_delta`, i.e. no incremental streaming at all)
- `/v1/responses` (`response.output_text.delta`)
- `/api/chat` (`message.content`)

`/v1/completions` (raw completion, no chat template / no thinking-split) streams token-by-token from the first chunk, confirming the thinking-split buffer is the cause.

## Expected

The `detect` phase should only hold back the minimum bytes needed to recognize a tag that might straddle a chunk boundary (as `passthrough` already does via `_longest_open_tag_suffix`), and flush everything else immediately — not accumulate up to a fixed 200-char watermark before releasing anything. First-byte latency should track actual per-token decode time, not an artificial buffering window.

## Severity

Medium-high — this is a systemic streaming-UX regression affecting the majority of chat traffic across every API surface (Ollama, OpenAI, Anthropic, Responses); short responses (a very common case) don't stream incrementally at all.

dpalmqvist:
Closes #656
Closes #659

Two independent API/streaming gaps surfaced by the nightly explorer.

## #659 — streaming buffers ~200 chars before the first byte

The thinking-split `detect` phase accumulated up to `_STREAM_DETECT_LIMIT` (200 chars) before emitting anything, taxing first-byte latency on **every** chat surface (`/api/chat`, `/api/generate`, `/v1/chat/completions`, `/v1/messages`, `/v1/responses`) and never streaming sub-200-char responses incrementally — even for models that emit no `<think>` tags.

The detect buffer is only needed for the orphan-`</think>` heuristic, which is gated on `thinking_expected`. When thinking is not expected, `detect` is equivalent to `passthrough`, so it now flushes everything but a trailing partial open-tag immediately.

**Verified live:** first chunk is now a single token (`'The'`, 3 chars) instead of the ~200-char blob the issue reported.

## #656 — `/api/generate` `context` field accepted but non-functional

The Ollama legacy `context` field (return the token-id array from one response, pass it back to continue statelessly) was declared in the schemas but silently accepted and never populated. This PR implements the real semantics on the text path:

- Tokenize the prompt, prepend any prior `context` tokens (dropping a doubled leading BOS), feed generation the exact token ids.
- Return `context` = prefilled input tokens + generated token ids, on **both** the non-streaming result and the streaming terminal chunk.
- Text models only (VLM token ids can't represent image patches); `/v1/completions` is untouched (only `/api/generate` opts in). Context requests bypass the batched path so per-token ids are captured.

**Verified live (real Qwen2.5-0.5B):**
```
Call 1: "Name three primary colors" -> "Red, Blue, and Yellow."  (context: 45 tokens)
Call 2: pass context back + " Now name three more." -> "Orange, Green, and Purple."  (context: 87)
```
Streaming carries `context` in the final NDJSON frame too.

Note: a stop-sequence hit in **non-streaming** leaves a few post-stop tokens in the returned `context` (the streaming path cancels at the stop, so it doesn't) — a minor stream/non-stream inconsistency judged acceptable for a best-effort compat feature rather than re-tokenizing truncated text.

## Tests

- `tests/test_routers_openai.py` — new detect-phase immediate-flush cases (#659).
- `tests/test_generate_context.py` — pure token-prefix + stream-augment unit tests, `generate_completion` assembly (mocked), and a real-model stream + non-stream round-trip (`real_model`-gated).
- `tests/test_routers_generate.py` — router forwarding + response/final-frame surfacing (#656).
- Full sweep of all touched suites: 646 passed. Ruff clean.

---

- [x] `ruff check` + `ruff format` clean
- [x] Tests pass (`uv run pytest`)
- [x] Each fixed issue has its own `Closes #N` line above


dpalmqvist:
Thanks for the review — addressed in 5f92b22:

**Fixed (real bugs):**
- **Speculative non-streaming context (finding @ inference.py:4031)** — confirmed: the speculative branch in `_full_completion_inner` never appended to `generated_tokens_out`, so a speculative model would return a prompt-only context. Now collects `response.token`, mirroring the mlx-lm branch. (The *streaming* speculative path was already fine — it shares the common token loop.)
- **Batched streaming path (finding @ inference.py:3749)** — confirmed: `_stream_completion` delegated to `_stream_completion_batched` without the `collect_generated_tokens` guard that `_full_completion` already has, so a batch-eligible context request would attach an empty continuation. Added the guard.
- Both are covered by new regression tests, and I also fixed the sole Pyright CI error (reassigning the `str`-typed `prompt` param to a token list — now a union-typed `gen_prompt` local).

**Design findings 3 & 4 (unconditional `return_context=True`):** kept intentional. Ollama returns `context` on *every* `/api/generate` response, including the first turn — a client can't opt in via `req.context` because the first request has no prior context to send, so gating on `req.context is not None` would break the primary continuation flow. On cost: when batching is off (the default — `OLMLX_BATCHING` is opt-in) there's no double tokenization (the prompt is tokenized once and passed as ids), so overhead is negligible. The only real effect is that a context request bypasses the batched path — which is *required*, since the batch scheduler doesn't surface per-token ids. Supporting context inside the batched path (accumulating `event["token"]`) is feasible but touches the concurrency-sensitive scheduler; I've left it as a follow-up rather than widen this PR's blast radius. Happy to do it here if you'd prefer.

**Finding 5 (final frame forwards both `context` and `generated_tokens`):** not reproducible — the router builds the terminal frame field-by-field (`model/created_at/response/done/done_reason/context/stats`) and never copies `generated_tokens`; additionally `_augment_stream_with_context` pops `generated_tokens` off the done chunk before it reaches the router. So the wire response only ever carries `context`.

PR #677, opened 2026-07-18
fix(speculative): GDN-capture hazards, doubled BOS, redundant eval, hook order (#633)

### LINKED ISSUE #633: Speculative decoders: GDN-capture memory hazards, doubled BOS, redundant pending-token eval, hook-patch ordering
## Summary
Grouped smaller fixes across the speculative decoders (beyond the SWA-trim and off-by-one issues filed separately).

## Findings

**1. DFlash/EAGLE leave GDN capture active across the whole prompt prefill (`dflash/decoder.py:250-251`, `eagle/decoder.py:274-275`).**
Both activate the capture buffer *before* the prompt forward on hybrid targets, so `_capturing_gdn_call` appends q/k/v/a/b + conv tensors for every prompt position in every GDN layer, held live until the first `step()` clears it. These captures are never consumed — rollback only replays the verify window, and the buffer is cleared before each verify. On long agentic prompts this transiently adds activation-scale memory on exactly the hybrid targets DFlash names as its primary use case. MTP does this correctly (`use_buffer(None)` during the chunked prefix, `mtp/decoder.py:313-323`). EAGLE's justification comment ("subsequent rollbacks may need to replay against [the prompt state]") is incorrect.

**2. Classic/PLD leave the GDN capture routed to a buffer after a mid-step exception (`speculative.py:867-870`, `:1605-1608`).**
On an exception during the target forward, `SpecDecoderBase.step()` resets, but the decoder-lifetime capture is untouched and `use_buffer(None)` never runs (`_reset_state` doesn't touch it; the next `prefill` does). Until then the class-level `GatedDeltaNet.__call__` patch keeps appending every GDN call — including *other* hybrid models' non-speculative inference sharing the GDN class — into the stale buffer: unbounded memory growth. Fix: `use_buffer(None)` in `_reset_state` or try/finally around the buffered forwards.

**3. Speculative paths bypass the BOS heuristic — doubled BOS on BOS-prefixed templates (`inference.py:3458-3462`, `:4068-4071`).**
The speculative branches tokenize with plain `lm.text_tokenizer.encode(prompt)` (`add_special_tokens=True`), while `tokenize_for_cache` (`:953-962`) exists specifically to replicate mlx-lm's BOS heuristic. Templates that render a literal BOS (Llama 3, Gemma, Mistral) get a duplicate BOS id on every speculative request — a silent prompt divergence from the non-speculative path. Both call sites should use `tokenize_for_cache`.

**4. Classic step pays a redundant Metal sync computing the pending token (`speculative.py:938-940`).**
`_pending_token = argmax(verification_logits[num_accepted-1])` after an `mx.eval` — but this is definitionally `accepted[-1]`, already computed by `verify_draft_greedy`. PLD does exactly this shortcut and documents why (`:1641-1646`). One avoidable eval + `.item()` round-trip per decode step; `_last_target_logit` becomes dead state.

**5. EAGLE/MTP patch layer hooks before `make_prompt_cache` (`eagle/decoder.py:202-206`, `mtp/decoder.py:208-212`).**
DFlash deliberately builds the cache *before* patching, with a comment (`dflash/decoder.py:225-232`) explaining that `make_prompt_cache` probes layer objects and an upstream move from `hasattr` to `isinstance` would silently pick the wrong cache type for a `_LayerHook`-wrapped layer. EAGLE/MTP use the fragile opposite order — works today via `__getattr__` proxying, but one mlx-lm bump from breaking.

## Severity
Low-to-medium individually; (1) and (2) are real memory hazards, (3) is a silent prompt divergence.

---

dpalmqvist:
## Summary
Five grouped fixes across the speculative decoders (SWA-trim and off-by-one are tracked separately).

1. **GDN capture across prompt prefill (memory)** — DFlash/EAGLE activated the GDN capture buffer *before* the prompt forward, so the patched `GatedDeltaNet.__call__` appended activation-scale q/k/v/conv tensors for every prompt position, held live until the first `step()` cleared it — on exactly the hybrid GDN targets these strategies name as their use case. Now suppress the buffer across the prefix pass and re-enable before the tail pass (mirroring MTP). `step()` only ever needs the verify window.

2. **Stale GDN buffer after a mid-step exception (memory)** — Classic/PLD now detach the decoder-lifetime GDN capture in `_reset_state` (which the base `step()` runs on exception). Previously the class-level patch stayed routed to a stale buffer, so every later GDN call — including *other* hybrid models' non-speculative inference sharing the patched class — kept appending, leaking until the next prefill. Regression test added.

3. **Doubled BOS on speculative paths** — both speculative tokenize sites use `tokenize_for_cache` instead of plain `encode`, replicating mlx-lm's BOS heuristic. Plain encode doubled the BOS on BOS-prefixed templates (Llama 3 / Gemma / Mistral), diverging from the non-speculative path.

4. **Redundant Metal sync for the pending token** — classic step derives the next token from `accepted[-1]` (already argmax'd by `verify_draft_greedy`) instead of a second `mx.eval` + `argmax` + `.item()` per decode step. Exactness-preserving; PLD already does this.

5. **Hook-patch ordering** — EAGLE/MTP build `make_prompt_cache` before patching layer hooks (mirroring DFlash), so cache-type selection can't be fooled by a `_LayerHook`-wrapped layer under a future mlx-lm `isinstance` probe.

## Tests
New regression for (2); full speculative suite (240 tests) passes; ruff clean.

Closes #633

### DIFF
--- a/olmlx/engine/dflash/decoder.py
+++ b/olmlx/engine/dflash/decoder.py
@@ -226,6 +226,11 @@ class DFlashDecoder(SpecDecoderBase):
                 for p, q in zip(captured_prefix, captured_last)
             ]
         else:
+            # Single-token prompt: no long prefix to suppress, so the buffer
+            # must be active for this forward so step()'s rollback has the
+            # captured GDN state (mirrors MTP's single-token branch).
+            if self._capture is not None:
+                self._capture.use_buffer(self._capture_buffer)
             target_out = self._target(prompt, cache=self._target_cache)
             captured = list(self._hidden_capture)
             if any(h is None for h in captured):
--- a/olmlx/engine/dflash/decoder.py
+++ b/olmlx/engine/dflash/decoder.py
@@ -173,7 +173,6 @@ class DFlashDecoder(SpecDecoderBase):
             # target has no ``GatedDeltaNet`` submodule, so a separate
             # pre-check would just duplicate the same error.
             self._install_gdn_capture()
-            self._capture.use_buffer(self._capture_buffer)
 
         # Two-pass prefill avoids materialising the full [batch, N, vocab] logit
         # (the OOM path for large-vocab models on long contexts).
@@ -191,8 +190,20 @@ class DFlashDecoder(SpecDecoderBase):
             "target_layer_ids — check that the layer indices exist on "
             f"this model (got {target_layer_ids})."
         )
+        # GDN capture is suppressed across the (potentially long) prefix pass
+        # and re-enabled before the tail pass, mirroring MTP (#633): the
+        # patched GDN ``__call__`` *appends* per forward, so capturing every
+        # prompt position would pile activation-scale q/k/v/conv tensors into
+        # the buffer and hold them live until the first ``step()`` clears it —
+        # re-bloating exactly the memory the two-pass split bounds, on the
+        # hybrid GDN targets DFlash names as its primary use case. ``step()``
+        # only ever needs the verify window (it clears the buffer and
+        # re-captures before each rollback), so the prompt captures are never
+        # consumed. ``_capture`` is only set on hybrid GDN targets.
         if prompt.shape[1] > 1:
             prefix, last = prompt[:, :-1], prompt[:, -1:]
+            if self._capture is not None:
+                self._capture.use_buffer(None)
             self._target(prefix, cache=self._target_cache)  # output discarded
             captured_prefix = list(self._hidden_capture)
             if any(h is None for h in captured_prefix):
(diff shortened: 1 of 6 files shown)


dpalmqvist:
Re: the aider suggestion to extend the `_reset_state` buffer-detach to EagleDecoder/MTPDecoder/DFlashDecoder — this isn't needed, and would be a no-op. Those three install their GDN capture as the **request-lifetime** base `self._capture` (via `SpecDecoderBase._install_gdn_capture`), and `SpecDecoderBase.reset()` already calls `self._capture.close()` (un-patching `GatedDeltaNet.__call__` and releasing the lock) before `_reset_state()` runs, on every reset including the mid-step-exception path. So there is no stale patch left appending into a buffer.

Classic/PLD are different: they own a **decoder-lifetime** `self._gdn_capture` that survives across requests and is *not* touched by base `reset()` (only by `close()`), so a mid-step exception left it routed to a stale buffer — that's the leak this PR fixes with the explicit `use_buffer(None)` in their `_reset_state`. By the time `_reset_state` runs for the request-lifetime decoders, `self._capture` is already `None`, so adding the same guard there would do nothing.

PR #678, opened 2026-07-18
fix(routers): provider protocol-consistency fixes across dialects (#627)

### LINKED ISSUE #627: Router protocol-consistency gaps: error envelopes, /v1/completions, panel advertising, Anthropic wire-format divergences
## Summary
A cluster of protocol-consistency gaps across the three API dialects, where the same request produces different behavior depending on the surface, or diverges from the upstream API's wire format. Grouped as router-consistency polish.

## Findings

**1. Router-raised `HTTPException`s bypass the provider-shaped error envelopes (`openai.py:347,361`, `anthropic.py:662,708`, `responses.py:187-190,752`, `chat.py:176`, `generate.py:33`, `audio.py:153-169`).**
`app.py:325-358` deliberately converts Pydantic validation errors into 400s with provider envelopes (`{"error": {...}}` for OpenAI, `{"type":"error","error":{...}}` for `/v1/messages`) because "Provider SDKs expect provider-shaped error envelopes." But these routers raise bare `HTTPException(422, detail=...)` for the same class of client error (malformed image block, bad response_format/schema), which FastAPI serializes as `{"detail": "..."}` — SDKs fail to extract the message, and the same invalid request gets 400 or 422 depending on which layer catches it. A shared `HTTPException` handler routing through `_make_error_response` (or raising `ValueError` and letting `app.py:360` map it, as `audio.py:80-84` already does) closes both gaps.

**2. `/v1/completions` uses only `prompt[0]` and never strips thinking (`openai.py:526,541-552,570`).**
Silently drops all but the first prompt of a list (the validator at `schemas/openai.py:157-162` only prevents the empty-list IndexError). And unlike `/api/generate`, neither streaming nor non-streaming completion strips `<think>` blocks, so thinking models leak reasoning into `text` on this surface only.

**3. Panels advertised on surfaces that can't run them (`openai.py:602`, `models.py:55-62`).**
Synthetic panels are listed in `/v1/models` and `/api/tags`, but only `chat.py:178-182` and `openai.py:396` dispatch to `panel_generate_chat`; `/v1/responses` and `/v1/messages` call `generate_chat` directly, so selecting an advertised panel there yields a 400 "model not found."

**4. Anthropic stop-sequence hits report `end_turn` instead of `stop_sequence` (`anthropic.py:480-485,930-936`).**
A client `stop_sequences` match ends with engine `done_reason="stop"` → `stop_reason: "end_turn"`; the real Anthropic API returns `"stop_sequence"` plus the matched string. Clients branching on it misclassify the stop.

**5. Anthropic incremental streaming appends a spurious empty text block after thinking-only responses (`anthropic.py:620-632`).**
`if not text_block_emitted` unconditionally emits an empty text block; the non-streaming and buffered-tools paths both suppress it when a thinking block exists. Same thinking-only response is `[thinking]` non-streaming but `[thinking, text("")]` streaming.

**6. `_convert_messages` reorders mixed user content (`anthropic.py:226-260`).**
`tool_result` blocks are appended immediately inside the block loop, while text/image parts are appended as one message after the loop — so a user turn `[text, tool_result]` becomes `tool` message then `user` text, inverting the client's order and changing KV-cache prefixes for interrupted-tool-use turns.

**7. OpenAI deltas repeat `role` in every chunk (`openai.py:431-435`).**
`"role": "assistant"` is put in every content delta, not just the first as OpenAI does. (Note: `stream_options.include_usage` being ignored is already tracked in #595.)

## Suggested fix
Centralize error-envelope shaping in a shared `HTTPException` handler; fix `/v1/completions` to honor multi-prompt + strip thinking; either dispatch panels on all surfaces or don't advertise them there; map stop-sequence stops to `stop_sequence`; suppress the empty streaming text block; preserve user-content order; emit `role` only on the first delta.

## Severity
Low-to-medium — wire-format divergences; individually minor, collectively a compatibility drag.

---

dpalmqvist:
## Summary
Seven grouped wire-format / cross-surface consistency fixes.

1. **Error envelopes** — a shared `HTTPException` handler (`app.py`) reshapes router-raised `HTTPException`s into the surface's provider error envelope instead of FastAPI's default `{"detail": ...}` (which provider SDKs can't parse). Status code + carried headers preserved; a status→error-type map keeps it consistent with the typed handlers.
2. **`/v1/completions`** — honors every prompt in a list (one choice each) instead of silently using `prompt[0]`; streaming rejects a multi-prompt request explicitly. Both paths now strip `<think>` reasoning like `/api/generate`.
3. **Panels** — dispatched on `/v1/messages` and `/v1/responses` via `panel_generate_chat`, so a panel advertised in `/v1/models` isn't a 400 "model not found" there.
4. **Anthropic stop_sequence** — a client `stop_sequences` match reports `stop_reason: "stop_sequence"` instead of `"end_turn"` (the engine sets `done_reason="stop"` only on a stop-sequence hit; natural EOS leaves it unset). Shared `_anthropic_stop_reason` used by all three mapping sites.
5. **Anthropic streaming** — no longer appends a spurious empty text block after a thinking-only response (tracks `thinking_block_emitted`).
6. **`_convert_messages` order** — a `[text, tool_result]` user turn stays user-then-tool instead of being reordered to tool-then-user (which changed KV-cache prefixes for interrupted tool use).
7. **OpenAI deltas** — `role` emitted only on the first delta, not every content chunk.

## Tests
Added for findings 1, 2, 3, 4, 6, 7 (RED-verified where applicable). Full router suite (343 tests) passes; ruff + pyright (0 errors) clean.

## Scope note
Surfacing the matched `stop_sequence` **string** would require engine plumbing across the generation paths (`StopScanner` doesn't return which sequence matched); this router-scoped PR fixes the client-visible `stop_reason` **classification** and leaves the string field as-is.

Closes #627

### DIFF
--- a/olmlx/app.py
+++ b/olmlx/app.py
@@ -11,6 +11,7 @@ from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from fastapi.staticfiles import StaticFiles
+from starlette.exceptions import HTTPException as StarletteHTTPException
 from starlette.middleware.base import BaseHTTPMiddleware
 
 from olmlx.config import settings
@@ -325,6 +326,25 @@ class RootSpanMiddleware(BaseHTTPMiddleware):
             return response
 
 
+def _error_types_for_status(status_code: int) -> tuple[str, str, str]:
+    """Map an HTTP status to ``(anthropic_type, openai_type, openai_code)``.
+
+    Used to give a bare ``HTTPException`` a provider-shaped error envelope
+    consistent with the typed handlers below (#627).
+    """
+    if status_code == 401:
+        return ("authentication_error", "invalid_request_error", "authentication_error")
+    if status_code == 403:
+        return ("permission_error", "invalid_request_error", "permission_denied")
+    if status_code == 404:
+        return ("not_found_error", "invalid_request_error", "not_found")
+    if status_code == 429:
+        return ("rate_limit_error", "rate_limit_error", "rate_limit_exceeded")
+    if 400 <= status_code < 500:
+        return ("invalid_request_error", "invalid_request_error", "invalid_value")
+    return ("api_error", "server_error", "internal_error")
+
+
 def _make_error_response(
     path: str,
     status_code: int,
@@ -395,6 +415,32 @@ def create_app() -> FastAPI:
             "invalid_request_error",
             "invalid_value",
         )
+
+    @app.exception_handler(StarletteHTTPException)
+    async def http_exception_handler(request: Request, exc: StarletteHTTPException):
+        # Router-raised ``HTTPException``s (e.g. 422 for a malformed image
+        # block or bad response_format) would otherwise serialize as
+        # FastAPI's default ``{"detail": ...}``, which provider SDKs can't
+        # parse. Reshape into the surface's provider error envelope,
+        # preserving the status code and any carried headers (#627).
+        detail = exc.detail if isinstance(exc.detail, str) else str(exc.detail)
+        anthropic_type, openai_type, openai_code = _error_types_for_status(
+            exc.status_code
+        )
+        logger.warning(
+            "HTTPException %d on %s: %s", exc.status_code, request.url.path, detail
+        )
+        response = _make_error_response(
+            request.url.path,
+            exc.status_code,
+            detail,
+            anthropic_type,
+            openai_type,
+            openai_code,
+        )
+        for key, value in (exc.headers or {}).items():
+            response.headers[key] = value
+        return response
 
     @app.exception_handler(ValueError)
     async def value_error_handler(request: Request, exc: ValueError):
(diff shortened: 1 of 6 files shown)


dpalmqvist:
Re aider's two points:

1. **HTTPException shape on non-`/v1` endpoints** — the new handler routes non-provider paths through `_make_error_response`, which returns `{"error": msg}` for them. That is actually Ollama's own error shape (`{"error": "..."}`), so for the `/api/*` surface this is *more* correct than FastAPI's default `{"detail": ...}`, not a regression. The full router suite (343 tests) passes, including the Ollama-surface tests.

2. **Completions thinking-strip** — this is the explicit ask in #627 ("unlike /api/generate, neither streaming nor non-streaming strips `<think>`"). It matches `/api/generate`'s existing unconditional strip, and `strip_thinking_streaming` only removes content between literal `<think>`/`</think>` tags, which a non-thinking model doesn't emit — so legitimate output is unaffected in practice, and the behavior is consistent with the other text surface.

Also pushed a ruff-format fix for the appended test block (the failing check).

PR #682, opened 2026-07-19
fix(lifecycle): failed-startup rollback, stop() closes models, evictee leaks (#626)

### LINKED ISSUE #626: Lifecycle leaks: failed startup poisons metrics registry, stop() skips model close, evictee leaks on exceptions
## Summary
Three lifecycle defects in app/model-manager teardown: a failed startup permanently poisons the metrics registry for later `create_app()` calls, `ModelManager.stop()` clears `_loaded` without closing the models, and evictees leak on two exception paths.

## Locations & details

**1. Failed startup leaks the expiry task and poisons the metrics registry (`app.py:136-159`).**
There's no exception protection between `manager.start_expiry_checker()` and `yield`. If `settings.models_dir.mkdir` (`:138`), `REGISTRY.register(stats_collector)` (`:146`), or `agent_service.startup()` (`:156`) raises, the shutdown block after `yield` never runs: the `_check_expiry_loop` task is orphaned and the collector stays registered in the module-level `REGISTRY`. The shutdown comment (`:166-167`) says unregistering exists so a subsequent `create_app()` (e.g. in tests) doesn't raise a duplicate-collector error — so after one failed startup, every later `create_app()` in the same process fails on collector registration. Fix: wrap post-`start_expiry_checker` startup in try/except that calls `manager.stop()` + unregister before re-raising.

**2. `ModelManager.stop()` clears `_loaded` without closing models (`model_manager.py:951-989`).**
`stop()` cancels tasks, drains orphaned loads, then does `self._loaded.clear()` — no `_close_loaded_model` for still-resident models. This contradicts the "used by every lifecycle exit" contract (`:1001`): Flash prefetcher/weight-store executors (48 threads + per-layer fds), batch schedulers, and the whisper `ModelHolder` strong reference are never released. Masked by process exit in production, but wrong for tests or any embedding of the manager.

**3. Evictees leak on two exception paths (`model_manager.py:1194-1211`, `:1356-1408`).**
(a) `_pop_lru_evictees` raises `RuntimeError` mid-loop after already popping victims into its local list — those models are gone from `_loaded` but never passed to `_close_evictees` (prefetcher/weight-store pools, disk cache, grammar-cache drops all leak). Reachable when `len(_loaded) > max_loaded_models` with insufficient idle models (e.g. after a settings decrease). (b) In `_load_adapter_model`, any raise between `evictees = self._pop_lru_evictees()` and `await self._close_evictees(evictees)` at `:1408` leaks the popped evictees the same way — `ensure_loaded` closes evictees in its own exception handling; this path doesn't.

## Suggested fix
See each finding — add try/except around startup, close resident models in `stop()`, and ensure evictees are always closed on the exception paths.

## Severity
Medium — resource leaks and a test-poisoning startup failure.

---

dpalmqvist:
## Summary
Three lifecycle defects in app / model-manager teardown.

1. **Failed startup poisoned the metrics registry.** There was no exception protection between `manager.start_expiry_checker()` and `yield`, so if `mkdir` / `REGISTRY.register` / `agent_service.startup()` raised, the post-`yield` shutdown never ran: the expiry task was orphaned and the stats collector stayed registered in the module-level `REGISTRY` — making every later `create_app()` (e.g. across tests in one process) fail with a duplicate-collector error. Now the post-`start_expiry_checker` startup is wrapped in try/except that calls `manager.stop()` + unregisters the collector before re-raising.
2. **`ModelManager.stop()` cleared `_loaded` without closing the models** — leaking Flash prefetcher/weight-store executors (threads + per-layer fds), batch schedulers, and the whisper `ModelHolder` strong ref. Now it closes all resident models via `_close_evictees` before clearing.
3. **Evictees leaked on two exception paths.** (a) `_pop_lru_evictees` raised `RuntimeError` mid-loop after popping victims into its local list — those models were gone from `_loaded` but never closed; it now closes the already-popped victims before raising. (b) `_load_adapter_model` leaked the popped evictees on any raise between `_pop_lru_evictees()` and `_close_evictees()`; it now closes them in an `except` handler like `ensure_loaded` already does.

## Tests
Added for all three: failed-startup-unregisters-collector (a second lifespan succeeds), stop()-closes-resident-models, _pop_lru_evictees-closes-partial-on-failure. Affected suites pass; ruff clean.

Closes #626

### DIFF
--- a/olmlx/app.py
+++ b/olmlx/app.py
@@ -136,25 +136,46 @@ async def _lifespan_inner(app: FastAPI):
     )
     manager.start_expiry_checker()
 
-    settings.models_dir.mkdir(parents=True, exist_ok=True)
+    # Everything from here to ``yield`` acquires resources whose teardown lives
+    # in the post-``yield`` shutdown block — which never runs if startup raises.
+    # Roll back on failure so a failed startup doesn't orphan the expiry task or
+    # leave the stats collector registered in the module-level REGISTRY (which
+    # would poison every later create_app() with a duplicate-collector error,
+    # e.g. across tests in one process) — #626.
+    try:
+        settings.models_dir.mkdir(parents=True, exist_ok=True)
 
-    app.state.registry = registry
-    app.state.model_manager = manager
-    app.state.model_store = store
+        app.state.registry = registry
+        app.state.model_manager = manager
+        app.state.model_store = store
 
-    # Lazy gauge/counter collector reads the manager at scrape time.
-    stats_collector = metrics_mod.OlmlxStatsCollector(manager)
-    metrics_mod.REGISTRY.register(stats_collector)
-    app.state.metrics_collector = stats_collector
+        # Lazy gauge/counter collector reads the manager at scrape time.
+        stats_collector = metrics_mod.OlmlxStatsCollector(manager)
+        metrics_mod.REGISTRY.register(stats_collector)
+        app.state.metrics_collector = stats_collector
 
-    # Autonomous agent (#445): recover crash-orphaned runs at startup. The
-    # service itself is built in create_app() (gated on agent_enabled) so the
-    # routes exist without a lifespan; this only runs the async startup scan.
-    agent_service = (
-        getattr(app.state, "agent_service", None) if settings.agent_enabled else None
-    )
-    if agent_service is not None:
-        await agent_service.startup()
+        # Autonomous agent (#445): recover crash-orphaned runs at startup. The
+        # service itself is built in create_app() (gated on agent_enabled) so
+        # the routes exist without a lifespan; this only runs the async scan.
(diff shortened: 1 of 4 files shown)


dpalmqvist:
Addressed the rollback point: each rollback step is now independently guarded, so a failing `manager.stop()` no longer skips the collector unregister (added a regression test that makes stop() raise and asserts a second lifespan still starts clean).

On the smaller concern — `stop()` clearing `_loaded` before closing: this is the same pop-then-close ordering eviction uses, and `_close_evictees` is robust to per-model close failures (it pops from its local list as it closes and drains the remainder synchronously on any abnormal exit), so the resident models are still closed even if one close fails. The local `resident` list keeps them referenced until then, so there's no leak. Left as-is.

PR #706, opened 2026-07-29
fix(reap): resolve lm_head through VLM language towers; refuse untied embed fallback

dpalmqvist:
Second bug found by the Qwen3.6-35B-A3B hybrid smoke run (follow-up to #705): `reap report`'s perplexity gate produced worse-than-uniform numbers (full-model ppl ~258k) on the VLM-wrapped `qwen3_5_moe` checkpoint.

**Root cause:** `_resolve_head` only looked for `lm_head` on the outer model. On qwen3_5_moe-style wrappers it lives on `model.language_model`, so the resolver silently fell back to `embed_tokens.as_linear` — which is only valid for tied embeddings; this checkpoint is untied, so every logit was computed against the wrong matrix.

**Fix:**
- `_find_lm_head` probes the model and its `language_model` tower.
- The embedding fallback now requires `tie_word_embeddings=True` (checked on both holders); an untied model with no findable head is a hard `ValueError` instead of silent garbage.
- The `keep_head=False` nullification path uses the same resolution, so nested heads are freed during calibration too.

**Tests:** 4 new unit tests on `_resolve_head` (outer head, nested VLM head, tied fallback, untied-raises). 101 reap tests green, pyright clean.

### DIFF
--- a/olmlx/engine/reap/calibrate.py
+++ b/olmlx/engine/reap/calibrate.py
@@ -239,13 +239,40 @@ def collect_saliency(
     return acc, meta
 
 
+def _find_lm_head(model):
+    """lm_head on the model or its VLM language tower (qwen3_5_moe-style)."""
+    for holder in (model, getattr(model, "language_model", None)):
+        head = getattr(holder, "lm_head", None) if holder is not None else None
+        if head is not None:
+            return head
+    return None
+
+
+def _is_tied(model) -> bool:
+    for holder in (model, getattr(model, "language_model", None)):
+        args = getattr(holder, "args", None) if holder is not None else None
+        if args is not None and getattr(args, "tie_word_embeddings", False):
+            return True
+    return False
+
+
 def _resolve_head(model, inner):
-    """Returns (norm, head_fn) for final logits; head_fn may use tied embeddings."""
+    """Returns (norm, head_fn) for final logits.
+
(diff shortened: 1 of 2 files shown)


olmlx/engine/reap/calibrate.py
```diff
@@ -239,13 +239,40 @@ def collect_saliency(
...
+    scoring an untied model through embed.as_linear yields random-quality
+    logits (worse-than-uniform perplexity), so that case is a hard error.
+    """
     norm = getattr(inner, "norm", None) or getattr(inner, "final_layernorm", None)
-    lm_head = getattr(model, "lm_head", None)
+    lm_head = _find_lm_head(model)
     if lm_head is not None:
         return norm, lm_head
-    return norm, inner.embed_tokens.as_linear
+    if _is_tied(model):
```
dpalmqvist, line 270: Deliberately scoped: `language_model` is the uniform tower attribute across mlx-lm/mlx-vlm (the same convention set `_get_backbone` handles), and no current model nests the head deeper. The key change in this PR is that an *unrecognized* layout now raises a clear ValueError instead of silently scoring through the embedding matrix — so a future wrapper variant fails loudly at the first report run rather than producing plausible-looking garbage. Probing speculative attribute names would be untestable dead code today; happy to extend the holder chain when a real model needs it.

Issue #470, opened 2026-06-09
ci: zero real inference in CI + self-hosted runner exposed to pull_request workflows


dpalmqvist:
Two related CI gaps:

## 1. CI never exercises real inference

The autouse MLX mock in `tests/integration/conftest.py` plus `-m "not real_model"` in `ci.yml` means CI validates orchestration only. History shows real-model bugs slipping through exactly this gap (e.g. the old VLM text-path hack silently dropping images, #429), and `tests/live/` exists specifically to dodge the mock — but is developer-run only.

**Proposal:** a nightly or manually-triggered workflow on the self-hosted Mac running a small curated `real_model` smoke subset (one text model, one VLM, one speculative config) with the known SIGBUS-at-Metal-teardown mitigations (run per-test process via `pytest-forked` or a per-file loop so exit 138 doesn't poison the batch).

Cheaper complementary guard: a pytest hook that fails a test which touches real MLX entry points without the `real_model` marker, so a forgotten marker can't silently produce a mocked false-positive.

## 2. `pull_request`-triggered workflows on a self-hosted runner

`ci.yml` (and the aider review workflow) run on `runs-on: self-hosted` with a `pull_request` trigger. If this repo ever accepts fork PRs, that is arbitrary code execution on the host (checkout + `uv sync` + pytest of attacker-controlled code). GitHub's first-time-contributor approval gate mitigates but doesn't eliminate this.

**Proposal:** restrict the workflows to same-repo branches (`if: github.event.pull_request.head.repo.full_name == github.repository`), and/or set Actions settings to require approval for *all* outside collaborators.

_Found during a full-repo code review._

PR #717, opened 2026-08-05
fix: fall back to af_bella when Kokoro voice crashes on TTS (closes #703)

### LINKED ISSUE #703: /v1/audio/speech: Kokoro af_heart voice crashes with broadcast_shapes error on ordinary sentences
## Summary

`/v1/audio/speech` with model `prince-canuma/Kokoro-82M` and voice `af_heart` crashes with a raw `[broadcast_shapes]` internal error for several ordinary, non-adversarial input sentences, while the same sentences succeed with other voices (e.g. `af_bella`) and while shorter/differently-worded sentences succeed with `af_heart` itself. The failure is deterministic/reproducible for a given input+voice pair.

## Reproduction

Server: `uv run olmlx serve` on `http://localhost:11434`. Requires the `audio` extra installed (`uv sync --extra audio`) plus the `en_core_web_sm` spaCy model, per the README/pyproject notes — otherwise `misaki` raises an `ImportError` first (a separate, expected setup-dependent failure, not this bug).

```
curl -s http://localhost:11434/v1/audio/speech -d '{
  "model": "prince-canuma/Kokoro-82M",
  "input": "Hello, this is a test of text to speech.",
  "voice": "af_heart"
}'
# => HTTP 400
# {"error":{"message":"[broadcast_shapes] Shapes (1,70200,1) and (1,70500,9) cannot be broadcast.","type":"invalid_request_error","code":"invalid_value"}}
```

Reproduced identically on a second, differently-worded sentence of similar length:

```
curl -s http://localhost:11434/v1/audio/speech -d '{
  "model": "prince-canuma/Kokoro-82M",
  "input": "The weather today is quite pleasant and sunny.",
  "voice": "af_heart"
}'
# => same HTTP 400, same "Shapes (1,70200,1) and (1,70500,9) cannot be broadcast." error
```

and again with:
```
"input": "This is a medium length sentence for testing."   # => HTTP 400, broadcast_shapes error
```

Control cases that succeed (HTTP 200, real MP3 returned):
```
"input": "Hello, this is a test of text to speech.", "voice": "af_bella"   # 200 OK
"input": "Hello, this is a test.",                    "voice": "af_heart" # 200 OK
"input": "The weather today is quite pleasant.",       "voice": "af_heart" # 200 OK
"input": "This is a short one.",                       "voice": "af_heart" # 200 OK
```

So it isn't purely a length threshold (a 54-char sentence succeeds, a 45-char one fails) — it looks input/segmentation-dependent, but is specific to the `af_heart` voice; every tested sentence with `af_bella` succeeded.

## Observed vs expected

Observed: HTTP 400 with a raw internal numpy/mlx broadcast error leaking from deep inside the Kokoro pipeline (`mlx_audio.tts.models.kokoro`), for ordinary English sentences that aren't malformed input.

Expected: either the request should succeed (if this is a genuine olmlx-side or mlx-audio-side bug in duration/style-vector broadcasting for the `af_heart` blended voice), or at minimum the server should not surface a raw `[broadcast_shapes] Shapes (...) cannot be broadcast` message as a client-facing `invalid_request_error` — this isn't a client input-validation problem, it's an internal computation failure during generation.

## Server log excerpt

```
2026-07-29 08:43:00 [WARNING] olmlx: [ac62b7ea] HTTPException 400 on /v1/audio/speech: [broadcast_shapes] Shapes (1,70200,1) and (1,70500,9) cannot be broadcast.
Creating new KokoroPipeline for language: a
Traceback (most recent call last):
  File ".../olmlx/routers/audio.py", line 223, in create_speech
    first = await pcm_agen.__anext__()
  File ".../olmlx/routers/audio.py", line 190, in _pcm_chunks
    async for seg in generate_speech(...)
  File ".../olmlx/engine/inference.py", line 5079, in generate_speech
    raise item
  File ".../olmlx/engine/inference.py", line 5058, in _worker
    for result in lm.model.generate(text, voice=voice, speed=speed):
  File ".../mlx_audio/tts/models/kokoro/kokoro.py", line ~, in generate
    ...
```

## Root cause direction

The crash originates inside `lm.model.generate()` (`mlx_audio.tts.models.kokoro.kokoro`, mlx-audio 0.4.4), not in olmlx's own request-handling code (`olmlx/engine/inference.py:5058` just iterates the third-party generator). `generate_speech`'s `_worker` (inference.py:5069) catches *any* `Exception` from the mlx-audio generator and re-raises it on the event loop, where it currently surfaces as a generic 400 `invalid_request_error` — worth distinguishing "bad client input" from "TTS backend crashed" (500) regardless of the upstream root cause.

Environment: olmlx main @ 02e7b4f, mlx-audio 0.4.4, macOS, `uv run olmlx serve`.

botbikamordehai2-sketch:
## What

Issue #703 reported that `/v1/audio/speech` with model `prince-canuma/Kokoro-82M` and voice `af_heart` crashes with a raw `[broadcast_shapes]` error for several ordinary sentences (e.g., "Hello, this is a test of text to speech."), while other voices succeed. The error originates inside the Kokoro TTS engine, not in this file, but VoiceIO currently lets that exception abort the entire response.

## Fix

This patch adds a per-sentence try/except around the `generate_speech` call in `VoiceIO.speak`. If the configured voice raises an exception (including the broadcast_shapes crash), we log a warning and retry the same sentence with the known-good voice `af_bella`. This keeps the chat session functional and avoids exposing raw internal errors to users.

Note: The root cause in Kokoro/misaki is a separate engine-level bug; this fix is a defensive workaround. A later PR could patch the engine to handle the waveform length mismatch gracefully.

Closes #703

### DIFF
--- a/olmlx/chat/voice/io.py
+++ b/olmlx/chat/voice/io.py
@@ -70,9 +70,22 @@ class VoiceIO:
         24 kHz float32 segments; each is played as it arrives.
         """
         for sentence in split_into_sentences(text):
-            async for segment in generate_speech(
-                self._manager, self._tts_model, sentence, voice=self._voice
-            ):
-                # play() blocks on sd.wait() until the segment finishes; keep it
-                # off the event loop.
-                await asyncio.to_thread(playback.play, segment, _TTS_SAMPLE_RATE)
+            try:
+                async for segment in generate_speech(
+                    self._manager, self._tts_model, sentence, voice=self._voice
+                ):
+                    # play() blocks on sd.wait() until the segment finishes; keep it
+                    # off the event loop.
+                    await asyncio.to_thread(playback.play, segment, _TTS_SAMPLE_RATE)
+            except Exception as exc:
+                # Some Kokoro voices (e.g., af_heart) crash with broadcast_shapes
+                # errors for certain ordinary sentences (issue #703). Fall back to
+                # a known-good voice so speech synthesis does not abort the session.
+                logger.warning(
+                    "TTS failed with voice %s on sentence %r (%s); retrying with af_bella",
+                    self._voice, sentence, exc
+                )
+                async for segment in generate_speech(
+                    self._manager, self._tts_model, sentence, voice="af_bella"
+                ):
+                    await asyncio.to_thread(playback.play, segment, _TTS_SAMPLE_RATE)


dpalmqvist:
## Code review

Thanks for taking a look at #703 — and for being upfront in the PR body that this is a defensive workaround rather than a root-cause fix. Unfortunately I don't think this one can land as written, for a few reasons that go beyond style.

### 1. `af_bella` is not a known-good fallback

This is the blocker. The premise of the patch is "retry with the known-good voice `af_bella`", but that's contradicted by the follow-up investigation on the issue ([#703 comment](https://github.com/motsognirr/olmlx/issues/703#issuecomment-5127504693)):

```
"input": "Quick test.", "voice": "af_bella"
# => HTTP 400
# [broadcast_shapes] Shapes (1,37200,1) and (1,37500,9) cannot be broadcast.  (repeatable)
```

`af_heart` and `am_adam` fail on the same input; `bf_emma` succeeds. The bug is a (voice-style, phoneme/duration-count) interaction inside `mlx_audio.tts.models.kokoro`'s duration/style-vector broadcasting across the whole `a` pipeline — it is **not** an `af_heart`-specific defect. So the retry will frequently fail too, at which point the exception propagates anyway, just slower and after a misleading warning.

### 2. It patches a different code path from the one in the report

#703 reproduces against `/v1/audio/speech`, which is `create_speech` in `olmlx/routers/audio.py:150`. This diff only touches `olmlx/chat/voice/io.py` — the in-process terminal chat voice mode, a separate entry point. The `curl` repro in the issue is entirely unaffected by this change.

Since the PR says **"Closes #703"**, merging would auto-close an issue that is still broken on the reported path.

### 3. It conflicts with the plan already accepted on the issue

#703 carries `plan:proposed`. The agreed approach is to add a `TTSGenerationError(RuntimeError)` in `generate_speech._worker` (`olmlx/engine/inference.py`) so mlx-audio backend crashes surface as HTTP 500 instead of a bogus 400 `invalid_request_error` with a raw mlx string leaking to the client. That plan explicitly weighed and rejected a retry workaround, on the grounds that a speculative retry could mask a real regression.

### Secondary issues

- **Duplicate audio.** `generate_speech` is a streaming generator. If the voice fails after N segments have already been played, the fallback re-synthesizes the sentence from the beginning and replays it.
- **Hardcoded Kokoro voice in a model-agnostic path.** `self._tts_model` can be any mlx-audio model (fish/spark/indextts). `af_bella` is meaningless for those, so the retry is a guaranteed second failure that also obscures the original error.
- **Bare `except Exception` is too broad.** It swallows `UnknownVoiceError`, `ServerBusyError`, model-load failures and OOM, then pointlessly retries them.
- **No tests.** `CLAUDE.md` is explicit: "For bug fixes, write a test that reproduces the bug before writing the fix."
- **Ruff would fail CI.** `uv run ruff format --check olmlx/chat/voice/io.py` → `Would reformat` (the `logger.warning` arguments are on one line; line 85 is 91 chars). Please run `ruff check` + `ruff format` before pushing.

### Where to go from here

I'd rather not merge a fallback that we've measured doesn't hold. If you'd like to keep going on #703, the error-classification change is well-scoped and genuinely useful:

1. Add `class TTSGenerationError(RuntimeError)` in `olmlx/engine/inference.py` next to `ServerBusyError`.
2. In `generate_speech._worker`, wrap whatever `lm.model.generate()` raises into `TTSGenerationError(f"{type(exc).__name__}: {exc}")` before enqueuing it, instead of re-raising bare. Leave the pre-loop `is_tts` `ValueError` alone — that one is a genuine 400.
3. Tests in `tests/test_inference_speech.py` (backend crash raises `TTSGenerationError`) and `tests/test_routers_speech.py` (that error yields 500, while `test_speech_non_tts_model_400` still yields 400).

That fixes the "at minimum" ask in the issue, on the path the issue actually reports. Happy to review it.

dpalmqvist:
### CI note + local verification results

Heads up on why this PR shows no checks: **CI cannot run on fork PRs by design.** Every job in `.github/workflows/ci.yml` (and `aider-code-review.yml`) is gated on:

```yaml
if: github.event.pull_request.head.repo.full_name == github.repository
```

The runners are a self-hosted Mac, so running a fork branch there would mean arbitrary code execution on that host (issue #470). That's a deliberate security control and not something to bypass for an external PR.

So I ran the equivalent checks locally against `c26e7db` instead:

| Check | Result |
|---|---|
| `ruff check .` | ✅ All checks passed |
| `ruff format --check .` | ❌ **Would reformat: `olmlx/chat/voice/io.py`** |
| `pyright olmlx/chat/voice/io.py` | ✅ 0 errors, 0 warnings |
| `pytest` — `test_voice_io`, `test_voice_cli`, `test_tts_voice`, `test_config_voice`, `test_routers_speech`, `test_inference_speech` | ✅ 31 passed |

Two things to note about that green test column:

1. The Ruff **format** job would fail CI as-is — please run `uv run ruff format` (the `logger.warning` arguments need to be split across lines; line 85 is 91 chars).
2. The 31 passing tests pass *vacuously* — none of them exercise the new fallback branch. There is currently **zero** test coverage of the added code, which is the `CLAUDE.md` TDD point from my review above.

Neither of these changes the main verdict in my earlier comment: the `af_bella` fallback is contradicted by the measurements on #703, and this file isn't on the `/v1/audio/speech` path the issue reports.

botbikamordehai2-sketch:
מי זה  >


‫בתאריך יום ג׳, 22 בספט׳ 2026 ב-14:38 מאת ‪Daniel Palmqvist‬‏ <‪
***@***.***‬‏>:‬

dpalmqvist:
Closing — #703 has been fixed at the root in #719 (float-rounded interpolate sizes in Kokoro caused the `[broadcast_shapes]` crash) and #720 (mlx-audio floor bumped to 0.4.5, shim removed), so a voice fallback is no longer needed. Thanks for the contribution!

